PR #79, opened 2019-02-25
O365Explorer and WindowsHostExplorer notebooks

ianhelle:
Fixes for Alert Investigation - Process Alerts notebook.
Minor updates to Demo notebook.
Also added requirements.txt for configuring AzureNotebooks projects.


ianhelle:
> Do you really want to put your personal email in the notebook? We are getting a feedback email from Koby that we can use for this.
@juliango2100 Ok - no - thought I'd removed those.

PR #102, opened 2019-02-27
Notebook fixes #3

ianhelle:
This includes mostly defensive fixes to the notebooks - no real functionality changes.
Also using the TenantId explicitly in the authentication to cover cases where the user account belongs to a different tenant than the workspace.
Added completed versions of the notebooks to samples folder (so that people can see what they are meant to look like).
Updated Get Started and Readme.md with a few fixes/additions (without overwriting Zhipeng's changes this time).


ianhelle:
Removed the msticpy reference from Get Started nb

PR #140, opened 2019-04-04
Expansion Queries New Format Example

laithhisham:
## Proposed Changes ##

  - Reformatted the SecurityEvent/ParentProcessesOnHost expansion query according to the new format that we defined recently.

> Notice the added comments as well which can be used as a general reference to new queries.

### DIFF
--- a/Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
+++ b/Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
@@ -29,7 +29,7 @@ and ParentProcessName !contains ":\\Windows\\CCM\\CcmExec.exe"
 | project min_TimeGenerated, max_TimeGenerated, EventID, Account, AccountType, Computer, ParentProcessName, NewProcessName, TargetAccount, CommandLine
 | top 10 by min_TimeGenerated asc
 // rename the relevant columns to the format {V3EntityType_FieldName}
-| project-rename Host_NetBiosName=Computer, Account_DisplayName=Account, Process_CommandLine=CommandLine, Process_ImageFile_FullPath=NewProcessName, Process_ParentProcess_ImageFile_FullPath=ParentProcessName
+| project-rename Process_Host_NetBiosName=Computer, Process_Account_DisplayName=Account, Process_CommandLine=CommandLine, Process_ImageFile_FullPath=NewProcessName, Process_ParentProcess_ImageFile_FullPath=ParentProcessName
 };
 // change <HostName> value below
 GetParentProcessesOnHost(toupper("<HostName>"))
--- a/Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
+++ b/Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
@@ -1,20 +1,18 @@
-// Name: Parent Processes running on Host
+// Id: 992b9912-a11e-487d-83b3-1d1ba6081746
+// DisplayName: Parent Processes running on Host
 // Description: Any 4688 event that contains Parent Process running on a given host during the time of or recently after compromise and in the range of +-10m
-// Entity: #Host
-// Input: Computername, Hostname, HostId
-// Output: Processes
-//
-// QueryPeriod: -1h and +30m default, change as needed, not all event 4688 have Parent Process available
-//
-// Data Source: #SecurityEvent
-//
+// InputEntityType: Host
+// InputFields: [HostName]
+// OutputEntityTypes: [Process]
+// QueryPeriodBefore: 10
+// QueryPeriodBeforeUnit: m
+// QueryPeriodAfter: 10
+// QueryPeriodAfterUnit: m
+// DataSources: [SecurityEvent]
 // Tactics: #Persistence, #Discovery, #LateralMovement, #Collection
-//
-let GetParentProcessesOnHost = (suspiciousEventTime:datetime, v_Host:string){
-let v_StartTime = suspiciousEventTime-10m;
-let v_EndTime = suspiciousEventTime+10m;
-SecurityEvent 
-| where TimeGenerated between (v_StartTime .. v_EndTime)
+
+let GetParentProcessesOnHost = (v_Host:string){
+SecurityEvent
 | where toupper(Computer) contains v_Host or toupper(WorkstationName) contains v_Host
 | where EventID == 4688 
 | where isnotempty(ParentProcessName)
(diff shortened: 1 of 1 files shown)


Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
```diff
@@ -1,20 +1,18 @@
-// Name: Parent Processes running on Host
+// Id: 992b9912-a11e-487d-83b3-1d1ba6081746
+// DisplayName: Parent Processes running on Host
 // Description: Any 4688 event that contains Parent Process running on a given host during the time of or recently after compromise and in the range of +-10m
```
timbMSFT: you have removed the timeframe variables so should adjust the comment/description accordingly

Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
```diff
@@ -1,20 +1,18 @@
...
-// Entity: #Host
-// Input: Computername, Hostname, HostId
-// Output: Processes
-//
-// QueryPeriod: -1h and +30m default, change as needed, not all event 4688 have Parent Process available
-//
-// Data Source: #SecurityEvent
-//
+// InputEntityType: Host
+// InputFields: [HostName]
```
timbMSFT: This doesn't relate to any field names in the data being queried. Should it?

Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
```diff
@@ -1,20 +1,18 @@
...
+// OutputEntityTypes: [Process]
+// QueryPeriodBefore: 10
+// QueryPeriodBeforeUnit: m
```
timbMSFT: This seems much less legible than the more concise 10m, 1d (which also translates directly into corresponding KQL).

Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
```diff
@@ -1,20 +1,18 @@
...
+// QueryPeriodAfter: 10
+// QueryPeriodAfterUnit: m
+// DataSources: [SecurityEvent]
```
timbMSFT: Again - why not align with the approach taken for queries under Detections folder?

Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
laithhisham: the timeframe was removed from the query but would still be applied by us when running the query.
but I think you meant that when copying the query and executing it externally it would not longer make sense
will change it accordingly

Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
laithhisham: not sure I get it... the input to the defined function "v_Host" is the HostName field.. so it is being used

Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
laithhisham: changed as suggested

Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
laithhisham: same, changed

Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
```diff
@@ -1,20 +1,16 @@
...
-//
+// Id: 992b9912-a11e-487d-83b3-1d1ba6081746
+// DisplayName: Parent Processes running on Host
+// Description: Any 4688 event that contains Parent Process running on a given host during the time of or recently after compromise (set time range to +- 10m when running the query)
+// InputEntityType: Host
+// InputFields: [HostName]
+// OutputEntityTypes: [Process]
+// QueryPeriodBefore: 10m
+// QueryPeriodAfter: 10m
+// DataSources: #SecurityEvent
```
laithhisham: Changed in order to maintain the GitHub searching functionality.

In case more than 1 DataSource is being used, it would become:
// DataSources: #SecurityEvent, #DnsEvents
[i.e. comma separated]

Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
```diff
@@ -29,7 +29,7 @@ and ParentProcessName !contains ":\\Windows\\CCM\\CcmExec.exe"
 | project min_TimeGenerated, max_TimeGenerated, EventID, Account, AccountType, Computer, ParentProcessName, NewProcessName, TargetAccount, CommandLine
 | top 10 by min_TimeGenerated asc
 // rename the relevant columns to the format {V3EntityType_FieldName}
-| project-rename Host_NetBiosName=Computer, Account_DisplayName=Account, Process_CommandLine=CommandLine, Process_ImageFile_FullPath=NewProcessName, Process_ParentProcess_ImageFile_FullPath=ParentProcessName
+| project-rename Process_Host_NetBiosName=Computer, Process_Account_DisplayName=Account, Process_CommandLine=CommandLine, Process_ImageFile_FullPath=NewProcessName, Process_ParentProcess_ImageFile_FullPath=ParentProcessName
```
laithhisham: The convention is to **always have the name of the output entity type as a prefix to the nested fields**.
In the example above, the query returns a Process, and the Process has a Host field within it. So the column representing the HostName of that field would be written as: **Process_Host_HostName**.

This is to allow for returning several entities as an output without having ambiguous naming.

If the same query is to return an Account, and that account also has a Host which has a HostName, then the column representing that would be written as: **Account_Host_HostName**

Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
```diff
@@ -9,9 +9,8 @@
 // DataSources: #SecurityEvent
 // Tactics: #Persistence, #Discovery, #LateralMovement, #Collection
 
-let GetParentProcessesOnHost = (v_Host:string){
+let GetParentProcessesOnHost = (v_Host_HostName:string){
```
laithhisham: agreed on the following naming for function parameters for clarity
**v_Entity_Field**

Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
```diff
@@ -23,6 +22,7 @@ and ParentProcessName !contains ":\\Windows\\CCM\\CcmExec.exe"
 // excluding well known parent process, child process combinations, feel free to add more specific to the environment
 | where (ParentProcessName !contains ":\\Windows\\System32\\svchost.exe" and (NewProcessName !contains ":\\Windows\\System32\\wbem\\WmiPrvSE.exe" or NewProcessName !contains ":\\Windows\\SysWOW64\\wbem\\WmiPrvSE.exe"))
 | where (ParentProcessName !contains ":\\Windows\\System32\\services.exe" and NewProcessName !contains ":\\Windows\\servicing\\TrustedInstaller.exe")
+| where toupper(Computer) contains v_Host_HostName or toupper(WorkstationName) contains v_Host_HostName
```
laithhisham: lines of the format (toupper(), tolower(), etc...) are computationally heavier than other simple clauses. So it is good to try and push them as much further down the query as possible so they do not affect the performance of the query

PR #147, opened 2019-04-27
Replaced 'Domain' with 'Computer' in the join because domain will be …

chm0dx:
…'Builtin' for the group and host/domain for the user, preventing a match. Also replaced 'CreatedUser' with 'CreatedUserSid' because the group add entry may not reliably include the name of the user added to the group (SID is also a more reliable identifier in general). Lasltly, removed 'where CreatedUserTime < GroupAddTime' because I'm not sure any other case is possible without manually rigging the logs.

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/SecurityEvent/UserCreatedAddedToBuiltinAdmins_1d.txt
+++ b/Detections/SecurityEvent/UserCreatedAddedToBuiltinAdmins_1d.txt
@@ -31,9 +31,8 @@ SecurityEvent
 | where EventID == 4732
 //TargetSid is the builin Admins group: S-1-5-32-544
 | where TargetSid == "S-1-5-32-544"
-| project GroupAddTime = TimeGenerated, GroupAddEventID = EventID, GroupAddActivity = Activity, Computer = toupper(Computer), GroupName = TargetUserName, Domain = toupper(TargetDomainName), GroupSid = TargetSid, UserAdded = SubjectUserName, UserAddedSid = SubjectUserSid, CreatedUser = tolower(SubjectUserName)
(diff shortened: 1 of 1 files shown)


Detections/SecurityEvent/UserCreatedAddedToBuiltinAdmins_1d.txt
```diff
@@ -31,9 +31,8 @@ SecurityEvent
 | where EventID == 4732
 //TargetSid is the builin Admins group: S-1-5-32-544
 | where TargetSid == "S-1-5-32-544"
-| project GroupAddTime = TimeGenerated, GroupAddEventID = EventID, GroupAddActivity = Activity, Computer = toupper(Computer), GroupName = TargetUserName, Domain = toupper(TargetDomainName), GroupSid = TargetSid, UserAdded = SubjectUserName, UserAddedSid = SubjectUserSid, CreatedUser = tolower(SubjectUserName)
+| project GroupAddTime = TimeGenerated, GroupAddEventID = EventID, GroupAddActivity = Activity, Computer = toupper(Computer), GroupName = TargetUserName, Domain = toupper(TargetDomainName), GroupSid = TargetSid, UserAdded = SubjectUserName, UserAddedSid = SubjectUserSid, CreatedUser = tolower(SubjectUserName), CreatedUserSid = MemberSid
 )
-on CreatedUser, Domain
+on CreatedUserSid, Computer
```
shainw: Agreed, SID is better and nice catch on the Domain, however, we should remove Computer also here as it is possible that the computer where the User create event occurs and the User add to group event occurs are different.  I think SID will suffice...like you said much higher fidelity match.

chm0dx: Agreed, I think just filtering on CreatedUserSid here is enough. I'll make the change and resubmit. Thanks!

chm0dx: Done d20075d0f7294cd02f4769999f6b966453f2be46

PR #156, opened 2019-05-07
Exploration Queries ReFormat according to the new guidelines

laithhisham:
## Proposed Changes

  - Re formatted the exploration queries according to the new format we agreed on [here](https://nam06.safelinks.protection.outlook.com/ap/w-59584e83/?url=https%3A%2F%2Fmicrosoft-my.sharepoint.com%2F%3Aw%3A%2Fp%2Frazhe%2FEbaL_DFRmhJHkenuo-ozMD4BJoRZv6nEdhuV-FOp_Y4xBw%3Fe%3Da1j7bT&data=02%7C01%7CLaith.Hisham%40microsoft.com%7C276779b5839e42f512d808d6b92be4ec%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C636899995570889071&sdata=9dL2agiIZgJU2XIT%2BOp%2BVTwyIsxkXZPIjvIwjJ6XVCo%3D&reserved=0)
  - Assigned a GUID id to each query
  - Added top 10 crop to each query in order to be consistent across all the queries and prevent from excessive results.

### DIFF
--- a/Exploration Queries/DnsEvents/LeastPrevClientIP-DNSNameQueryToIP.txt
+++ b/Exploration Queries/DnsEvents/LeastPrevClientIP-DNSNameQueryToIP.txt
@@ -1,27 +1,23 @@
-// Name: DNS Name Lookup Query from least prevalent ClientIP to remote IP Address
-// Description: Summary of Bottom 10 Client IP and Domain Names for a given remote IPAddress from DnsEvent Lookup Query data.
-//
-// Entity: IPAddress
-// Input: IPAddress
-// Output: ClientIP
-//
-// QueryPeriod:  +-3h
-//
-// Data Source: #DnsEvents
-//
+// Id: 801bacb0-612a-4195-a84f-7939cca63b92
+// DisplayName: DNS Name Lookup Query from least prevalent ClientIP to remote IP Address
+// Description: Summary of Bottom 10 Client IP and Domain Names for a given remote IPAddress from DnsEvent Lookup Query data (set time range to +-3h when running the query)
+// InputEntityType: Ip
+// InputFields: [Address]
+// OutputEntityTypes: [Ip]
+// QueryPeriodBefore: 3h
+// QueryPeriodAfter: 3h
+// DataSources: #DnsEvents
 // Tactics: #CommandAndControl, #Exfiltration
-//
-let GetAllIPByClientIP = (suspiciousEventTime:datetime, v_IPAddress:string){
-let v_StartTime = suspiciousEventTime-3h;
-let v_EndTime = suspiciousEventTime+3h;
+
+let GetAllIPByClientIP = (v_IP_Address:string){
 DnsEvents
-| where TimeGenerated between (v_StartTime .. v_EndTime)
-| where SubType == "LookupQuery"
-| where IPAddresses contains v_IPAddress
+| where SubType == 'LookupQuery'
+| where IPAddresses contains v_IP_Address
 | summarize min(TimeGenerated), max(TimeGenerated), makeset(Name), count() by ClientIP, IPAddresses
-| project StartTimeUtc = min_TimeGenerated, EndTimeUtc = max_TimeGenerated, ClientIP, DomainNames = set_Name, IPAddresses, count_ 
+| project StartTimeUtc = min_TimeGenerated, EndTimeUtc = max_TimeGenerated, ClientIP, count_
 | top 10 by count_ asc nulls last
-| project StartTimeUtc, EndTimeUtc, ClientIP, IPAddresses, DomainNames
+| project StartTimeUtc, EndTimeUtc, ClientIP
+| project-rename IP_Address=ClientIP
 };
-// change datetime value and <ipaddress> value below
-GetAllIPByClientIP(datetime('2019-02-05T10:36:07Z'), "<ipaddress>")
+// change <Address> value below
+GetAllIPByClientIP('<Address>')
--- a/Exploration Queries/DnsEvents/LeastPrevClientIP-DNSNameQueryToIP.txt
+++ b/Exploration Queries/DnsEvents/LeastPrevClientIP-DNSNameQueryToIP.txt
@@ -14,9 +14,9 @@ DnsEvents
 | where SubType == 'LookupQuery'
 | where IPAddresses contains v_IP_Address
 | summarize min(TimeGenerated), max(TimeGenerated), makeset(Name), count() by ClientIP, IPAddresses
-| project StartTimeUtc = min_TimeGenerated, EndTimeUtc = max_TimeGenerated, ClientIP, count_
+| project StartTimeUtc = min_TimeGenerated, EndTimeUtc = max_TimeGenerated, ClientIP, DomainNames = set_Name, IPAddresses, count_
 | top 10 by count_ asc nulls last
-| project StartTimeUtc, EndTimeUtc, ClientIP
+| project StartTimeUtc, EndTimeUtc, ClientIP, IPAddresses, DomainNames
 | project-rename IP_Address=ClientIP
 };
 // change <Address> value below
--- a/Exploration Queries/DnsEvents/LeastPrevClientIP-DNSNameQueryToIP.txt
+++ b/Exploration Queries/DnsEvents/LeastPrevClientIP-DNSNameQueryToIP.txt
@@ -6,7 +6,7 @@
 // OutputEntityTypes: [Ip]
 // QueryPeriodBefore: 3h
 // QueryPeriodAfter: 3h
-// DataSources: #DnsEvents
+// DataSource: #DnsEvents
 // Tactics: #CommandAndControl, #Exfiltration
 
 let GetAllIPByClientIP = (v_IP_Address:string){
--- a/Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
+++ b/Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
@@ -1,35 +1,32 @@
-// Name: Parent Processes running on Host
-// Description: Any 4688 event that contains Parent Process running on a given host during the time of or recently after compromise and in the range of +-10m
-// Entity: #Host
-// Input: Computername, Hostname, HostId
-// Output: Processes
-//
-// QueryPeriod: -1h and +30m default, change as needed, not all event 4688 have Parent Process available
-//
-// Data Source: #SecurityEvent
-//
+// Id: 07da3cc8-c8ad-4710-a44e-334cdcb7882b
+// DisplayName: Parent Processes running on Host
+// Description: Any 4688 event that contains Parent Process running on a given host during the time of or recently after compromise (set time range to +-10m when running the query)
+// InputEntityType: Host
+// InputFields: [HostName]
+// OutputEntityTypes: [Process]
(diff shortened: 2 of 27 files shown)


Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
```diff
@@ -1,35 +1,32 @@
...
 // Tactics: #Persistence, #Discovery, #LateralMovement, #Collection
-//
-let GetParentProcessesOnHost = (suspiciousEventTime:datetime, v_Host:string){
-let v_StartTime = suspiciousEventTime-10m;
-let v_EndTime = suspiciousEventTime+10m;
+
+let GetParentProcessesOnHost = (v_Host_HostName:string){
 SecurityEvent 
-| where TimeGenerated between (v_StartTime .. v_EndTime)
-| where toupper(Computer) contains v_Host or toupper(WorkstationName) contains v_Host
```
laithhisham: lines of the format (toupper(), tolower(), etc...) are computationally heavier than other simple clauses. So it is good to try and push them as much further down the query as possible so they do not affect the performance of the query

Exploration Queries/DnsEvents/LeastPrevClientIP-DNSNameQueryToIP.txt
```diff
@@ -1,27 +1,23 @@
...
-| where IPAddresses contains v_IPAddress
+| where SubType == 'LookupQuery'
+| where IPAddresses contains v_IP_Address
 | summarize min(TimeGenerated), max(TimeGenerated), makeset(Name), count() by ClientIP, IPAddresses
-| project StartTimeUtc = min_TimeGenerated, EndTimeUtc = max_TimeGenerated, ClientIP, DomainNames = set_Name, IPAddresses, count_ 
+| project StartTimeUtc = min_TimeGenerated, EndTimeUtc = max_TimeGenerated, ClientIP, count_
 | top 10 by count_ asc nulls last
-| project StartTimeUtc, EndTimeUtc, ClientIP, IPAddresses, DomainNames
+| project StartTimeUtc, EndTimeUtc, ClientIP
+| project-rename IP_Address=ClientIP
```
laithhisham: rename the relevant columns to the format {V3EntityType_FieldName}

Exploration Queries/DnsEvents/LeastPrevClientIP-DNSNameQueryToIP.txt
```diff
@@ -1,27 +1,23 @@
...
-let v_EndTime = suspiciousEventTime+3h;
+
+let GetAllIPByClientIP = (v_IP_Address:string){
 DnsEvents
-| where TimeGenerated between (v_StartTime .. v_EndTime)
-| where SubType == "LookupQuery"
-| where IPAddresses contains v_IPAddress
+| where SubType == 'LookupQuery'
+| where IPAddresses contains v_IP_Address
 | summarize min(TimeGenerated), max(TimeGenerated), makeset(Name), count() by ClientIP, IPAddresses
```
laithhisham: always make sure to keep all the columns required for the construction of the fields in the output entity types, and **remove the ones that are not used, since this might cause duplicates in the constructed entities when ignoring these extra columns**

Exploration Queries/DnsEvents/LeastPrevClientIP-DNSNameQueryToIP.txt
```diff
@@ -1,27 +1,23 @@
...
-//
+// Id: 801bacb0-612a-4195-a84f-7939cca63b92
+// DisplayName: DNS Name Lookup Query from least prevalent ClientIP to remote IP Address
+// Description: Summary of Bottom 10 Client IP and Domain Names for a given remote IPAddress from DnsEvent Lookup Query data (set time range to +-3h when running the query)
+// InputEntityType: Ip
+// InputFields: [Address]
+// OutputEntityTypes: [Ip]
+// QueryPeriodBefore: 3h
+// QueryPeriodAfter: 3h
+// DataSources: #DnsEvents
```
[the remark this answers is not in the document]
timbMSFT: I think we should standardize on either DataSources or DataSource (irrespective of how many there are). Everywhere else uses DataSource today?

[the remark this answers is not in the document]
laithhisham: no problem, will change it to DataSource

Exploration Queries/DnsEvents/LeastPrevClientIP-DNSNameQueryToIP.txt
```diff
@@ -1,27 +1,23 @@
...
-// QueryPeriod:  +-3h
-//
-// Data Source: #DnsEvents
-//
+// Id: 801bacb0-612a-4195-a84f-7939cca63b92
+// DisplayName: DNS Name Lookup Query from least prevalent ClientIP to remote IP Address
+// Description: Summary of Bottom 10 Client IP and Domain Names for a given remote IPAddress from DnsEvent Lookup Query data (set time range to +-3h when running the query)
+// InputEntityType: Ip
+// InputFields: [Address]
+// OutputEntityTypes: [Ip]
```
timbMSFT: do we need an OutputField as well? relying on the entity as prefix feels fragile (imagine there are several fields with that prefix..)

laithhisham: query writers are free to use any field they want within the entity (of course the more the better)
however, there is no need to explicitly mention them since we know they are all fields of the given OutputEntityType(s), which do not overlap

Actually, all the fields are with the exact same prefix and this is what we want to have:
- IP_Address
- IP_Location_Latitude
- IP_Location_Longitude

--> this would result in constructing a V3 IP entity with the inner fields filled as appropriate.

Exploration Queries/DnsEvents/LeastPrevClientIP-DNSNameQueryToIP.txt
shainw: This query now results in useless info.  It simply returning the IP you looked up.  The goal here is to look up an IP and see what Domain Names are associated with it.  Needs to include -

| project StartTimeUtc = min_TimeGenerated, EndTimeUtc = max_TimeGenerated, ClientIP, DomainNames = set_Name, count_
| top 10 by count_ asc nulls last
| project StartTimeUtc, EndTimeUtc, ClientIP, DomainNames

shainw: I should say, it is returning the ClientIP's who did the lookup, but not the associated Domains.  This is very relevant and should be included in the result set

shainw: Clarifying a bit here, I think it would be good if we could pack the domainname with the IPaddress and then be able to pivot on both as it is possible that domainname has multiple IP's

laithhisham: True, this is an issue that we were trying to solve recently. Please have a look at the mail that I sent you. 

I want to be sure that I did not affect the query incorrectly, so I would like you to have a look and check that everything is good. 

Once we fully agree on the solution I will add it to all the queries

Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
```diff
@@ -1,35 +1,33 @@
...
-| where (ParentProcessName !contains ":\\Windows\\System32\\svchost.exe" and (NewProcessName !contains ":\\Windows\\System32\\wbem\\WmiPrvSE.exe" or NewProcessName !contains ":\\Windows\\SysWOW64\\wbem\\WmiPrvSE.exe"))
-| where (ParentProcessName !contains ":\\Windows\\System32\\services.exe" and NewProcessName !contains ":\\Windows\\servicing\\TrustedInstaller.exe")
-| summarize min(TimeGenerated), max(TimeGenerated) by EventID, Account, AccountType, Computer, ParentProcessName, NewProcessName, TargetAccount
-| project min_TimeGenerated, max_TimeGenerated, EventID, Account, AccountType, Computer, ParentProcessName, NewProcessName, TargetAccount
+| where(ParentProcessName !contains ':\\Windows\\System32\\svchost.exe' and (NewProcessName !contains ':\\Windows\\System32\\wbem\\WmiPrvSE.exe' or NewProcessName !contains ':\\Windows\\SysWOW64\\wbem\\WmiPrvSE.exe'))
+| where(ParentProcessName !contains ':\\Windows\\System32\\services.exe' and NewProcessName !contains ':\\Windows\\servicing\\TrustedInstaller.exe')
+| where toupper(Computer) contains v_Host_HostName or toupper(WorkstationName) contains v_Host_HostName
```
shainw: I agree normally, but I am not sure it is more expensive here in this scenario.  The query now is looking at all the events from all the computers in the table to exclude these NewProcessName using contains, whereas before it was matching the HostName first and then only iterating over the NewProcessName.  In general, I think HostName is probably a smaller index, but maybe not.

Exploration Queries/SecurityEvent/ProcessesOnHost.txt
```diff
@@ -1,35 +1,31 @@
...
 // excluding well known parent process, child process combinations, feel free to add more specific to the environment
-| where (ParentProcessName !contains ":\\Windows\\System32\\svchost.exe" and (NewProcessName !contains ":\\Windows\\System32\\wbem\\WmiPrvSE.exe" or NewProcessName !contains ":\\Windows\\SysWOW64\\wbem\\WmiPrvSE.exe"))
-| where (ParentProcessName !contains ":\\Windows\\System32\\services.exe" and NewProcessName !contains ":\\Windows\\servicing\\TrustedInstaller.exe")
-| summarize min(TimeGenerated), max(TimeGenerated) by Computer, Account, NewProcessName, CommandLine
-| project min_TimeGenerated, max_TimeGenerated, Computer, Account, NewProcessName, CommandLine
+| where (ParentProcessName !contains ':\\Windows\\System32\\svchost.exe' and (NewProcessName !contains ':\\Windows\\System32\\wbem\\WmiPrvSE.exe' or NewProcessName !contains ':\\Windows\\SysWOW64\\wbem\\WmiPrvSE.exe'))
+| where (ParentProcessName !contains ':\\Windows\\System32\\services.exe' and NewProcessName !contains ':\\Windows\\servicing\\TrustedInstaller.exe')
+| where toupper(Computer) contains v_Host_HostName or toupper(WorkstationName) contains v_Host_HostName
```
shainw: I agree normally, but I am not sure it is more expensive here in this scenario. The query now is looking at all the events from all the computers in the table to exclude these NewProcessName using contains, whereas before it was matching the HostName first and then only iterating over the NewProcessName. In general, I think HostName is probably a smaller index, but maybe not.

Exploration Queries/SecurityEvent/ParentProcessesOnHost.txt
laithhisham: 👍

Exploration Queries/SecurityEvent/ProcessesOnHost.txt
laithhisham: true

PR #169, opened 2019-05-29
Added missing fields for detections

ashwin-patil:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/CommonSecurityLog/TimeSeriesAnomaly-MultiVendor_NetworkTraffic.txt
+++ b/Detections/CommonSecurityLog/TimeSeriesAnomaly-MultiVendor_NetworkTraffic.txt
@@ -6,6 +6,16 @@
 //
 // Id: 06a9b845-6a95-4432-a78b-83919b28c375
 //
+// Severity: Medium
+//
+// QueryFrequency: 1h
+//
+// QueryPeriod: 30d
+//
+// AlertTriggerOperator: gt
+//
+// AlertTriggerThreshold: 0
+//
 // DataSource: #CommonSecurityLog
 //
 // Tactics: #Exfiltration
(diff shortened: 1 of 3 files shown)


Detections/SecurityEvent/TimeSeriesAnomaly-ProcessExecutions.txt
```diff
@@ -6,6 +6,16 @@
...
 //
+// Severity: Medium
+//
+// QueryFrequency: 1h
+//
+// QueryPeriod: 30d
+//
+// AlertTriggerOperator: gt
+//
+// AlertTriggerThreshold: 0
```
shainw: I think you want this to be 5, as that is your value in the script - let TotalEventsThreshold = 5;

Detections/CommonSecurityLog/TimeSeriesAnomaly-MultiVendor_NetworkTraffic.txt
```diff
@@ -6,6 +6,16 @@
...
 //
+// Severity: Medium
+//
+// QueryFrequency: 1h
+//
+// QueryPeriod: 30d
+//
+// AlertTriggerOperator: gt
+//
+// AlertTriggerThreshold: 0
```
shainw: Maybe this should be 3 as that is what you are using in the series_decompose_anomalies function.

ashwin-patil: Agreed per discussion to match scorethreshold with alerttriggerthreshold. 
Submitted changes to both files where scorethreshold is used.

Detections/SecurityEvent/TimeSeriesAnomaly-ProcessExecutions.txt
ashwin-patil: event threshold 5 used here to exclude events with low number of process executions count to reduce false positives however it is ambiguous in this case. Per discussion, we can wait to see how alerts are going to be configured using threshold to make changes here.

PR #171, opened 2019-06-05
parser update usage and new field autocomplete

ashwin-patil:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Parsers/SysmonParser.txt
+++ b/Parsers/SysmonParser.txt
@@ -1,12 +1,15 @@
 // KQL Sysmon Event Parser
-// Date: June 4, 2019
+// Last Updated Date: June 4, 2019
+// Sysmon Version : 9.1.0.0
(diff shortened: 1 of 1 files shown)


Parsers/SysmonParser.txt
```diff
@@ -1,12 +1,16 @@
 // KQL Sysmon Event Parser
-// Date: May 8, 2019
+// Date: June 4, 2019
 // Notes:
 // 1. This parser works against the current version of events from Sysmon, it may need updates if Sysmon is updated.
```
shainw: Lets add in the version of Sysmon.exe this came from so it is more clear.  This came from version 9.1.0.0

ashwin-patil: sure, added version and xml config reference.

TheRealEliteOwl:
One of swift's forks has some really good references to MITRE Techniques.
https://github.com/ion-storm/sysmon-config/blob/master/sysmonconfig-export.xml

Which might be useful with sentinel and MISP configurations.

PR #174, opened 2019-06-06
Update Office Mailbox forwarding hunt query

petebryan:
Updated the query to take into account other logs form O365 that mail forwarding rule creation may appear in.

Fixes #
None

## Proposed Changes
Improvement to scope of hunt query

### DIFF
--- a/Hunting Queries/OfficeActivity/OfficeMailForwarding_hunting.txt
+++ b/Hunting Queries/OfficeActivity/OfficeMailForwarding_hunting.txt
@@ -9,9 +9,9 @@
 // Techniques: #Exfiltration
 //
 OfficeActivity
+| where TimeGenerated >= ago(30d)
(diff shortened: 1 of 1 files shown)


Hunting Queries/OfficeActivity/OfficeMailForwarding_hunting.txt
```diff
@@ -9,10 +9,11 @@
 // Techniques: #Exfiltration
 //
 OfficeActivity
-| where Operation == "Set-Mailbox"
+| where (Operation == "Set-Mailbox" and Parameters contains 'ForwardingSmtpAddress') 
+    or (Operation == 'New-InboxRule' and Parameters contains 'ForwardTo')
 | where TimeGenerated >= ago(30d)
```
shainw: Only thought is to put the Time clause above the where clause.  This could help Perf on a large data set.

petebryan: Good point, change made.

PR #180, opened 2019-06-21
New Office Mailbox Forwarding Parser

petebryan:
Added new Office Mailbox Forwarding Parser to support blog:
https://techcommunity.microsoft.com/t5/Azure-Sentinel/Using-KQL-functions-to-speed-up-analysis-in-Azure-Sentinel/ba-p/712381


shainw:
Maybe write this as a function...like Ashwin did for my Sysmon Parser.

petebryan:
Changed this from a parser to a function. Essentially the same thing but tweak to the working and moved folder.

oshezaf:
Hi @petebryan ,

I don't know if we have official guidelines around this, but I think that a parser should not include a time range. It would be up to the user using the parser as part of their query to specify a time range. 

I recall for the Sysmon parser there was initially a time range built in and it was removed.

~ Ofer

petebryan:
As per @oshezaf time range has been removed from the query.

PR #196, opened 2019-06-27
Adding in check for okhttp

shainw:
Fixes #

## Proposed Changes

  -
  -
  -


Hunting Queries/MultipleDataSources/UseragentExploitPentest.txt
```diff
@@ -27,7 +27,8 @@ let UserAgentList = dynamic(["Internet Explorer ",
 "X-FORWARDED-FOR",
 "DotDotPwn v2.1",
 "SIPDROID","wordpress hash grabber",
-"exploit"]);
+"exploit",
+"okhttp/"]);
```
timbMSFT: given the volume, do we want to summarize the results?

PR #215, opened 2019-07-30
Python modules update for GA

zhzhao8888:
Fixes #

## Proposed Changes

1. Remove dependency on DSVM + Mono, instead, using Python encryption  for Anomaly Lookup
2. Log utility is in the check in list, not used in the code, but can be used for our debugging during development

  -
  -
  -

### DIFF
--- a/Notebooks/SentinelUtilities/SentinelLog/log.py
+++ b/Notebooks/SentinelUtilities/SentinelLog/log.py
@@ -1,0 +1,38 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+"""
+log module:
+This module provides log functionalities through Azure Application Insights
+"""
+
+
+from applicationinsights import TelemetryClient
+from SentinelUtils.crypto_utility import CryptoUtility
+
+
+class Log:
+    """ The class performs log action """
+
+    def __init__(self):
+        self.key = b'IVnGT69i43R1i6qokpVxIx_tE2MyBlMebu4yJfJh1Ow='
+        # pylint: disable=line-too-long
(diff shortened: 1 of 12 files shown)


Notebooks/SentinelUtilities/SentinelLog/log.py
```diff
@@ -0,0 +1,38 @@
...
+    def __init__(self):
+        self.key = b'IVnGT69i43R1i6qokpVxIx_tE2MyBlMebu4yJfJh1Ow='
+        # pylint: disable=line-too-long
+        self.ai_key = b'gAAAAABdN3UmwZJHHxTybj0KRbKNehSo55DZ5Bi2QtohQsrEgy-WZNWDQAETVnaeJbQ4S0ltZLkebi_hhueaxl_uxYE5HheuB0ZFobq1IzgE163jUsjSLqilcqzy_uLKkSTYHAxYNLxL'
+
+    def log(self, telemetry_instance):
+        """ Log Telemetry instance """
+
+        crypto = CryptoUtility(self.key)
+        client = TelemetryClient(crypto.decrypt_text(self.ai_key))
```
ianhelle: I think to avoid bad publicity it would be better to use terms like obfuscate, deobfuscate, obfuscation_seed. etc. Otherwise someone is going to find the code with an embedded key and tweet out why we don't understand security.

Notebooks/SentinelUtilities/SentinelUtils/crypto_utility.py
```diff
@@ -0,0 +1,37 @@
...
+    def __init__(self, key):
+        self.key = key
+
+    @staticmethod
+    def generate_key():
+        """ Generate key """
+
+        return Fernet.generate_key()
+
+    def encrypt_text(self, text):
```
ianhelle: I think to avoid bad publicity it would be better to use terms like obfuscate, deobfuscate, obfuscation_seed. etc.

PR #220, opened 2019-08-13
Change Detection file format from txt to yaml

sagamzu:


juliango2100:
Renaming to Yaml.

juliango2100:
Changing to yaml extension

PR #229, opened 2019-08-19
Signin event correlation rules

petebryan:
Correlation rules between AAD <> hosts as well as AAD  <> AWS.
Set signin threshold at 5 as appears to be best balance across available data sets. Customers using proxy services will need to change this or explicitly exclude certain IPs before using these detections.

### DIFF
--- a/Detections/MultipleDataSources/AADAWSConsoleCorrelation.yaml
+++ b/Detections/MultipleDataSources/AADAWSConsoleCorrelation.yaml
@@ -1,5 +1,5 @@
 id: 643c2025-9604-47c5-833f-7b4b9378a1f5
-name: IP with multiple failed AAD logins successfully logs in to the AWS Console.
+name: IP with multiple failed Azure AD logins successfully logs in to the AWS Console.
 description: |
 	This query creates a list of IP addresses with a number failed login attempts to AAD 
 	above a set threshold.  It then looks for any successful AWS Console logins from any
--- a/Detections/MultipleDataSources/AADAWSConsoleCorrelation.yaml
+++ b/Detections/MultipleDataSources/AADAWSConsoleCorrelation.yaml
@@ -1,0 +1,40 @@
+id: 643c2025-9604-47c5-833f-7b4b9378a1f5
+name: IP with multiple failed AAD logins successfully logs in to the AWS Console.
+description: |
+	This query looks for IP addresses with multiple failed login attempts to AAD 
+	with a successful AWS Console login within the same timeframe.
+severity: Medium
+requiredDataConnectors:
+  - connectorId: AzureActiveDirectory
+    dataTypes:
+     - SigninLogs
+  - connectorId: AWS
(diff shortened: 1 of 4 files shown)


petebryan:
Feedback very much appreciated on these. I tried to incorporate other data-sets to support these and/or include some more complex logic but these queries were the ones that ended up being the reliable detections that would work in a range of environments.

Detections/MultipleDataSources/AADHostLoginCorrelation.yaml
```diff
@@ -0,0 +1,51 @@
...
+	| where ResultType !in ("0", "50125", "50140")
+	| summarize count() by IPAddress
+	| where count_ > signin_threshold
+	| summarize makelist(IPAddress);
+	let linux_logons =
+	Syslog
+	| where TimeGenerated >= ago(1d)
+	| where Facility contains "auth" and ProcessName != "sudo"
+	| where SyslogMessage has "Accepted"
+	| where SyslogMessage in (suspicious_signins)
```
ianhelle: Is the IPAddress the whole syslogmessage? There's a gnarly way look for items in a list inside something else ISTR but you have to join each ip address entry to you target data and do message contains(ip)

ianhelle: or use a regex to extract the ip

petebryan: Thanks, yes meant to use has_any for the syslog look up, have fixed this now.

Detections/MultipleDataSources/AADAWSConsoleCorrelation.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - InitialAccess
+  
+query: |
+
+	let signin_threshold = 5; 
+	let suspicious_signins = 
+	SigninLogs
+	| where TimeGenerated >= ago(1d)
+	| where ResultType !in ("0", "50125", "50140")
+	| summarize count() by IPAddress
```
shainw: Since we are dealing with IP's, do we want to exclude 127.0.0.1.  Applies to all 4 queries here.

petebryan: Good suggestion, I had done this in development when also trying to bring in IPs in security alerts (which in the end was too noisy) but hadn't seen too many issues with sign ins in the data I had but definitely worth excluding still

Detections/MultipleDataSources/AADHostLoginCorrelation.yaml
petebryan: Had to extract the IP in the end to support the entity mapping in the new standards so fell back to in.

petebryan:
> Couple of new syntax items we want to add to items in the Detections folder. Will handle offline due to length of comments.

Should be done now.

PR #245, opened 2019-08-22
Azure activity diag format update

shainw:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/AzureActivity/NewResourceGroupsDeployedTo.yaml
+++ b/Detections/AzureActivity/NewResourceGroupsDeployedTo.yaml
@@ -1,32 +1,35 @@
 ﻿id: 9fb57e58-3ed8-4b89-afcf-c8e786508b1c
-name: Common deployed resources
+name: Suspicious Resource deployment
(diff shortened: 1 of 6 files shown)


Detections/AzureActivity/NewResourceGroupsDeployedTo.yaml
```diff
@@ -1,32 +1,35 @@
 ﻿id: 9fb57e58-3ed8-4b89-afcf-c8e786508b1c
-name: Common deployed resources
+name: Suspicious Resource deployment
 description: |
-  this query looks for common deployed resources (resource name and resource groups). Alerts is
-  generated for new ResourceGroups first seen associated with deployment operations.
+  'Identifies when a rare Resource and ResourceGroup deployment occurs by a previously unseen Caller.'
```
sagamzu: nit:
I wrote it in another PR. 
I think the decription should be wrapped with ' character only when : character use in the description, otherwise it could be used but not necessary.

PR #258, opened 2019-08-23
Format updates for Syslog

shainw:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/Syslog/FailedLogonAttempts_UnknownUser.yaml
+++ b/Detections/Syslog/FailedLogonAttempts_UnknownUser.yaml
@@ -1,41 +1,40 @@
 ﻿id: e7ec9fa6-e7f7-41ed-a34b-b956837a3ee6
 name: Failed Logon Attempts
 description: |
-  This query uses the Syslog, authpriv logs to show when there are failed logon attempts from unknown users.
-  In this case, the unknown user means the account that tried to log in isn't provisioned on the machine. A 
-  few hits could indicate someone attempting to access a machine they aren't authorized on. If there are a lot
-  of hits, especially from outside your network, it could indicate a brute force attack. The query looks at 
-  the last seven (7) days by default.
+  'Identifies failed logon attempts from unknown users in Syslog authpriv logs. The unknown user means the account that tried to log in 
+  isn't provisioned on the machine. A few hits could indicate someone attempting to access a machine they aren't authorized to access. 
+  If there are many of hits, especially from outside your network, it could indicate a brute force attack. 
+  Default threshold for logon attempts is 6.'
 severity: Medium
 requiredDataConnectors:
   - connectorId: Syslog
     dataTypes: 
       - Syslog
-queryFrequency: 24h
-queryPeriod: 24h
+queryFrequency: 1d
+queryPeriod: 1d
 triggerOperator: gt
 triggerThreshold: 0
 tactics:
-  - InitialAccess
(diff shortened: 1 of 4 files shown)


Detections/Syslog/FailedLogonAttempts_UnknownUser.yaml
```diff
@@ -1,41 +1,40 @@
...
   | where SyslogMessage contains "authentication failure" and SyslogMessage contains " uid=0"
   | parse SyslogMessage with * "rhost=" ExternalIP
   | project TimeGenerated, Computer, ProcessName, HostIP, ExternalIP, ProcessID;  
   // Below pulls messages from syslog-authpriv logs that show each instance an unknown user tried to logon. 
   let userfail = Syslog 
   | where Facility == "authpriv" 
-  | where TimeGenerated >= startdate
+  | where TimeGenerated >= ago(startdate)
   | where SyslogMessage contains "user unknown"
   | project TimeGenerated, Computer, HostIP, ProcessID;
```
ms-bnick: Just checking, is it possible to also pull out the user account or set of accounts that the remote attacker was targeting? Is that even available in syslog (pardon, I am no syslog expert)? If only one account is targeted, it seems like a perfect candidate for creating an account entity.

Detections/Syslog/ssh_potentialBruteForce.yaml
```diff
@@ -1,23 +1,30 @@
...
 
+  let timeframe = 1d;
+  let threshold = 15;
   Syslog
+  | where TimeGenerated >= ago(timeframe)
   | where SyslogMessage contains "Failed password for invalid user"
   | where SyslogMessage contains "ssh2" 
   | parse kind=relaxed SyslogMessage with * "invalid user" user " from " ip " port" port " ssh2"
   | project user, ip, port, SyslogMessage, EventTime
   | summarize count() by ip, bin(EventTime, 4h)
```
ms-bnick: maybe include the user info? 
```suggestion
  | summarize count(), makeset(user) by ip, bin(EventTime, 4h)
```

Detections/Syslog/ssh_potentialBruteForce.yaml
```diff
@@ -1,23 +1,30 @@
...
-  | where count_ > 15
\ No newline at end of file
+  | where count_ > threshold
+  | extend timestamp = EventTime, IPCustomEntity = ip
```
ms-bnick: Perhaps include a user entity if there is only one account?  I think I got the syntax right, but I don't know which workspace I should test this against. 
```suggestion
  | extend timestamp = EventTime, IPCustomEntity = ip, AccountCustomEntity = iff(array_length(set_user) == 1, setuser[0], "")
```

Detections/Syslog/FailedLogonAttempts_UnknownUser.yaml
shainw: I was mainly doing cleanup, we can review about adding in User, but I did not author this and the author is on extended leave.  I will review this in more detail next week, but this PR is mainly about cleaning up metadata and some best practices in the query syntax.

Detections/Syslog/ssh_potentialBruteForce.yaml
shainw: Same as other comment, will look at next week.  we will not add this change yet

PR #288, opened 2019-08-30
Updating entities for detections in officeactivity thru IIS

shainw:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/OfficeActivity/Office_MailForwarding.yaml
+++ b/Detections/OfficeActivity/Office_MailForwarding.yaml
@@ -20,7 +20,7 @@ relevantTechniques:
   - T1020
 query: |
 
-  let timeframe = 7d
+  let timeframe = 7d;
   OfficeActivity
   | where TimeGenerated >= ago(timeframe)
   | where Operation =~ "Set-Mailbox"
@@ -28,9 +28,9 @@ query: |
   | extend parsed = parse_json(Parameters)
   | extend parameterName = parsed[1].Name, fwdingDestination = tostring(parsed[1].Value)
   | where isnotempty(fwdingDestination)
-  | summarize TimeGenerated=max(TimeGenerated) by UserId, fwdingDestination
-  | summarize TimeGenerated=max(TimeGenerated), userCount = dcount(UserId), UserId = makelist(UserId) by fwdingDestination
+  | summarize TimeGenerated=max(TimeGenerated) by UserId, ClientIP, fwdingDestination
+  | summarize TimeGenerated=max(TimeGenerated), userCount = dcount(UserId), UserId = makelist(UserId), ClientIP = makeset(ClientIP) by fwdingDestination
   | where userCount > 1
-  | mvexpand UserId
-  | extend timestamp = TimeGenerated, AccountCustomEntity = UserId
+  | mvexpand UserId, ClientIP
+  | extend timestamp = TimeGenerated, AccountCustomEntity = UserId, IPCustomEntity = ClientIP 
 
--- a/Detections/SigninLogs/DisabledAccountSigninsAcrossManyApplications.yaml
+++ b/Detections/SigninLogs/DisabledAccountSigninsAcrossManyApplications.yaml
@@ -26,6 +26,7 @@ query: |
   | where TimeGenerated >= ago(timeframe)
   | where ResultType == "50057"
   | where ResultDescription =~ "User account is disabled. The account has been disabled by an administrator."
-  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), count(), applicationCount = dcount(AppDisplayName), applicationSet = makeset(AppDisplayName) by UserPrincipalName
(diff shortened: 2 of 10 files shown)


Detections/OfficeActivity/Office_MailForwarding.yaml
```diff
@@ -20,17 +20,17 @@ relevantTechniques:
...
   OfficeActivity
   | where TimeGenerated >= ago(timeframe)
   | where Operation =~ "Set-Mailbox"
   | where Parameters contains "ForwardingSmtpAddress"
   | extend parsed = parse_json(Parameters)
   | extend parameterName = parsed[1].Name, fwdingDestination = tostring(parsed[1].Value)
   | where isnotempty(fwdingDestination)
-  | summarize TimeGenerated=max(TimeGenerated) by UserId, fwdingDestination
-  | summarize TimeGenerated=max(TimeGenerated), userCount = dcount(UserId), UserId = makelist(UserId) by fwdingDestination
+  | summarize TimeGenerated=max(TimeGenerated) by UserId, ClientIP, fwdingDestination
```
timbMSFT: not sure we want to do this: this will explode the number of results if there are many IPs?

Detections/SigninLogs/DisabledAccountSigninsAcrossManyApplications.yaml
```diff
@@ -26,6 +26,7 @@ query: |
   | where TimeGenerated >= ago(timeframe)
   | where ResultType == "50057"
   | where ResultDescription =~ "User account is disabled. The account has been disabled by an administrator."
-  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), count(), applicationCount = dcount(AppDisplayName), applicationSet = makeset(AppDisplayName) by UserPrincipalName
+  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), count(), applicationCount = dcount(AppDisplayName),
```
timbMSFT: same comment as above: if the IP is _already_ a column then let's ensure it's extracted as an entity. But we shouldn't change the queries to return 1 row per IP.

Detections/SigninLogs/SigninBruteForce-AzurePortal.yaml
```diff
@@ -33,8 +33,9 @@ query: |
...
   // Split out failure versus non-failure types
   | extend FailureOrSuccess = iff(ResultType in ("0", "50125", "50140"), "Success", "Failure")
-  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), makeset(IPAddress), makeset(OS), makeset(Browser), makeset(City), 
+  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), IPAddress = makeset(IPAddress), makeset(OS), makeset(Browser), makeset(City), 
   makeset(ResultType), FailureCount = countif(FailureOrSuccess=="Failure"), SuccessCount = countif(FailureOrSuccess=="Success") 
   by bin(TimeGenerated, authenticationWindow), UserDisplayName, UserPrincipalName, AppDisplayName
   | where FailureCount >= failureCountThreshold and SuccessCount >= successCountThreshold
-  | extend timestamp = StartTimeUtc, AccountCustomEntity = UserPrincipalName
+  | mvexpand IPAddress
+  | extend timestamp = StartTimeUtc, AccountCustomEntity = UserPrincipalName, IPCustomEntity = IPAddress
```
timbMSFT: i don't think IPCustomEntity handles sets?

Detections/OfficeActivity/Office_MailForwarding.yaml
```diff
@@ -20,17 +20,17 @@ relevantTechniques:
...
+  | summarize TimeGenerated=max(TimeGenerated), userCount = dcount(UserId), UserId = makelist(UserId), ClientIP = makeset(ClientIP) by fwdingDestination
```
timbMSFT: this could be a large set / and don't think entity extraction handles sets anyway?

Detections/OfficeActivity/Office_MailForwarding.yaml
shainw: Tested on more than one workspace with varied data, we should be good here.

Detections/SigninLogs/DisabledAccountSigninsAcrossManyApplications.yaml
shainw: Tested on more than one workspace with varied data, we should be good here.

Detections/SigninLogs/SigninBruteForce-AzurePortal.yaml
shainw: we use mvexpand to bring this back to single IPAddress per line.  Tested on more than one workspace with varied data, we should be good here.

Detections/OfficeActivity/Office_MailForwarding.yaml
shainw: We use mvexpand below to break out into single instances, Tested on more than one workspace with varied data, we should be good here.

PR #289, opened 2019-09-01
Change to YAML and Entity upd hunting audit thru azure activity

shainw:
Fixes #

## Proposed Changes

  -
  -
  -


timbMSFT:
Assuming that there are no functional changes then this is good to go. Per other PR we should make sure that we are not creating more result rows by breaking the results out by IP for example

PR #291, opened 2019-09-01
DNS Hunting query updates for YAML and entities

shainw:
…l and fixing up entity output

Fixes #

## Proposed Changes

  -
  -
  -


timbMSFT:
Assuming that there are no functional changes then this is good to go. Per other PR we should make sure that we are not creating more result rows by breaking the results out by IP for example -- some of these queries will be very sensitive to such a change.

ms-bnick:
Looks good to me.  I don't know what the basket() function does.  Assuming it generates timestamps and entities, I am good with this.

PR #293, opened 2019-09-03
Hunting - Multidata - updating to yaml and adding entities

shainw:
Fixes #

## Proposed Changes

  -
  -
  -


Hunting Queries/MultipleDataSources/PermutationsOnLogonNames.yaml
```diff
@@ -0,0 +1,115 @@
...
+  ) on $left.CharSet == $right.CharSet and $left.TimeGenerated == $right.TimeGenerated 
+  | project-away TimeGenerated1, CharSet1 
+  // Make sure that we get more than a single match for First or Last
+  | where UserNameMatchOnFirstCount >= UserNameMatchMin or UserNameMatchOnLastCount >= UserNameMatchMin
+  | where FailedLogonCountForFirst >= FailedLogonMin or FailedLogonCountForLast >= FailedLogonMin
+  ; 
+  (union isfuzzy=true char4, char6, char8) 
+  | project Type, TimeGenerated, CharSet, UserNameMatchOnFirst, UserNameMatchOnFirstCount, FailedLogonCountForFirst, UserNameMatchOnLast, UserNameMatchOnLastCount, FailedLogonCountForLast 
+  | sort by UserNameMatchOnFirstCount desc, UserNameMatchOnLastCount desc
+  | extend timestamp = TimeGenerated
```
ms-bnick: Maybe add AccountCustomEntity for the target account(s?)?

shainw: This is a complex query which I happened to write, I don't want to mess with it as part of these changes.  I will see if we can do this in the future.

PR #314, opened 2019-09-10
Avoid false positives on AD FS

pasilva-msft:
Added one line to exclude events from "AD FS Auditing" as Event Source Name to avoid generating false positives when detecting this on AD FS servers.

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/SecurityEvent/SecurityEventLogCleared.yaml
+++ b/Detections/SecurityEvent/SecurityEventLogCleared.yaml
@@ -1,7 +1,8 @@
 ﻿id: 80da0a8f-cfe1-4cd0-a895-8bc1771a720e
 name: Security Event log cleared
 description: |
-  'Checks for event id 1102 which indicates the security event log was cleared.'
+  'Checks for event id 1102 which indicates the security event log was cleared. 
+  Exclude events from "AD FS Auditing" as Event Source Name to avoid generating false positives when detecting this on AD FS servers.'
 severity: Low
 requiredDataConnectors:
   - connectorId: SecurityEvents
@@ -17,9 +18,10 @@ relevantTechniques:
   - T1107
 query: |
 
-  let timeframe = 1d;
+let timeframe = 1d;
   SecurityEvent
   | where TimeGenerated >= ago(timeframe)
   | where EventID == 1102
+  | where EventSourceName <> "AD FS Auditing" 
   | summarize count() by StartTimeUtc = TimeGenerated, Computer, Account, EventID, Activity
   | extend timestamp = StartTimeUtc, AccountCustomEntity = Account, HostCustomEntity = Computer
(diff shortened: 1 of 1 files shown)


Detections/SecurityEvent/SecurityEventLogCleared.yaml
```diff
@@ -17,9 +18,10 @@ relevantTechniques:
   - T1107
 query: |
 
-  let timeframe = 1d;
+let timeframe = 1d;
   SecurityEvent
   | where TimeGenerated >= ago(timeframe)
   | where EventID == 1102
+  | where EventSourceName <> "AD FS Auditing"
```
shainw: Probably better to do this just in case there is some other ID collision that could occur - 
| where EventID == 1102 and EventSourceName == "Microsoft-Windows-Eventlog"

pasilva-msft: Fixed detection logic. Thank you!

Detections/SecurityEvent/SecurityEventLogCleared.yaml
```diff
@@ -17,9 +18,9 @@ relevantTechniques:
   - T1107
 query: |
 
-  let timeframe = 1d;
+ let timeframe = 1d;
```
shainw: Something I did not notice before, we lost a space here, this breaks YAML, please add the additional space in front of let.  You can verify format with any online YAML validator, I often use - https://codebeautify.org/yaml-validator/

Detections/SecurityEvent/SecurityEventLogCleared.yaml
```diff
@@ -1,7 +1,8 @@
 ﻿id: 80da0a8f-cfe1-4cd0-a895-8bc1771a720e
 name: Security Event log cleared
 description: |
-  'Checks for event id 1102 which indicates the security event log was cleared.'
+  'Checks for event id 1102 which indicates the security event log was cleared. 
+It uses Event Source Name "Microsoft-Windows-Eventlog" to avoid generating false positives from other sources, like AD FS servers for instance.'
```
shainw: Something I did not notice before, we need 2 spaces here, this breaks YAML, please add the additional space in front of It.  You can verify format with any online YAML validator, I often use - https://codebeautify.org/yaml-validator/

Detections/SecurityEvent/SecurityEventLogCleared.yaml
```diff
@@ -17,9 +18,9 @@ relevantTechniques:
   - T1107
 query: |
```
shainw: Not required, but we like to have an empty line between "query:  |" and the start of the query.

Detections/SecurityEvent/SecurityEventLogCleared.yaml
pasilva-msft: Thank you for that! Validated :)

Detections/SecurityEvent/SecurityEventLogCleared.yaml
pasilva-msft: Also resolved this.

PR #328, opened 2019-09-27
Update DomainEntity_PaloAlto.yaml

srisang:
Update the CommonSecurityLog lookup based on ingestion time.

Fixes #

## Proposed Changes

  -
  -
  -


srisang:
@petebryan , can you please review this change?

petebryan:
Looks good to me. @shainw it might be an idea to update all detection queries with this logic to make sure we don't miss anything in the detection window.

PR #337, opened 2019-10-04
Azure Sentinel Hunting Query to check critical ports opened to entire…

Oscarbralo:
Hunting query to check all critical ports opened to the entire internet. 

(I didn't know if this file is the one to be needed to update, or the .yaml already formed) Please, some advice and i will reupload

Fixes #


Oscarbralo:
> The queries are in YAML format. You can use any of the templates as a reference. Additionally, there is guidance in the WIKI - https://github.com/Azure/Azure-Sentinel/wiki/Contribute-to-Sentinel-GitHub-Community-of-Queries
> You will want to create a new folder under Hunting Queries named AzureDiagnostics as that folder is not created yet.



Already submitted in yaml format.  I have used another hunting query as template as you said. 

The folder i think is created, at least I already created that folder on my branch and submitted to.

Thank you very much

Hunting Queries/AzureDiagnostics/CriticalPortsOpened.yaml
```diff
@@ -0,0 +1,38 @@
+id: 66d494c0-233c-438a-9b1b-5fe839790d38
+name: Check critical ports opened to the entire internet
+description: |
+  'Discover all critical ports from a list having rules like 'Any' for sourceIp, which means that they are opened to everyone. Critial ports shouldn't be opened to everyone, and should be filtered.'
```
shainw: Because there is an extra single quote in "shouldn't", the format is incorrect when being used in Log Analytics directly.  Change "shouldn't" to "should not" and then this issue will not occur.  The YAML validator does not catch this because it is valid YAML, just for when using directly in Log Analytics. :)

Hunting Queries/AzureDiagnostics/CriticalPortsOpened.yaml
```diff
@@ -0,0 +1,38 @@
...
+requiredDataConnectors:
+  - connectorId: AzureDiagnostic
+tactics:
+  - InitialAccess
+query: |
+
+  //Check critical ports opened to the entire internet
+  AzureDiagnostics
+  | where Category == "NetworkSecurityGroupEvent"
```
shainw: There are many hits for "blocks" if not excluded.  I would recommend that you include the following, if you want to leave blocks in, I would include it, comment it out and then put a note above it about when and how to use it.

| where type_s !~ "block"

Hunting Queries/AzureDiagnostics/CriticalPortsOpened.yaml
```diff
@@ -0,0 +1,38 @@
...
+  ,"27017","27017-27017"//MongoDB
+  ,"20","20-20"         //FTP
+  ,"21","21-21"         //FTP
+  ,"445","445-445"      //Active Directory
+  ,"161","161-161"      //SNMP
+  ,"25","25-25"         //SMTP
+  )
+   or (conditions_destinationPortRange_s == "0-65535" and conditions_sourcePortRange_s == "0-65535")
+  | where priority_d < 65000    //Not to check the Azure defaults
+  | where conditions_sourceIP_s == "0.0.0.0/0,0.0.0.0/0" or conditions_sourceIP_s == "0.0.0.0/0" //With rules Any/Any
```
shainw: I would suggest a summarize is included so that the result set it smaller when we are looking at duplicate events over time.  The following will keep the time frame of events based on first and last time seen, but provide a much smaller result set along with a count of how many times it occurred.

  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), count() by OperationName, systemId_g, vnetResourceGuid_g, subnetPrefix_s, macAddress_s, primaryIPv4Address_s, ruleName_s, 
  direction_s, priority_d, type_s, conditions_destinationIP_s, conditions_destinationPortRange_s, conditions_sourceIP_s, conditions_sourcePortRange_s, ResourceId

Hunting Queries/AzureDiagnostics/CriticalPortsOpened.yaml
Oscarbralo: done

Hunting Queries/AzureDiagnostics/CriticalPortsOpened.yaml
Oscarbralo: done

Hunting Queries/AzureDiagnostics/CriticalPortsOpened.yaml
Oscarbralo: done

PR #346, opened 2019-10-15
Change SID regex

pasilva-msft:
Changed the variable "WellKnownGroupSID" to include two more SID/RID (-498 and -1000) and a $ sign on the first SID/RID regex, since it was generating false positives when last number has more than two number after 5. Example: S-1-5-21-1111111111-2222222222-333333333-550390.

Fixes #

## Proposed Changes

  -
  -
  -


pasilva-msft:
This is duplicated from #344

PR #347, opened 2019-10-16
fixed typo from mas.util to mas.utils

yoshiomura:
Fixes #

## Proposed Changes
fixed wrong imported module name from mas.util to mas.utils


ianhelle:
We'll also fix this one to match current msticpy namespace re-organization.

yoshiomura:
@ianhelle thanks for checking !!

PR #349, opened 2019-10-16
fixed typo from mas.util to mas.utils

yoshiomura:
Fixes #

fixed typo from mas.util to mas.utils

  -
  -
  -


ianhelle:
@yoshiomura  - thanks for the bug detection. We are going to revamp/update these notebooks in the next 2 weeks so I'll make sure that we sweep this up.

PR #355, opened 2019-10-21
RDP lateral movement detections

shainw:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/SecurityEvent/RDP_MultipleConnectionsFromSingleSystem.yaml
+++ b/Detections/SecurityEvent/RDP_MultipleConnectionsFromSingleSystem.yaml
@@ -18,13 +18,29 @@ relevantTechniques:
   - T1076
 query: |
 
-  let timeframe = 1d;
-  // The threshold below includes matching on computer counts by account higher than 3 in a given day.  Change the threshold as needed.
-  let threshold = 3;
+  let endtime = 1d;
+  let starttime = 8d;
+  let avgDays = 7;
   SecurityEvent
-  | where TimeGenerated >= ago(timeframe) 
+  | where TimeGenerated >= ago(starttime) and TimeGenerated < ago(endtime) 
   | where EventID == 4624 and LogonType == 10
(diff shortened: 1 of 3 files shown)


Detections/SecurityEvent/RDP_MultipleConnectionsFromSingleSystem.yaml
```diff
@@ -0,0 +1,30 @@
...
+triggerThreshold: 0
+tactics:
+  - LateralMovement
+relevantTechniques:
+  - T1076
+query: |
+
+  let timeframe = 1d;
+  // The threshold below includes matching on computer counts by account higher than 3 in a given day.  Change the threshold as needed.
+  let threshold = 3;
```
petebryan: Given this figure will vary a lot per environment could this be more effectively implemented by doing some baselining? Say if the number of connections is higher than the average in the last week?

Detections/SecurityEvent/RDP_Nesting.yaml
```diff
@@ -0,0 +1,39 @@
+id: 69a45b05-71f5-45ca-8944-2e038747fb39
+name: RDP Nesting
+description: |
+  'Identifies when an RDP connection is made to a first system and then an RDP connection is made from the first system 
+  to another system with the same account within the same day.  
+  RDP connections are indicated by the logged EventID 4624 with LogonType = 10'
+severity: Medium
```
petebryan: Given there are quite a few legitimate use cases for this situation (granted same account name but still fairly common I would guess) should this be low rather than medium?

Detections/SecurityEvent/RDP_MultipleConnectionsFromSingleSystem.yaml
shainw: Done

Detections/SecurityEvent/RDP_Nesting.yaml
shainw: Done

PR #361, opened 2019-10-22
Adding in Resource Creation with Network Activity and fixing type in …

shainw:
…detection in Azure Activity

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Hunting Queries/MultipleDataSources/AzureResourceCreationWithNetworkActivity.yaml
+++ b/Hunting Queries/MultipleDataSources/AzureResourceCreationWithNetworkActivity.yaml
@@ -1,0 +1,99 @@
+id: ac25d05d-362d-4a8d-b4e7-58c0edd2379c
+name: Anomalous Resource Creation and related Network Activity
+description: |
+  'Indicates when an anomalous number of resources are created successfully in Azure via the AzureActivity log.
+  This is then joined with the AzureNetworkAnalytics_CL data to identify any network related activity for the created resource.
+  The anomaly detection identifies activities that have occured both since the start of the day 1 day ago and the start of the day 7 days ago.
+  The start of the day is considered 12am UTC time.
+  Resources: 
+  https://docs.microsoft.com/en-us/azure/azure-monitor/insights/azure-networking-analytics
+  https://docs.microsoft.com/en-us/azure/network-watcher/traffic-analytics-schema'
+requiredDataConnectors:
+  - connectorId: AzureActivity
+    dataTypes:
+      - AzureActivity
+  - connectorId: AzureNetworkWatcher
+    dataTypes:
+      - AzureNetworkAnalytics_CL
+tactics:
+  - Impact
+relevantTechniques:
+  - T1496
+query: |
+
+  let starttime = 7d;
+  let endtime = 1d;
+  let Activity = AzureActivity
+  | where TimeGenerated >= startofday(ago(starttime))
+  // We look for any Operation that created and then succeeded where ActivitySubStatus has a value so that we can provide context
+  | where OperationName has "Create"
+  | where ActivityStatus has "Succeeded"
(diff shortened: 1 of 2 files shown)


Hunting Queries/MultipleDataSources/AzureResourceCreationWithNetworkActivity.yaml
```diff
@@ -0,0 +1,99 @@
+id: ac25d05d-362d-4a8d-b4e7-58c0edd2379c
+name: Anomalous Resource Creation and related Network Activity
+description: |
+  'Indicates when an anomalous number of resources are created successfully in Azure via the AzureActivity log.
```
petebryan: Maybe add a why to the description as well as the what. If this fires an alert it might be useful for the analyst to understand why this is bad.

Hunting Queries/MultipleDataSources/AzureResourceCreationWithNetworkActivity.yaml
```diff
@@ -0,0 +1,99 @@
...
+  let endtime = 1d;
+  let Activity = AzureActivity
+  | where TimeGenerated >= startofday(ago(starttime))
+  // We look for any Operation that created and then succeeded where ActivitySubStatus has a value so that we can provide context
+  | where OperationName has "Create"
+  | where ActivityStatus has "Succeeded"
+  | make-series dResourceCount=dcount(ResourceId) default=0 on EventSubmissionTimestamp in range(startofday(ago(7d)), now(), 1d) by Caller, Resource, OperationName
+  | extend (RSquare,Slope,Variance,RVariance,Interception,LineFit)=series_fit_line(dResourceCount)
+  // Comment slope reference below to see all returns
+  //| where Slope > 0.2
```
petebryan: Is the slope value something we could set in a let statement to make it easier for people to tune the value if needed?

Hunting Queries/MultipleDataSources/AzureResourceCreationWithNetworkActivity.yaml
```diff
@@ -0,0 +1,99 @@
...
+  | join kind=leftsemi (
+  // Last day's activity is anomalous
+  AzureActivity
+  | where TimeGenerated >= startofday(ago(endtime))
+  // We look for any Operation that created and then succeeded where ActivitySubStatus has a value so that we can provide context
+  | where OperationName has "Create"
```
petebryan: Do you want to filter out resources you know will definitely create network activity such as network interface creation, availability set changes or LB updates? The anomaly detection might mean these are not an issue but I could see it potentially FPing.

Hunting Queries/MultipleDataSources/AzureResourceCreationWithNetworkActivity.yaml
petebryan: Also consider un-commenting this line by default to make its default state quieter.

Hunting Queries/MultipleDataSources/AzureResourceCreationWithNetworkActivity.yaml
shainw: It is hunting, but will think on a bit more details in the description

Hunting Queries/MultipleDataSources/AzureResourceCreationWithNetworkActivity.yaml
shainw: That was an accident, should have been uncommented.

Hunting Queries/MultipleDataSources/AzureResourceCreationWithNetworkActivity.yaml
shainw: This is hunting, so again we are building context for a potential investigation.

PR #364, opened 2019-10-28
Including password change and reset tracking across multiple data sou…

shainw:
…rces

Fixes #

## Proposed Changes

  -
  -
  -


ashwin-patil:
Similar hunting query exists with overlapping logic, although has some differences on trigger criteria and datasources list. Do we want to keep both or merge with differences ?  https://github.com/Azure/Azure-Sentinel/blob/master/Hunting%20Queries/MultipleDataSources/MultiplePasswordresetsbyUser.yaml .

PR #390, opened 2019-11-26
Change print in rsyslog config

Ronmarsiano:
We used to check correctly but print content for local4.
now the content will be printed correctly 
  -
  -


morshabi:
Approved 😊

PR #406, opened 2019-12-16
Updating Process entropy and adding in rare process path

shainw:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Hunting Queries/SecurityEvent/RareProcessPath.yaml
+++ b/Hunting Queries/SecurityEvent/RareProcessPath.yaml
@@ -12,7 +12,7 @@ tactics:
 query: |
 
   let end = startofday(now());
-  let start = end - 7d;
+  let start = end - 8d;
   let processEvents=
   SecurityEvent
   | where TimeGenerated >= start and TimeGenerated <= end
@@ -22,7 +22,7 @@ query: |
   | where ParentProcessName !endswith ":\\Windows\\System32\\wuauclt.exe" and NewProcessName !startswith "C:\\Windows\\SoftwareDistribution\\Download\\Install\\AM_Delta_Patch_"
   | where NewProcessName !has ":\\Windows\\WinSxS\\amd64_microsoft-windows-servicingstack_" and ParentProcessName !has ":\\Windows\\WinSxS\\amd64_microsoft-windows-servicingstack_"
   | where NewProcessName !endswith ":\\WindowsAzure\\SecAgent\\WaSecAgentProv.exe" 
-  //| where ParentProcessName !has ":\\WindowsAzure\\GuestAgent_" and NewProcessName !has ":\\WindowsAzure\\GuestAgent_"
+  | where ParentProcessName !has ":\\WindowsAzure\\GuestAgent_" and NewProcessName !has ":\\WindowsAzure\\GuestAgent_"
   | where ParentProcessName !has ":\\WindowsAzure\\WindowsAzureNetAgent_" and NewProcessName !has ":\\WindowsAzure\\WindowsAzureNetAgent_"
   | where ParentProcessName !has ":\\ProgramData\\Microsoft\\Windows Defender\\platform\\" and ParentProcessName !endswith "\\MpCmdRun.exe" 
   | where NewProcessName !has ":\\ProgramData\\Microsoft\\Windows Defender\\platform\\" and NewProcessName !endswith "\\MpCmdRun.exe" 
@@ -51,7 +51,7 @@ query: |
   | project TimeGenerated, Computer, Account, Process, ProcessPath, 
   NormalizedProcessPath = replace("[0-9A-Fa-f]{8}[-][0-9A-Fa-f]{4}[-][0-9A-Fa-f]{4}[-][0-9A-Fa-f]{4}[-][0-9A-Fa-f]{12}", "<guid>", NormalizedProcessPath)
   // normalize digits away
-  | project TimeGenerated, Computer, Account, Process, ProcessPath, NormalizedProcessPath = replace(@'\d', 'n', NormalizedProcessPath)
+  | project TimeGenerated, Computer, Account, Process, ProcessPath, NormalizedProcessPath = replace(@'\d', '#', NormalizedProcessPath)
   ; 
   let freqs = normalizedProcessPath
   | summarize makelist(Computer), makelist(Account), makelist(ProcessPath), frequency=count() by NormalizedProcessPath, Process
--- a/Hunting Queries/SecurityEvent/RareProcessPath.yaml
+++ b/Hunting Queries/SecurityEvent/RareProcessPath.yaml
@@ -1,0 +1,66 @@
+id: 0ff22697-dc58-4623-b844-a767629840cd
+name: Rare Process Path
+description: |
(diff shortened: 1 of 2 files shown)


Hunting Queries/SecurityEvent/RareProcessPath.yaml
```diff
@@ -0,0 +1,66 @@
...
+  | extend path_parts = parse_path(NewProcessName)
+  | extend ProcessPath = tostring(path_parts.DirectoryPath)
+  ;
+  let normalizedProcessPath = processEvents
+  | extend NormalizedProcessPath = ProcessPath
+  // normalize guids
+  | project TimeGenerated, Computer, Account, Process, ProcessPath, 
+  NormalizedProcessPath = replace("[0-9A-Fa-f]{8}[-][0-9A-Fa-f]{4}[-][0-9A-Fa-f]{4}[-][0-9A-Fa-f]{4}[-][0-9A-Fa-f]{12}", "<guid>", NormalizedProcessPath)
+  // normalize digits away
+  | project TimeGenerated, Computer, Account, Process, ProcessPath, NormalizedProcessPath = replace(@'\d', 'n', NormalizedProcessPath)
```
ashwin-patil: Probably use * wildcard instead of n. output is easily identifiable and does not collide with existing n letters in the process names

Hunting Queries/SecurityEvent/RareProcessPath.yaml
```diff
@@ -0,0 +1,66 @@
...
+  let processEvents=
+  SecurityEvent
+  | where TimeGenerated >= start and TimeGenerated <= end
+  | where EventID==4688
+  // excluding well known processes
+  | where NewProcessName !endswith ':\\Windows\\System32\\conhost.exe' and ParentProcessName !endswith ':\\Windows\\System32\\conhost.exe'
+  | where ParentProcessName !endswith ":\\Windows\\System32\\wuauclt.exe" and NewProcessName !startswith "C:\\Windows\\SoftwareDistribution\\Download\\Install\\AM_Delta_Patch_"
+  | where NewProcessName !has ":\\Windows\\WinSxS\\amd64_microsoft-windows-servicingstack_" and ParentProcessName !has ":\\Windows\\WinSxS\\amd64_microsoft-windows-servicingstack_"
+  | where NewProcessName !endswith ":\\WindowsAzure\\SecAgent\\WaSecAgentProv.exe" 
+  //| where ParentProcessName !has ":\\WindowsAzure\\GuestAgent_" and NewProcessName !has ":\\WindowsAzure\\GuestAgent_"
```
petebryan: Is this one meant to be commented out? If so maybe add a comment line to explain when you might want to uncomment it.

Hunting Queries/SecurityEvent/RareProcessPath.yaml
```diff
@@ -0,0 +1,66 @@
...
+  | where ParentProcessName !has ":\\WindowsAzure\\WindowsAzureNetAgent_" and NewProcessName !has ":\\WindowsAzure\\WindowsAzureNetAgent_"
+  | where ParentProcessName !has ":\\ProgramData\\Microsoft\\Windows Defender\\platform\\" and ParentProcessName !endswith "\\MpCmdRun.exe" 
+  | where NewProcessName !has ":\\ProgramData\\Microsoft\\Windows Defender\\platform\\" and NewProcessName !endswith "\\MpCmdRun.exe" 
+  | where NewProcessName !has ':\\Program Files\\Microsoft Monitoring Agent\\Agent\\'
+  // filter out common randomly named paths and files
+  | where not(NewProcessName matches regex @"\\TRA[0-9A-Fa-f]{3}\.tmp")
+  | where not(NewProcessName matches regex @"\\TRA[0-9A-Fa-f]{4}\.tmp")
```
petebryan: Could compress this to:
| where not(NewProcessName matches regex @"\\TRA[0-9A-Fa-f]{3,4}\.tmp"

Hunting Queries/SecurityEvent/RareProcessPath.yaml
```diff
@@ -0,0 +1,66 @@
...
+  | where not(NewProcessName matches regex @"Installer\\MSI[0-9A-Fa-f]{3}\.tmp")
+  | where not(NewProcessName matches regex @"Installer\\MSI[0-9A-Fa-f]{4}\.tmp")
```
petebryan: Could compress to: 
 | where not(NewProcessName matches regex @"Installer\\MSI[0-9A-Fa-f]{3,4}\.tmp")

Hunting Queries/SecurityEvent/RareProcessPath.yaml
```diff
@@ -0,0 +1,66 @@
...
+  - Execution
+query: |
+
+  let end = startofday(now());
+  let start = end - 7d;
+  let processEvents=
+  SecurityEvent
+  | where TimeGenerated >= start and TimeGenerated <= end
+  | where EventID==4688
+  // excluding well known processes
```
petebryan: Have you also considered excluding processes that only run once in the time window? They will always have a rare path by this measurement but they might be running from a completely expected path.

Hunting Queries/SecurityEvent/RareProcessPath.yaml
shainw: Yeah, this came from Tim's other query.  Agree that maybe we use a symbol, such as # or ?.  I don't want to use *.  I will go with # as that is more indicative of a number.

Hunting Queries/SecurityEvent/RareProcessPath.yaml
shainw: Yeah, but we don't want to exclude those either but I meant to make the query 8 days so we cover items that are run weekly as part of a scheduled task or deployment that only runs once.  I will change this to 8 days.  If it is a task that runs less often, the user will have to tune this.

Hunting Queries/SecurityEvent/RareProcessPath.yaml
shainw: True, simply pulled this from another query that Tim did awhile back.  I am not sure I will change this, while I can't see the negative impact, I don't want to have to test again across multiple data sets to verify as I don't think there is much perf hit.

Hunting Queries/SecurityEvent/RareProcessPath.yaml
shainw: Yeah, but we don't want to exclude those either but I meant to make the query 8 days so we cover items that are run weekly as part of a scheduled task or deployment that only runs once. I will change this to 8 days. If it is a task that runs less often, the user will have to tune this.

Hunting Queries/SecurityEvent/RareProcessPath.yaml
shainw: Wrong response, the above was meant for the last comment.  No good catch, I will uncomment.

Hunting Queries/SecurityEvent/RareProcessPath.yaml
shainw: True, simply pulled this from another query that Tim did awhile back. I am not sure I will change this, while I can't see the negative impact, I don't want to have to test again across multiple data sets to verify as I don't think there is much perf hit.

PR #427, opened 2020-01-06
THALLIUM IOCs

timbMSFT:
Fixes #

## Proposed Changes

  -
  -
  -


Detections/MultipleDataSources/ThalliumIOCs.yaml
```diff
@@ -0,0 +1,53 @@
...
+queryFrequency: 1d
+queryPeriod: 1d
+triggerOperator: gt
+triggerThreshold: 0
+tactics:
+  - CommandAndControl
+  - CredentialAccess
+query: |
+  let timeframe = 1d;
+  let DomainNames = dynamic(["seoulhobi.biz", reader.cash", pieceview.club", app-wallet.com", bigwnet.com", bitwoll.com", cexrout.com", change-pw.com", checkprofie.com", cloudwebappservice.com", ctquast.com", dataviewering.com", day-post.com", dialy-post.com", documentviewingcom.com", dovvn-mail.com", down-error.com", drivecheckingcom.com", drog-service.com", encodingmail.com", filinvestment.com", foldershareing.com", golangapis.com", hotrnall.com", lh-logins.com", login-use.com", mail-down.com", matmiho.com", mihomat.com", natwpersonal-online.com", nidlogin.com", nid-login.com", nidlogon.com", pw-change.com", rnaii.com", rnailm.com", sec-live.com", secrityprocessing.com", securitedmode.com", securytingmail.com", set-login.com", usrchecking.com", com-serviceround.info", mai1.info", reviewer.mobi", files-download.net", fixcool.net", hanrnaii.net", office356-us.org", smtper.org"]);
```
shainw: It looks like you only have a quote after each domain listed except the first one.

Detections/MultipleDataSources/ThalliumIOCs.yaml
```diff
@@ -0,0 +1,53 @@
...
+  (union isfuzzy=true
+  (CommonSecurityLog 
+  | where TimeGenerated >= ago(timeframe) 
+  | parse Message with * '(' DNSName ')' * 
+  | where isnotempty(FileHash)
+  | where DNSName in~ (DomainNames)
```
shainw: Wondering if DNSName is always just the second level domain or is it sometimes more and we should use has_any versus in~

Detections/MultipleDataSources/ThalliumIOCs.yaml
```diff
@@ -0,0 +1,53 @@
...
+  | extend Account = SourceUserID, Computer = DeviceName, IPAddress = SourceIP
+  ),
+  (DnsEvents 
+  | where TimeGenerated >= ago(timeframe) 
+  | where isnotempty(DNSName)
+  | where DNSName  in~ (DomainNames)
```
shainw: Wondering if DNSName is always just the second level domain or is it sometimes more and we should use has_any versus in~

Detections/MultipleDataSources/ThalliumIOCs.yaml
```diff
@@ -0,0 +1,53 @@
...
+  | extend IPAddress = ClientIP
+  ),
+  (VMConnection 
+  | where TimeGenerated >= ago(timeframe) 
+  | parse RemoteDnsCanonicalNames with * '["' DNSName '"]' *
+  | where isnotempty(DNSName)
+  | where DNSName  in~ (DomainNames)
```
shainw: Wondering if DNSName is always just the second level domain or is it sometimes more and we should use has_any versus in~

Detections/MultipleDataSources/ThalliumIOCs.yaml
timbMSFT: thanks - fixed

Detections/MultipleDataSources/ThalliumIOCs.yaml
timbMSFT: agreed - this is the approach taken for other activity group IOCs. let's review separately and make a bulk edit as appropriate.

Detections/MultipleDataSources/ThalliumIOCs.yaml
timbMSFT: agreed - this is the approach taken for other activity group IOCs. let's review separately and make a bulk edit as appropriate.

PR #435, opened 2020-01-13
added a query for IoTDevice

moranraz:
added a query for IoTDevice

### DIFF
--- a/Exploration Queries/IoT/Process_byIoTDevice.yaml
+++ b/Exploration Queries/IoT/Process_byIoTDevice.yaml
@@ -1,6 +1,6 @@
 ﻿Id: 1f3ecde7-5c69-4d44-ac93-5feac6d1cd2f
-DisplayName: Most Frequent Command executions on The Device
-Description: Most Frequent Command executions on The Device
+DisplayName: Most frequent command executions on the device
+Description: Most frequent command executions on the device
 InputEntityType: IoTDevice
 InputFields:
   - DeviceId
@@ -9,7 +9,7 @@ OutputEntityTypes:
   - Process
 QueryPeriodBefore: 30m
 QueryPeriodAfter: 30m
-DataSource: SecurityIoTRawEvent
+DataSources: SecurityIoTRawEvent
 query: |
 
   let Process_byIoTDevice = (v_IotDevice_DeviceId:string, v_IoTDevice_IoTHub:string){
--- a/Exploration Queries/IoT/Process_byIoTDevice.yaml
+++ b/Exploration Queries/IoT/Process_byIoTDevice.yaml
@@ -9,8 +9,7 @@ OutputEntityTypes:
   - Process
 QueryPeriodBefore: 30m
 QueryPeriodAfter: 30m
-DataConnector:
-  ConnectorId: SecurityIoTRawEvent
+DataSource: SecurityIoTRawEvent
 query: |
 
   let Process_byIoTDevice = (v_IotDevice_DeviceId:string, v_IoTDevice_IoTHub:string){
--- a/Exploration Queries/IoT/Process_byIoTDevice.yaml
+++ b/Exploration Queries/IoT/Process_byIoTDevice.yaml
@@ -14,11 +14,15 @@ query: |
 
   let Process_byIoTDevice = (v_IotDevice_DeviceId:string, v_IoTDevice_IoTHub:string){
   SecurityIoTRawEvent 
-  | where RawEventName == 'ProcessCreate'
-  | where AssociatedResourceId == parse_json(v_IoTDevice_IoTHub)['ResourceId'] and DeviceId == v_IotDevice_DeviceId
+  | where RawEventName =~ 'ProcessCreate'
+  | where AssociatedResourceId =~ parse_json(v_IoTDevice_IoTHub)['ResourceId'] and DeviceId =~ v_IotDevice_DeviceId
   | extend Process_CommandLine = tostring(parse_json(EventDetails)['CommandLine'])
-  | summarize procCount = count() by Process_CommandLine
-  | top 10 by procCount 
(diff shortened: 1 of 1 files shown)


Exploration Queries/IoT/Process_byIoTDevice.yaml
```diff
@@ -0,0 +1,24 @@
...
+query: |
+
+  let Process_byIoTDevice = (v_IotDevice_DeviceId:string, v_IoTDevice_IoTHub:string){
+  SecurityIoTRawEvent 
+  | where RawEventName == 'ProcessCreate'
+  | where AssociatedResourceId == parse_json(v_IoTDevice_IoTHub)['ResourceId'] and DeviceId == v_IotDevice_DeviceId
+  | extend Process_CommandLine = tostring(parse_json(EventDetails)['CommandLine'])
+  | summarize procCount = count() by Process_CommandLine
+  | top 10 by procCount 
+  | project-away procCount
```
shainw: I don't think we should be projecting away the process count, this is useful information for sorting in the investigation graph.

Exploration Queries/IoT/Process_byIoTDevice.yaml
```diff
@@ -0,0 +1,24 @@
...
+QueryPeriodAfter: 30m
+DataSources: SecurityIoTRawEvent
+query: |
+
+  let Process_byIoTDevice = (v_IotDevice_DeviceId:string, v_IoTDevice_IoTHub:string){
+  SecurityIoTRawEvent 
+  | where RawEventName == 'ProcessCreate'
+  | where AssociatedResourceId == parse_json(v_IoTDevice_IoTHub)['ResourceId'] and DeviceId == v_IotDevice_DeviceId
+  | extend Process_CommandLine = tostring(parse_json(EventDetails)['CommandLine'])
+  | summarize procCount = count() by Process_CommandLine
```
shainw: We are only producing the Command line, I think it is important to include other information in a packed field.  Similar to what is done here - https://github.com/Azure/Azure-Sentinel/blob/master/Exploration%20Queries/SecurityEvent/WinHosts_WithThisProcess.yaml

Exploration Queries/IoT/Process_byIoTDevice.yaml
```diff
@@ -0,0 +1,24 @@
+﻿Id: 1f3ecde7-5c69-4d44-ac93-5feac6d1cd2f
+DisplayName: Most frequent command executions on the device
+Description: Most frequent command executions on the device
+InputEntityType: IoTDevice
+InputFields:
+  - DeviceId
+  - IoTHub
+OutputEntityTypes:
+  - Process
+QueryPeriodBefore: 30m
```
shainw: Just verifying 30 minutes is the right amount of time for the amount of data available?

Exploration Queries/IoT/Process_byIoTDevice.yaml
```diff
@@ -0,0 +1,24 @@
...
+QueryPeriodAfter: 30m
+DataSources: SecurityIoTRawEvent
+query: |
+
+  let Process_byIoTDevice = (v_IotDevice_DeviceId:string, v_IoTDevice_IoTHub:string){
+  SecurityIoTRawEvent 
+  | where RawEventName == 'ProcessCreate'
+  | where AssociatedResourceId == parse_json(v_IoTDevice_IoTHub)['ResourceId'] and DeviceId == v_IotDevice_DeviceId
```
shainw: I don't think this parse will work.  the AssociatedResourceId value in the data is a simply string.  for example - /subscriptions/<guid>/resourcegroups/test-rg/providers/microsoft.devices/iothubs/testhub-iothub.  A simple endswith should match the iothub name of "testhub-iothub".  | where AssociatedResourceId endswith v_IoTDevice_IoTHub or something similar

Exploration Queries/IoT/Process_byIoTDevice.yaml
moranraz: I randomly chose a time frame, as I understood these fields are practically meaningless right now, and aren't used for anything (to my understanding, same thing has been made in other queries as well).

Exploration Queries/IoT/Process_byIoTDevice.yaml
kfbehar: Hi Shain, how should I project this field so that Sentinel can make use of it?

Exploration Queries/IoT/Process_byIoTDevice.yaml
kfbehar: I'll see what I can add, but we don't have much more info from IoT

Exploration Queries/IoT/Process_byIoTDevice.yaml
kfbehar: As discussed by email, you are correct, will fix

Exploration Queries/IoT/Process_byIoTDevice.yaml
shainw: My fault, had regular query review on my mind, no need for this in the exploration queries.

Exploration Queries/IoT/Process_byIoTDevice.yaml
shainw: Okay, if there is no other context that would be useful to an investigation, then I am good here.  I am not an expert on IoT logs, so if this is all all the context we can provide we go with it. :)

Exploration Queries/IoT/Process_byIoTDevice.yaml
shainw: Yeah, for exploration they for right now, but I wanted to confirm 30m is the right time frame you recommend if that functionality is used with exploration queries.

Exploration Queries/IoT/Process_byIoTDevice.yaml
shainw: Thanks much!

PR #441, opened 2020-01-14
updated displayNames, changed dataSources format, added a query templ…

moranraz:
-Updated displayNames
-Changed dataSources format
-Added a query template


moranraz:
> I am good with this, just wondering why we opted for DataSources instead of going with what we have in the Detections which is DataTypes?

I understood this is what our PMs agreed on.

Issue #496, opened 2020-02-20
Playbook – Snapshot an Azure VM


preetikr:
**Requirements**

- Create an Azure Logic Apps playbook for Azure Sentinel that includes the following workflow:
    * Triggers on an Azure Sentinel alert. 
    * Get the incident associated with the alert. 
    * Create a new snapshot of the Azure VM
    * Update the incident with the snapshot location.
- Validate the playbook works as expected per the functional requirement mentioned above. 
- Submit the playbook as a GitHub Pull Request per the playbook contribution guidance.

The first submission that meets the requirements gets the reward. Rewards are subject to [Microsoft terms and conditions](https://aka.ms/sentinelrewardsterms). 
 
**Hint**: HTTP connector for Azure Logic Apps might be helpful here
**Helpful resources**
- [Azure Sentinel Playbook documentation](https://docs.microsoft.com/azure/sentinel/tutorial-respond-threats-playbook)
- [Playbook contribution guidance](https://aka.ms/sentinelgithubplaybooks)
- [Create Azure Logic Apps Playbook](https://docs.microsoft.com/azure/logic-apps/quickstart-create-first-logic-app-workflow)
- [Create / Update Azure VM snapshots](https://docs.microsoft.com/en-us/rest/api/compute/snapshots/createorupdate)

preetikr:
@ThijsLecomte - Congratulations! We've reviewed and merged your submission.
Thanks, for your contribution.
We'll reach out to you to figure the next steps on Rewards for this contribution.

Thanks,
Preeti

PR #501, opened 2020-02-23
Create-AzureSnapshot

ThijsLecomte:
Added Playbook to create a snaphot of a VM.

### DIFF
--- a/Playbooks/Create-AzureSnapshot/Create-Snapshot.json
+++ b/Playbooks/Create-AzureSnapshot/Create-Snapshot.json
@@ -1,0 +1,440 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "PlaybookName": {
+            "defaultValue": "PlaybookName",
+            "type": "string"
+        },
+        "UserName": {
+            "defaultValue": "<username>@<domain>",
+            "type": "string"
+        }
+    },
+    "variables": {
+		"AzureSentinelConnectionName": "[concat('azuresentinel-', parameters('PlaybookName'))]"
+	},
+    "resources": [
+		{
+            "type": "Microsoft.Web/connections",
+            "apiVersion": "2016-06-01",
+            "name": "[variables('AzureSentinelConnectionName')]",
+            "location": "[resourceGroup().location]",
+            "properties": {
+                "displayName": "[parameters('UserName')]",
+                "customParameterValues": {},
+                "api": {
+                    "id": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Web/locations/', resourceGroup().location, '/managedApis/azuresentinel')]"
+                }
+            }
+        },
+        {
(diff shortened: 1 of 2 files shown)


ThijsLecomte:
#496

dicolanl:
@ThijsLecomte Thank you for your submission.  We have reviewed the playbook and believe it is missing some required steps and data in the URLs for the Azure Management REST API.  
For example to get a VM: https://docs.microsoft.com/en-us/rest/api/compute/virtualmachines/get  you need additional properties like subscription and resource group in the URL.

ThijsLecomte:
@dicolanl 

Please find the screenshot for the successful run here:
image

I currently get the subscription ID and resource group from the Account Entity:
"Entities": [
            {
                "$id": "3",
                "DnsDomain": "Compute/virtualMachines/Exchange",
                "HostName": "/subscriptions/23ab6747-a21f-4e9d-ad68-0b0237ad7cd8/resourceGroups/SecOps/providers/Microsoft",
                "Type": "host"
            }
        ]

In my KQL query, I defined ResourceId as the HostCustomEntity and that leverages me all the information I need.

I am doing it this way because I need Subscription ID, Resource Group and VM name to make a snapshot. I cannot add these details in an entity as there are a limited number of entities.. So it makes it difficult to pass these on cleanly. 
The only way to do this would be to query the logs through the 'Run query and list results' action which would get us the rest of the data. Then I could filter out the Subscription ID, Resource Group, VM but that would require a lot of extra steps, which would add an extra Service Principal Required and an extra parameter for the Workspace ID.


I have updated the readme file to reflect the fact that ResourceID is expected as HostCustomEntity.
If you have something better in mind, I am more than happy to look into it with you.

dicolanl:
Hi @ThijsLecomte 
AWESOME seeing the success.

I am concerned there is still a problem.  your entity properties seem off for some reason.

Here is an on-prem node that came from an MDATP alert
image

Here is an azure node that came from ASC alert.
image
you can see dns and hostname shouldnt be azure resource id. there is an AzureID property.

I am not sure how you got that.  I think if other customers run this using your strings for the HTTP its likely going to fail.

ThijsLecomte:
Hi @dicolanl 

It's true that this will only work for certain alerts. More specifically. only alerts where ResourceID is present. It's also not very user friendly.

Let me walk you through my thought process:
As a rule type, we need to use a scheduled query rule as the Microsoft incident creation rule doesn't support playbooks.

The query for the rule type for ASC alerts should be something like:
SecurityAlert 
| where ProductName == "Azure Security Center"

If you run that query manually you will see the Entities populated as you showed in your previous answer. But if you add a Playbook to that alert, the Get Hosts step will return blank as Sentinel overwrites the 'Azure Security Center Entities'. I haven't found a way to copy the Security Center entities to Sentinel entities as entities only support String and Integers (so there is no way to assign an existing object).
I could modify the query used to map the AzureID you mentioned to the HostCustomEntity, but then the same thing will happen as I did with the resourceID, as Sentinel splits the data when it sees a ".".

The only way I see it possible is to query Log Analytics directly through the 'Run query and list results' connector and then parse that output to get the resourceID.
That's how I did it with my latest pull request.

I think this is the best solution for the issue.

image

PR #506, opened 2020-02-25
updated params fields

swiftsolves-msft:
Updated the playbook's parameters to use Azure Sentinel's log analytics workspace name and rg.

Fixes #

## Proposed Changes

  -
  -
  -


dicolanl:
I think you have the default playbook name wrong in the arm template
 "PlaybookName": {
            "defaultValue": "Close-Incident-ASCAlert",
            "type": "String"
        },

PR #526, opened 2020-03-14
Save-NamedLocations

ThijsLecomte:
New playbook to save named locations from AAD to a custom log analytics table.
This way, these can be used in hunting queries.

Successful run: 
image


dicolanl:
@ThijsLecomte Really cool!!!  Please make 2 updates:
1.  Add LogA workspaceID and Key as parameters to use in the API Connection resource.
2.  Either add to the readme or preferably, copy your function code to this repo so users know that the ARM is download additional code.  They may want to review before deploying.

ThijsLecomte:
@dicolanl both updates are done :)
Also blogged about it with more info: https://365bythijs.be/2020/03/23/saving-corporate-ips-to-log-analytics-with-logic-apps/

PR #532, opened 2020-03-18
New Playbook - Isolate-AzureVMtoNSG

swiftsolves-msft:
This playbook will take host entites from triggered incident and search for matches in the enterprises subscriptions. An email for approval will be sent to isolate Azure VM. Upon approval a new NSG Deny All is created and applied to the Azure VM, The Azure VM is restarted to remove any persisted connections.

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Playbooks/Isolate-AzureVMtoNSG/azuredeploy.json
+++ b/Playbooks/Isolate-AzureVMtoNSG/azuredeploy.json
@@ -1,0 +1,659 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "metadata":{
+        "comments": "This playbook will take host entites from triggered incident and search for matches in the enterprises subscriptions. An email for approval will be sent to isolate Azure VM. Upon approval a new NSG Deny All is created and applied to the Azure VM, The Azure VM is restarted to remove any persisted connections.",
+        "author": "Nathan Swift"
+    },
+    "parameters": {
+        "PlaybookName": {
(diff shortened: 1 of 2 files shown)


ThijsLecomte:
Hi @swiftsolves-msft 

Just wanted to check out your Playbook as the functionality looks really great, but I got an error because the LogicApp tried to deploy before the API connection was fully configured.

It is possible you forgot a dependsOn?

swiftsolves-msft:
Thank you for testing this, I did indeed, updated with dependsOn

dicolanl:
@swiftsolves-msft Great Stuff.  One Note... Host Entities (depending on alert source) should have Azure Resource ID as a property.  This could make this workflow easier.

swiftsolves-msft:
> @swiftsolves-msft Great Stuff. One Note... Host Entities (depending on alert source) should have Azure Resource ID as a property. This could make this workflow easier.

@dicolanl Thanks, I should make a conditional check at beginning for the ASC enriched ResourceID field, if none available then move into the Host based lookup method. Another concept I was thinking to use was if no Host entity then check and take the entity IP search for it in Azure whether PubIP or PrivIP and associate it with the AzureResource ID - could be useful from CommonSecurityLogs or other NVA based alerts they may not have a Host entity produced or correlated ?

neogeek3110:
@swiftsolves-msft Where does that email go? To the end user? I don’t see that action anywhere in the playbook. Why would it depends on the end user to take action, If we don’t want the end use to have a say when the device is isolated, then? Because When we are Isolating from MDE portal – there is no user notification is generated.

PR #535, opened 2020-03-19
New Playbook - Run-AzureVMPacketCapture

swiftsolves-msft:
This playbook will take start a packet capture on a Azure VM Windows or Linux using Network Watcher, the capture will run for ten minutes, and will be stored on a blob storage account.

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Playbooks/Run-AzureVMPacketCapture/azuredeploy.json
+++ b/Playbooks/Run-AzureVMPacketCapture/azuredeploy.json
@@ -1,0 +1,354 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "metadata":{
+        "comments": "This playbook will take start a packet capture on a Azure VM Windows or Linux using Network Watcher, the capture will run for ten minutes, and will be stored on a blob storage account.",
+        "author": "Nathan Swift"
+    },
+    "parameters": {
+        "PlaybookName": {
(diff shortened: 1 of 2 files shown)


dicolanl:
@swiftsolves-msft Nice playbook. Will approve, but a couple of improvement recommendations.
1.  you assume the VM is in the same subscription as Sentinel.  Might want to use Azure Resource ID (entity properties) as a cleaner way incase the VM is not in the same sub.
2.  Have the playbook create a storage account and use its properties by passing the values in OR change the parameters to ask for the fields needed and concat them in a variable.  Trying to scroll right on the long string might make a user miss it.

swiftsolves-msft:
> @swiftsolves-msft Nice playbook. Will approve, but a couple of improvement recommendations.
> 
> 1. you assume the VM is in the same subscription as Sentinel.  Might want to use Azure Resource ID (entity properties) as a cleaner way incase the VM is not in the same sub.
> 2. Have the playbook create a storage account and use its properties by passing the values in OR change the parameters to ask for the fields needed and concat them in a variable.  Trying to scroll right on the long string might make a user miss it.

Thanks @dicolanl , I struggle with this because not all alerts come from ASC which has that amazing AzureResourceID field. but I think I can adopt from the isolate playbook Host entity matching logic though my concern remains on a false positive like a host entity name being to generic in larger enterprises and matching across multiple subs, as I recall I introduced the email asking SOC before isolating, maybe something similar here in the beginning approach.

PR #539, opened 2020-03-20
Improving failed logon detection with thresholds, making sure we get …

shainw:
…UPN if we don't have it in the event and fixing issue with IPAddress not being a string.

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/SigninLogs/FailedLogonToAzurePortal.yaml
+++ b/Detections/SigninLogs/FailedLogonToAzurePortal.yaml
@@ -1,8 +1,9 @@
 ﻿id: 223db5c1-1bf8-47d8-8806-bed401b356a4
 name: Failed login attempts to Azure Portal
 description: |
-  'Identifies failed login attempts in the Azure Active Directory SigninLogs to the Azure Portal.  Multiple failed logon 
-  attempts could indicate a potential brute force attack.  We exclude the following success and non-failure results:
+  'Identifies failed login attempts in the Azure Active Directory SigninLogs to the Azure Portal.  Many failed logon 
+  attempts or some failed logon attempts from multiple IPs could indicate a potential brute force attack.  
+  The following are excluded due to success and non-failure results:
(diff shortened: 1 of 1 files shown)


YaronFruchtmann:
#59: I would suggest not to name time columns "TimeXxx**UTC**", since the time zone can be changed by the user, and the name will not change. This would be confusing
image

Detections/SigninLogs/FailedLogonToAzurePortal.yaml
```diff
@@ -23,21 +24,38 @@ relevantTechniques:
...
-  | mvexpand IPAddresses, Status, StatusDetails
-  | extend Status = strcat(Status, "-", StatusDetails)
-  | summarize Status = makeset(Status) by StartTimeUtc, EndTimeUtc, tostring(IPAddresses), UserPrincipalName, AppDisplayName, tostring(Browser), tostring(OS), Location, IPAddressCount, CorrelationId
-  | where Status !has "MFA requirement satisfied by claim in the token"
-  | summarize Status = makeset(Status), IPAddresses = makeset(IPAddresses), CorrelationIds = makeset(CorrelationId) by StartTimeUtc, EndTimeUtc, UserPrincipalName, AppDisplayName, tostring(Browser), tostring(OS), Location, IPAddressCount
-  | extend timestamp = StartTimeUtc, AccountCustomEntity = UserPrincipalName, IPCustomEntity = tostring(IPAddresses)
+  // Azure Portal only and exclude non-failure Result Types
+  | where AppDisplayName has "Azure Portal" and ResultType !in ("0", "50125", "50140")
+  // Tagging identities not resolved to friendly names
+  | extend Unresolved = iff(Identity matches regex "[0-9a-z]{8}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{12}", true, false)
```
ashwin-patil: May want to promote regex pattern to the top as variable as it is used at 2 places so can be changed at one central place if needed in the future

shainw: done

PR #549, opened 2020-03-25
Fixing Services Created on Host and also fixed issue with missing par…

shainw:
…ameters for function in Possible BF

Fixes #

## Proposed Changes

  -
  -
  -


YaronFruchtmann:
Notice that some of the returned columns don't match format and will not be available to user, i.e. ServiceAccount, ServiceName, _ResourceId.... You could change it to Process_Aux_ServiceAccount etc..

PR #551, opened 2020-03-26
Removed ResourceId as it is not  a Host Entity Field.

YaronFruchtmann:
Fixes #

## Proposed Changes

  -
  -
  -


moran94:
the YAML validation is failing, please check it.

PR #593, opened 2020-04-14
query to detect account lockouts

mcpjanmarek:
this is a hunting query for account lockouts in active directory environments.

## Proposed Changes
- new hunting query

### DIFF
--- a/Hunting Queries/SecurityEvent/ADAccountLockouts.yaml
+++ b/Hunting Queries/SecurityEvent/ADAccountLockouts.yaml
@@ -7,8 +7,7 @@ requiredDataConnectors:
     dataTypes:
       - SecurityEvent
 tactics:
-  - InitialAccess
-  - CredentialAccess
+  - Impact
 relevantTechniques:
   - T1531
 query: |
--- a/Hunting Queries/SecurityEvent/ADAccountLockouts.yaml
+++ b/Hunting Queries/SecurityEvent/ADAccountLockouts.yaml
@@ -1,0 +1,20 @@
+id: e7642e6e-cf27-46ec-a4b9-e4475228fead
+name: AD Account Lockout
+description: |
+  'Detects Active Directory account lockouts'
+requiredDataConnectors:
+  - connectorId: SecurityEvents
+    dataTypes:
+      - SecurityEvent
+tactics:
+  - InitialAccess
+  - CredentialAccess
+relevantTechniques:
+  - T1531
+query: |
+  let timeframe = 7d;
+  SecurityEvent
+  | where TimeGenerated >= ago(timeframe)
+  | where EventID == 4740
+  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), count() by Account
+  | extend Timestamp = StartTimeUtc
--- a/Hunting Queries/SecurityEvent/ADAccountLockouts.yaml
+++ b/Hunting Queries/SecurityEvent/ADAccountLockouts.yaml
@@ -12,9 +12,10 @@ tactics:
 relevantTechniques:
   - T1531
 query: |
-  let timeframe = 7d;
-  SecurityEvent
-  | where TimeGenerated >= ago(timeframe)
-  | where EventID == 4740
-  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), count() by Account
-  | extend Timestamp = StartTimeUtc
+let timeframe = 7d;
+SecurityEvent
+| where TimeGenerated >= ago(timeframe)
+| where EventID == 4740
+| extend AccountCustomEntity = tostring(split(Account, '\\', 1)), SourceDomainController = Computer, HostCustomEntity = TargetDomainName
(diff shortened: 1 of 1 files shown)


Hunting Queries/SecurityEvent/ADAccountLockouts.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - InitialAccess
+  - CredentialAccess
+relevantTechniques:
+  - T1531
+query: |
+  let timeframe = 7d;
+  SecurityEvent
+  | where TimeGenerated >= ago(timeframe)
+  | where EventID == 4740
+  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), count() by Account
```
shainw, line 19: We are transitioning to just use StartTime/EndTime versus StartTimeUtc/EndTimeUtc labeling.  UTC is not always how the UX is represented depending on what the User has chosen for their settings in the UX.  Also, we prefer to label items like count so that we don't get the default "count_" label, but not a major issue.  Lastly, it is best to build out context in the results such as what Domain controller this occurred on, what host the logon attempt came from and so on.  Would you be willing to include that?  Documentation on the event is here - https://docs.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4740

Hunting Queries/SecurityEvent/ADAccountLockouts.yaml
```diff
@@ -0,0 +1,20 @@
...
+  | extend Timestamp = StartTimeUtc
```
shainw, line 20: The entity extension here is used for entity mapping in the Azure Sentinel UX.  The value should be timestamp = StartTime, notice the lower case "t".  Also, if you could add in AccountCustomEntity = Account and HostCustomEntity = Computer that helps for mapping also.

Hunting Queries/SecurityEvent/ADAccountLockouts.yaml
mcpjanmarek: hi shain, thank you for additional comments ;) i will definitely update the query as requested. just give me some time ;)

Hunting Queries/SecurityEvent/ADAccountLockouts.yaml
mcpjanmarek: i will update the query as requested. again, please, just give me some time ;)

Hunting Queries/SecurityEvent/ADAccountLockouts.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - connectorId: SecurityEvents
+    dataTypes:
+      - SecurityEvent
+tactics:
+  - InitialAccess
+  - CredentialAccess
+relevantTechniques:
+  - T1531
+query: |
+let timeframe = 7d;
```
shainw, line 15: 2 spaces before each line in the query, otherwise the YAML template format is wrong and the build checks break.

Hunting Queries/SecurityEvent/ADAccountLockouts.yaml
```diff
@@ -0,0 +1,21 @@
...
+SecurityEvent
+| where TimeGenerated >= ago(timeframe)
+| where EventID == 4740
+| extend AccountCustomEntity = tostring(split(Account, '\\', 1)), SourceDomainController = Computer, HostCustomEntity = TargetDomainName
+| summarize StartTime = min(TimeGenerated), EndTime = max(TimeGenerated), LockoutsCount = count() by AccountCustomEntity, HostCustomEntity, SourceDomainController
+| extend timestamp = StartTime
```
shainw, line 21: In general, good stuff.  We usually bring thru the common name and add in the extended name so we don't lose the specific context of the original event.  Here is what we can do and I put some comments in to help understand why we bring a value thru, you can choose to leave those in or take them out.

let timeframe = 4h;
SecurityEvent
| where EventID == 4740
// Bring thru 
// Activity <-- this is the event id with the friendly name and is helpful for context
// TargetSid <-- strong identifer for the user account
// TargetDomainName <-- Host where the request came from
// SourceComputerId <-- this is a strong identifer to use with other tables such as Heartbeat to get something like the IP at the time
// Computer as SourceDomainController <-- this is the DC where the Event was logged that is associated to the account that locked out
| summarize StartTime = min(TimeGenerated), EndTime = max(TimeGenerated), LockoutsCount = count() by Activity, Account, TargetSid, TargetDomainName, SourceComputerId, SourceDomainController = Computer
| extend timestamp = StartTime, AccountCustomEntity = Account, HostCustomEntity = TargetDomainName

shainw:
@mcpjanmarek - I posted some additional changes and some required fixes for build break issues.  Please have a look.  Thanks, Shain

shainw:
@mcpjanmarek - Reminder to look at suggested changes.  Thanks! Shain

mcpjanmarek:
> @mcpjanmarek - Reminder to look at suggested changes. Thanks! Shain

Hi Shain, I'm sorry for a delay, I work on other priorities these days. Could you please give me a couple days? Thank you, Jan

shainw:
@mcpjanmarek - if you want, I can fix this up and complete it.  Understand that you may not have the time now. :)

PR #607, opened 2020-04-21
Create SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts

jross1012:
In addition to the existing Sign ins from IPs that attempt sign ins to disabled accounts analytic; This analytic will additionally identify the successful signed in accounts as the mapped account entities for investigation in Sentinel. This analytic will additionally identify the successful signed in accounts as the mapped account entities for investigation in Sentinel.
[SignInIPsDisabled.zip](https://github.com/Azure/Azure-Sentinel/files/4507065/SignInIPsDisabled.zip)

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/SigninLogs/SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts
+++ b/Detections/SigninLogs/SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts
@@ -1,0 +1,48 @@
+name: Sign-ins from IPs that attempt sign-ins to disabled accounts
+description: |
+  'Identifies IPs with failed attempts to sign in to one or more disabled accounts signed in successfully to another account.
+  References: https://docs.microsoft.com/en-us/azure/active-directory/reports-monitoring/reference-sign-ins-error-codes
+  50057 - User account is disabled. The account has been disabled by an administrator.' This analytic will additionally identify the successful signed in accounts as the mapped account entities for investigation in Sentinel.
+severity: Medium
+requiredDataConnectors:
+  - connectorId: AzureActiveDirectory
+    dataTypes:
+      - SigninLogs
+queryFrequency: 1d
+queryPeriod: 1d
+triggerOperator: gt
+triggerThreshold: 0
+tactics:
+  - InitialAccess
+  - Persistence
+relevantTechniques:
+  - T1078
+  - T1098
+query: |
+ let lookBack = 1d;
+  SigninLogs 
+  | where TimeGenerated >= ago(lookBack)
+  | where ResultType == "50057" 
+  | where ResultDescription == "User account is disabled. The account has been disabled by an administrator." 
+  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), disabledAccountLoginAttempts = count(), 
+  disabledAccountsTargeted = dcount(UserPrincipalName), applicationsTargeted = dcount(AppDisplayName), disabledAccountSet = makeset(UserPrincipalName), 
+  applicationSet = makeset(AppDisplayName) by IPAddress
+  | order by disabledAccountLoginAttempts desc
+  | join kind= leftouter (
+      // Consider these IPs suspicious - and alert any related  successful sign-ins
+      SigninLogs
+      | where TimeGenerated >= ago(lookBack)
+      | where ResultType == 0
+      | summarize successfulAccountSigninCount = dcount(UserPrincipalName), successfulAccountSigninSet = makeset(UserPrincipalName, 15) by IPAddress
+      // Assume IPs associated with sign-ins from 100+ distinct user accounts are safe
+      | where successfulAccountSigninCount < 100
+  ) on IPAddress  
+  // IPs from which attempts to authenticate as disabled user accounts originated, and had a non-zero success rate for some other account
+  | where successfulAccountSigninCount != 0
+  | project StartTimeUtc, EndTimeUtc, IPAddress, disabledAccountLoginAttempts, disabledAccountsTargeted, disabledAccountSet, applicationSet, 
+  successfulAccountSigninCount, successfulAccountSigninSet
+  | order by disabledAccountLoginAttempts
+  | extend timestamp = StartTimeUtc, IPCustomEntity = IPAddress
+// Break up the string of Succesfully signed into accounts into individual events
+| mvexpand successfulAccountSigninSet
+| extend AccountCustomEntity = tostring(successfulAccountSigninSet)
--- a/Detections/SigninLogs/SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts
+++ b/Detections/SigninLogs/SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts
@@ -20,11 +20,12 @@ relevantTechniques:
   - T1098
 query: |
  let lookBack = 1d;
+ let threshold = 100;
   SigninLogs 
   | where TimeGenerated >= ago(lookBack)
   | where ResultType == "50057" 
   | where ResultDescription == "User account is disabled. The account has been disabled by an administrator." 
-  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), disabledAccountLoginAttempts = count(), 
+  | summarize StartTime = min(TimeGenerated), EndTime = max(TimeGenerated), disabledAccountLoginAttempts = count(), 
   disabledAccountsTargeted = dcount(UserPrincipalName), applicationsTargeted = dcount(AppDisplayName), disabledAccountSet = makeset(UserPrincipalName), 
   applicationSet = makeset(AppDisplayName) by IPAddress
   | order by disabledAccountLoginAttempts desc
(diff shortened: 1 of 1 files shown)


Detections/SigninLogs/SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts
```diff
@@ -0,0 +1,48 @@
...
+triggerOperator: gt
+triggerThreshold: 0
+tactics:
+  - InitialAccess
+  - Persistence
+relevantTechniques:
+  - T1078
+  - T1098
+query: |
+ let lookBack = 1d;
```
shainw, line 22: Make sure to get 2 spaces before each of the query lines so YAML format passes checks during build.

Detections/SigninLogs/SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts
```diff
@@ -0,0 +1,48 @@
...
+  SigninLogs 
+  | where TimeGenerated >= ago(lookBack)
+  | where ResultType == "50057" 
+  | where ResultDescription == "User account is disabled. The account has been disabled by an administrator." 
+  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), disabledAccountLoginAttempts = count(),
```
shainw, line 27: We are transitioning to StartTime and EndTime, dropping the Utc as this depends on per user settings and we don't want the label to confuse users if they have chosen local time as their setting in the UX.

Detections/SigninLogs/SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts
```diff
@@ -0,0 +1,48 @@
...
+      | summarize successfulAccountSigninCount = dcount(UserPrincipalName), successfulAccountSigninSet = makeset(UserPrincipalName, 15) by IPAddress
+      // Assume IPs associated with sign-ins from 100+ distinct user accounts are safe
+      | where successfulAccountSigninCount < 100
+  ) on IPAddress  
+  // IPs from which attempts to authenticate as disabled user accounts originated, and had a non-zero success rate for some other account
+  | where successfulAccountSigninCount != 0
+  | project StartTimeUtc, EndTimeUtc, IPAddress, disabledAccountLoginAttempts, disabledAccountsTargeted, disabledAccountSet, applicationSet, 
+  successfulAccountSigninCount, successfulAccountSigninSet
+  | order by disabledAccountLoginAttempts
+  | extend timestamp = StartTimeUtc, IPCustomEntity = IPAddress
```
shainw, line 45: You could do this all as the same extend at the end, a bit cleaner.  Also, remember to change to StartTime after you make the change above.

Detections/SigninLogs/SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts
```diff
@@ -0,0 +1,48 @@
...
+      SigninLogs
+      | where TimeGenerated >= ago(lookBack)
+      | where ResultType == 0
+      | summarize successfulAccountSigninCount = dcount(UserPrincipalName), successfulAccountSigninSet = makeset(UserPrincipalName, 15) by IPAddress
+      // Assume IPs associated with sign-ins from 100+ distinct user accounts are safe
+      | where successfulAccountSigninCount < 100
+  ) on IPAddress  
+  // IPs from which attempts to authenticate as disabled user accounts originated, and had a non-zero success rate for some other account
+  | where successfulAccountSigninCount != 0
+  | project StartTimeUtc, EndTimeUtc, IPAddress, disabledAccountLoginAttempts, disabledAccountsTargeted, disabledAccountSet, applicationSet,
```
shainw, line 42: Above the project, it might be good to notate if the success was before or after the failure.  Helps build context.  You could add a column that identifies this, such as | extend SuccessBeforeFailure = iff(SuccessStartTime < StartTime, true, false)

Detections/SigninLogs/SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts
```diff
@@ -0,0 +1,48 @@
...
+  applicationSet = makeset(AppDisplayName) by IPAddress
+  | order by disabledAccountLoginAttempts desc
+  | join kind= leftouter (
+      // Consider these IPs suspicious - and alert any related  successful sign-ins
+      SigninLogs
+      | where TimeGenerated >= ago(lookBack)
+      | where ResultType == 0
+      | summarize successfulAccountSigninCount = dcount(UserPrincipalName), successfulAccountSigninSet = makeset(UserPrincipalName, 15) by IPAddress
+      // Assume IPs associated with sign-ins from 100+ distinct user accounts are safe
+      | where successfulAccountSigninCount < 100
```
shainw, line 38: Change to a variable so it is easily discoverable:
At top add - let threshold = 100;
Then change to - | where successfulAccountSigninCount < threshold

Detections/SigninLogs/SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts
shainw: You would need to add something like this in the project inside the join -
| summarize SuccessStartTime = min(TimeGenerated), ....

shainw:
@jross1012 - Please have a look at the suggested changes here also.  Thanks for the multiple submissions.  Shain

shainw:
@jross1012 - another reminder.  Please review feedback.  Thanks!

Detections/SigninLogs/SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts
jross1012: @shainw While I see that I could add this "| extend successbeforefailure = iff(TimeGenerated < StartTime, true, false) " to show per successful login, I don't see how to format this in the project even using the summarize you recommend. Because the project shows a set of successful sign-ins I don't see how I could summarize this.

jross1012:
@shainw I updated to query to reflect the changes you recommended. I did not include the addition of the 'SuccessBeforeFailure' in the project as recommended as I need some assistance in drafting the summarize function.
My comment - While I see that I could add this "| extend successbeforefailure = iff(TimeGenerated < StartTime, true, false) " to show per successful login, I don't see how to format this in the project even using the summarize you recommend. Because the project as is shows a set of successful sign-ins I don't see how I could summarize this as the multiple signins recorded in the SuccessfulSigninSet could of occurred before or after the "StartTime"

jross1012:
@shainw it looks like the checks errored out at the start of my query, do you know why that could be, the check report didn't seem too informative: 
Lines 22-24
query: |
  let lookBack = 1d;
  let threshold = 100;
  SigninLogs

shainw:
@jross1012 - apologies for delay, I extended the holiday weekend and didn't get to this before then.  Anyhow, the build error is due to documentation links in the comments having a locale in them. If all else checks out, I can merge and fix this in a subsequent PR.  The piece related to SuccessBeforeFailure we can discuss further, but not a blocker on bringing this in as is.

Detections/SigninLogs/SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts
shainw: Will take offline for potential later change, no requirement as of now.

Detections/SigninLogs/SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts
```diff
@@ -0,0 +1,50 @@
...
+      | where TimeGenerated >= ago(lookBack)
+      | where ResultType == 0
+      | summarize successfulAccountSigninCount = dcount(UserPrincipalName), successfulAccountSigninSet = makeset(UserPrincipalName, 15) by IPAddress
+      // Assume IPs associated with sign-ins from 100+ distinct user accounts are safe
+      | where successfulAccountSigninCount < threshold
+  ) on IPAddress  
+  // IPs from which attempts to authenticate as disabled user accounts originated, and had a non-zero success rate for some other account
+  | where successfulAccountSigninCount != 0
+  // Successful Account Signins occur within the same lookback period as the failed 
+  | extend SuccessBeforeFailure = iff(TimeGenerated < StartTime, true, false)
```
shainw, line 44: Remove this line for now, we will take offline to discuss, but not needed for the intent of your detection.

Detections/SigninLogs/SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts
```diff
@@ -0,0 +1,50 @@
...
+  ) on IPAddress  
+  // IPs from which attempts to authenticate as disabled user accounts originated, and had a non-zero success rate for some other account
+  | where successfulAccountSigninCount != 0
+  // Successful Account Signins occur within the same lookback period as the failed 
+  | extend SuccessBeforeFailure = iff(TimeGenerated < StartTime, true, false) 
+  | project StartTime, EndTime, IPAddress, disabledAccountLoginAttempts, disabledAccountsTargeted, disabledAccountSet, applicationSet, 
+  successfulAccountSigninCount, successfulAccountSigninSet
+  | order by disabledAccountLoginAttempts
+  // Break up the string of Succesfully signed into accounts into individual events
+  | mvexpand successfulAccountSigninSet
```
shainw, line 49: I would suggest you change this to disabledAccountSet, otherwise, you effectively get a bunch of duplicate time entries for every enabled account that came from the same IP.  In my testing if I use the successful Account I get about 60 rows returned for 1 day, likely too many alerts and I think the focus should start with the disabled account as part of the alert and then the investigation would be into the successful accounts, the associated IP and the AppName used.  Change this mvexpend to | mvexpand disabledAccountSet

Detections/SigninLogs/SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts
```diff
@@ -0,0 +1,50 @@
...
+  | extend AccountCustomEntity = tostring(successfulAccountSigninSet), timestamp = StartTime, IPCustomEntity = IPAddress
```
shainw, line 50: Change the AccountCustomEntity to | extend AccountCustomEntity = tostring(disabledAccountSet).  This reduced my returns to 4 rows, versus 60 and I think allows for the proper account focus.

jross1012: This will return the disabled accounts that were used to sign in versus the accounts that successfully signed in which we are trying surface with this detection.

jross1012: Sorry @shainw I did not see a notification on your request prior

shainw: @jross1012 - looks like I missed yours in return. :)  The challenge here is that it returns too many rows for a single day and if not grouped there would be potentially 60 individual alerts.  We can leave it as is, but we would want it moved to the Hunting Queries sections versus Detections to start with.  If you want to keep as a Detection, then we should think about ways to reduce the result set so that it is not a consistently high number.

shainw: @jross1012 - Shall we move this to a Hunting query for now?

jross1012: @shainw I agree I think that is the best next step. Do I need to do another pull request?

shainw: @jross1012 - I will will do the location change and there is a minor fix I will do in relation to a new check with have on localized URLs which I will do.

PR #615, opened 2020-04-23
Restrict-MDATPIPAddress

dicolanl:
Fixes #

## Proposed Changes

  -
  -
  -


dicolanl:
Ill update next week, but I would like to point out that for #2 I changed two fields in the JSON body to go from URL to IP.  It was very easy

PR #620, opened 2020-04-25
add Azure-Sentinel-Notebooks as submodule

zhzhao8888:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "NotebooksV2"]
+	path = NotebooksV2
+	url = https://github.com/Azure/Azure-Sentinel-Notebooks.git


zhzhao8888:
> 
> 
> Why isn't this Notebooks instead of NotebooksV2? The intent is to replace the redirection from the current Azure Sentinel GitHub "Notebooks" folder directly to the new Notebooks repo.

Technically we cannot rename the submodule as Notebooks because it is already used unless we delete it first.  If we delete it, then we will lose all the change history. Therefore, I renamed it to NotebooksV2.  If MSTIC guys are fine to lose history, then we must delete the folder first.

zhzhao8888:
> 
> 
> We should make it 'Notebooks' and not 'NotebooksV2'

Technically we cannot rename the submodule as Notebooks because it is already used unless we delete it first.  If we delete it, then we will lose all the change history. Therefore, I renamed it to NotebooksV2.  If MSTIC guys are fine to lose history, then we must delete the folder first.

PR #621, opened 2020-04-25
Add Block-IP-on-MDATP-Using-GraphSecurity playbook

chinguyen1:
This playbook will get the IP information from an Azure Sentinel incident when the security incident triggers in Azure Sentinel. The IP info will then be sent to a security analyst over email. The analyst will either approve or reject the blocking of the IP. If approved, the IP will then be pushed to be blocked on Microsoft Defender ATP using the Graph Security TI indicator Post method.

### DIFF
--- a/Playbooks/Block-IPs-on-MDATP-Using-GraphSecurity/Readme.md
+++ b/Playbooks/Block-IPs-on-MDATP-Using-GraphSecurity/Readme.md
@@ -2,4 +2,4 @@ author: Chi Nguyen
 
 This playbook will get the IP information from an Azure Sentinel incident when the security incident triggers in Azure Sentinel. The IP info will then be sent to a security analyst over email. The analyst will either approve or reject the blocking of the IP. If approved, the IP will then be pushed to be blocked on Microsoft Defender ATP using the Graph Security TI indicator Post method.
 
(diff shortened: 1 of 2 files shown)


dicolanl:
@chinguyen1 
the deployment failed.
image

a couple of things i found.  
1.  you need a dependson as part of the logic app resource to make sure the connections deploy 1st then the workflow.  see other playbook samples for this.
2.  your variable names for the connections have random #s in them.  Please remove them.
"variables": {
		"GraphSecurityConnectionName": "[concat('microsoftgraphsecurity-11-', parameters('PlaybookName'))]",
		"AzureSentinelConnectionName": "[concat('azuresentinel-', parameters('PlaybookName'))]",
		"Office365ConnectionName": "[concat('office365-7-', parameters('PlaybookName'))]"
	},

chinguyen1:
@dicolanl Thanks for pointing that out - all makes sense. I should've tested it in other tenants beside my demo tenant where the logic app resides.
Fixed a few things in azuredeploy.json, tested in other tenants & all deployments worked ok on these tenants on my end. Can you please try deploying it again & let me know if it works?

chinguyen1:
@dicolanl Thanks for pointing that out - all makes sense. I initially tested deployment on my demo tenant; somehow it passed, but I should've also tested on different tenants. 
I fixed the template, tested it on other tenants & all deployments worked fine on my end. Can you please retry it?

Issue #520, opened 2020-03-12
Playbook – Create a list of discovered vulnerabilities pertaining to a machine/host 


preetikr:
**Requirements**

- Create an Azure Logic Apps playbook for Azure Sentinel that includes the following workflow:
   * Triggers on an Azure Sentinel alert. 
   * If the alert has hosts, find list of vulnerabilities pertaining to each host/machine. Include name, description, severity, CVE at a minimum. 
   * Feed that into a document and post the document onto a SharePoint site.
   * Update the Azure Sentinel incident with the link to the vulnerability report.
- Validate the playbook works as expected per the functional requirement mentioned above. 
- Submit the playbook as a GitHub Pull Request per the playbook contribution guidance.
- Attach a screenshot of the working playbook with your submission. 
- Attach a screenshot of the final output with the vulnerability data. 
- Provide a readme with the steps with your submission to document any prerequisite steps required before running this playbook.

The first submission that meets the requirements gets the reward. Rewards are subject to [Microsoft terms and conditions](https://aka.ms/sentinelrewardsterms). 
 
**Hints** 

- Connect with any product (API) of your choice that provides vulnerability list. Few examples are [Qualys](https://www.qualys.com/documentation/), [Microsoft Defender ATP](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/get-discovered-vulnerabilities), etc.
- Feel free to leverage HTTP Connector for Azure Logic Apps as needed.

**Helpful resources**
- [Azure Sentinel Playbook documentation](https://docs.microsoft.com/azure/sentinel/tutorial-respond-threats-playbook)
- [Playbook contribution guidance](https://aka.ms/sentinelgithubplaybooks)
- [Create Azure Logic Apps Playbook](https://docs.microsoft.com/azure/logic-apps/quickstart-create-first-logic-app-workflow)
- [SharePoint Logic Apps connector](https://docs.microsoft.com/en-us/connectors/sharepointonline/)

sysgoblin:
Any preference or stipulation on the document type you want to hold the vulnerability data? (csv, word doc, txt etc?)

shainw:
[...]

PR #625, opened 2020-04-28
Get-MDATPVulnerabilities

WayneJLee:
Hey guys, first submission here for #520. The playbook takes the Host entity from an alert and queries the Defender ATP API for the list of vulnerabilities associated with the machine. Please use the Word template provided in the pull as well. Not great at Word so sorry for the plain set up.

You can find further instructions in the readme. Let me know if you come across any issues because I tried to make it as dynamic as I could to ensure it's straight forward to set up.

Screenshots:

image

image

image


preetikr:
Thanks, @WayneJLee for your first contribution to the Azure Sentinel GitHub repo. We will review this playbook and get back.

dicolanl:
Hi @WayneJLee 
Great playbook!  I have some fixes that must be resolved and some improvements I recommend.

Fixes:
The deployment failed using your template.
image

-logic app resource had diff location than then connections resources.  Connection resource was static.  You can actually removed the parameter LogicApp location.
image
template deployments ask the user for RG and location.  So you can use [resourceGroup().location] in the resource location property on all resources and this will make the template much more dynamic.
 
-your connection resources  have the wrong id.
{
      "type": "MICROSOFT.WEB/CONNECTIONS",
      "apiVersion": "2018-07-01-preview",
      "name": "[variables('AzureSentinelConnectionName')]",
      "location": "japaneast",
      "properties": {
        "api": {
          "id": "[concat(subscription().id, '/providers/Microsoft.Web/locations/', 'japaneast', '/managedApis/', 'azuresentinel')]"
        },
        "displayName": "[variables('AzureSentinelConnectionDisplayName')]"
      }
    }, 

for api.id on these connection resources use the following as it makes it dynamic for the deployment:
"id": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Web/locations/', resourceGroup().location, '/managedApis/azuresentinel')]"
See now its getting the subid, and rg location.

Improvements:
-You could make Tenantid dynamic.  using [subscription().tenantId], then you would not need to prompt the user for the tenant id.

- Consider putting the client secret in a key vault and have a step retrieve the key.
- Could also consider making it use managed identity then you would not need the app id at all.

- Use the Azure Sentinel - get hosts action instead of parse json for entities.  we have this built in connector, will make it easier for other users to see what you are doing.
image

- You dont need to get auth token using HTTP connector.  when using HTTP you can provide the info for OAUTH authentication - this saves 2 steps!  just click add parameter and click authentication.
image

- instead of get list via HTTP.  defender has a connection already for this.  Plus then you can put an odata filter to get just that machine.  Once you have that you can save many steps by removing the loops and conditions and just call the get vuln http passing in the machine id returned from the builtin connector!
image

WayneJLee:
Thank you very much for this @dicolanl . I'm still pretty new at ARM templates and LogicApp so the feedback is definitely appreciated. 

Will work on the fixes and improvements right away. Regarding the tenant ID, I was actually contemplating making the TenantID dynamic. The reason I didn't was because I work with an MSSP so we usually need to provide a separate Tenant ID for the client's MDATP and Sharepoint environments.

If this is something Microsoft would rather not have, I can definitely make it dynamic.

dicolanl:
Totally understand the mssp scenario.

One idea might be to have two templates. One for mssp one for customer!!

Nicholas
Security Jedi
I
________________________________

WayneJLee:
Hey @dicolanl, I reworked the entire playbook (editing was too much of a hassle) and I incorporated your suggestions as much as I could. The loops unfortunately won't go away due to the output of the connectors but it's cleaner and should handle multiple hosts in an alert better.

Added:

- Get secret - Azure KV connector
- Get hosts - Azure Sentinel connector
- Get list of machines - WDATP Connector
- Utilization of OAuth - HTTP Connector
_For my MSSP use case, I added used tenantId as a parameter so if the user requires to change it, they can just change the default value of the parameter._ 

I also tried exploring Managed Identities but I'm not sure how to configure the necessary permissions for WDATP API. Would appreciate it if you could point it out.

I believe I've also fixed the ARM template to ensure everything is dynamic. Please let me know if you're still facing any issues. In terms of params, right now all you need is to provide the Playbook name and set up the connectors post deployment.


Here's a screenshot of a successful run.

image

Issue #634, opened 2020-04-30
O365API Logging Logic App errors


thetootall:
I followed the configuration process to setup and install the logic app to feed data into the custom O365API Log Analytics logging source and read that when there is no data the "Parse JSON" step will fail with":

InvalidTemplate. Unable to process template language expressions in action 'Parse_JSON' inputs at line '1' and column '2868': 'The template language expression 'body('HTTP')[0]' cannot be evaluated because array index '0' cannot be selected from empty array. Please see https://aka.ms/logicexpressions for usage details.'.

It's super annoying because that means a majority of the runs will come back with a failure code.

However the times when there is data and gets to the "Send Data" step I'm finding that these inputs come back with the following as part of my raw OUTPUT:

"body": {
        "error": {
            "code": 400,
            "message": "A value for subdomain must be specified in token:username.",
            "source": "logic-apis-eastus.azure-apim.net",
            "path": "",
            "clientRequestId": xxxxxx
        }

When I look thru my raw INPUT i get the following:

{
    "method": "post",
    "headers": {
        "Log-Type": "O365API",
        "time-generated-field": "CreationTime"
    },
    "path": "/api/logs",
    "host": {
        "connection": {
            "name": "/subscriptions/xxxxxxxxxxxxxxxxxxxxxxx/resourceGroups/xxxxxxxxxx/providers/Microsoft.Web/connections/azureloganalyticsdatacollector-Get-O365Data"
        }
    },
  
Along with a large blob of data, so I know that its connecting to my tenant but its failing with the 400 error message.

Please advise

dicolanl:
Hi @thetootall 
I think we have a bug.  Let me take it with @petebryan

petebryan:
Thanks @thetootall we heard the feedback on the failure situation when there is no Teams data to parse, we will look at updating the logic app so that it has a conditional flow where if no data is returned it just ends the execution rather than failing.

With the "Send Data" step issue the input you are seeing is the correct format. Looking at the [API documentation](https://docs.microsoft.com/en-us/azure/azure-monitor/platform/data-collector-api) the most common cause of the 400 response is an issue with the content or the workspace. 
To help with this can you please check that the Send Data Step is correctly connected to the Workspace you have, also can you please confirm if the blob of data you are seeing being sent appears to be json data (it should be). In addition could you let us know what setting you have set of the Send Data step.

[...]

PR #639, opened 2020-05-01
Get-0365Data

dicolanl:
Fixes #
#634 

## Proposed Changes
- Including a step that looked at the most recent Teams log time and extending the query out to this (or if they are no logs try and get the last day's worth)
-Add a conditional statement so if no data is available the function app gracefully exits rather than failing.
-Handling when multiple contentUris are returned.

### DIFF
--- a/Playbooks/Get-O365Data/azuredeploy.json
+++ b/Playbooks/Get-O365Data/azuredeploy.json
@@ -16,17 +16,45 @@
         "AADAppSecret": {
             "type": "string"
         },
-        "AADTenantID": {
+        "workspaceId": {
             "type": "string"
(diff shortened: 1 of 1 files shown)


mrblacyk:
hey guys!

Just a little input here to maybe prioritize it a little more.

I want to use this in my work because it seems to resolve the now broken [logic app template](https://github.com/Azure/Azure-Sentinel/tree/master/Playbooks/Get-O365Data) but I'd really appreciate the code review and a merge, @Yaniv-Shasha @KobyKoren, to be sure that it's production ready.

Thanks in advance and thank you @dicolanl for the potential fix.

PR #641, opened 2020-05-02
Move files into folders without changing line endings

chinguyen1:
Fixes #

## Proposed Changes

  -
  -
  -


chinguyen1:
@preetikr those files were added separately after the first two commits. I can open a brandnew branch & redo the PR & new commits will only include files move without line ending changes

PR #663, opened 2020-05-07
Update FluentD-VMSS-UB-Templatev2.json

happy-jo:
## Proposed Changes

- Removed the extensionProfiles as they are replaced by the code executed in the cloud-init. (not sure if the agent dependencies need to be written into the cloud-init.)
- Added The Syslog 5514 UDP rule to the NSG.
- Adding TCP to the Syslog 5514 TCP rule name.
- Changed the Allow SSH NSG priority.
- URL errors were found in the Github links where FluentD-VMSS should be Fluend-vmss and the yaml link was missing a /plugins/ folder.
- Change the wget command to that downloads the GeoIP MMDB to curl. For some reason, the mmdb was not being downloaded on install.


happy-jo:
One last recommendation would be to change the VM instance size to something smaller like a Standard_D2s_V3 or Standard_D4s_V3.
The one in the template fails in some location due to instance core count limitations.

Issue #651, opened 2020-05-05
Cloud-init script fails


happy-jo:
I'm trying to run the ARM template and cannot seem to get the get my syslog events to flow into log analytics so I started doing some RCA on the issue and it looks like the Clound-init or the "cef_installer.py" is not functioning correctly

When looking at the VMSS console I see the error that the oms-agent has failed and that the oms config could not be changed (see attached pic 1.) I check for the location of the config but the path did not exist. (e.g. etc/opt/microsoft...) (see pic 2.) So it looks like the install might have failed.

![Cloud-init issue pic 1](https://user-images.githubusercontent.com/16526141/81082191-f09e6980-8ec0-11ea-920c-b1ef1b4fde99.png)

So from here I started with a fresh VMSS deployment and started running some of the command separately so to see where the failure was coming into play.

**The first one was below directly from the cloud-init file:**
`sudo wget https://raw.githubusercontent.com/Azure/Azure-Sentinel/master/DataConnectors/CEF/cef_installer.py && sudo python cef_installer.py ', <LA Workspace ID>, ' ', <LA Workspace Secret>, '`

which fails on the installation because the config path cannot be found.

![Cloud-init issue pic 2](https://user-images.githubusercontent.com/16526141/81082215-f98f3b00-8ec0-11ea-8891-406b07225a6a.png)

**The next script I ran was**
`onboard_agent.sh -w  <LA Workspace ID>  -s  <LA Workspace Secret>`

and this works but there is obviously some incomplete steps.

So looking back at the cef_installer.py, it looks like the where the python script executes the bash script in the cloud-init file, the params are not being passed to the VM console correctly,  where something like the below is executed and leaving <**,**> in the shell execution. (see pic 3.)

```
Installing omsagent
sh onboard_agent.sh -w , <LA Workspace ID>,  -s , <LA Workspace Secret>,  -d opinsights.azure.com
```

![Cloud-init issue pic 3](https://user-images.githubusercontent.com/16526141/81082292-1461af80-8ec1-11ea-89f7-12f928f7ba3a.png)

**So, I might be completely going the wrong way, but is this something that can be fix?**

dicolanl:
@happy-jo  which cloud init is this?  CEF, FluentD, or logstash?

I tested these and used them with some customers without this issue.  Did you change OS/Image?  i only tested with RedHat and Ubuntu.

happy-jo:
A few issues where observed:

[...]

PR #668, opened 2020-05-11
Update CEF-VMSS-RH-Templatev2.json

happy-jo:
## Proposed Fixes

- Added Syslog 514 UDP for to NSG
- Changed Instance size to Standard_D2s_v3 to correct the core count deployment error
- Removed the extension profiles since the OMS agent install is written into the cloud-init string


dicolanl:
#651

PR #678, opened 2020-05-13
(Work in Progress) Documentation links should not include locale - fix and add validations

igalshapira:
Fixes #

## Proposed Changes

  - Documents link should not contain locale in it - user should be redirected to the correct langauge
  - Add a test to verify that
  -


.script/documentsLinkValidator.ts
```diff
@@ -0,0 +1,27 @@
+import { runCheckOverChangedFiles } from "./utils/changedFilesValidator";
+import { ExitCode } from "./utils/exitCode";
+import fs from "fs";
+import * as logger from "./utils/logger";
+
+export async function IsFileContainsLinkWithLocale(filePath: string): Promise<ExitCode> {
+  const content = fs.readFileSync(filePath, "utf8");
+  if (/(https:\/\/docs.microsoft.com|https:\/\/azure.microsoft.com)(\/[a-z]{2}-[a-z]{2})/i.test(content)) {
+      throw new Error();
```
sagamzu, line 9: Maybe add custom error message here?
I think this message will be displayed in `onExecError`  by `{e.message}`

PR #679, opened 2020-05-13
ASC Auto connect to Sentinel

lior-tamir:
I'm posting a blogpost with very extended instructions, as soon as it is posted I'll add a link to the readme.


dicolanl:
@lior-tamir 
sweet playbook.  could you create a folder for the playbook?  we put each playbook into its own folder and add a readme to explain the playbook and any additional steps.

PR #680, opened 2020-05-13
ASC connections workbook

lior-tamir:
This workbook allows to see all connected ASC subscriptions.
It is useful when you are a analyst in the SOC and don't have permissions to see all the connections in the connector page, but you want to monitor the amount of connections, the last log received and which subs are connected.
Uses an api call to sentinel and log analytics query.

Fixes #

## Proposed Changes

  -
  -
  -


morshabi:
@lior-tamir is this PR still relevent?

liatlishams:
@lior-tamir Hi, still relevant? I am closing it for now.

PR #692, opened 2020-05-18
Fixed Azure Deploy button URL

happy-jo:
Fixes #

## Proposed Changes

Fixed Deploy URL


dicolanl:
@happy-jo this url is correct why the update?

we have two urls.  1 for Commercial (azure.com) and one for gov (azure.us)

happy-jo:
@dicolanl 
The deploy button for was public cloud and the link was for gov cloud.
The public deploy button did not work for my public azure.

PR #716, opened 2020-05-29
Update DistribPassCrackAttempt.yaml

aadelnabil:
The rule isn’t firing as expected when we tested it against the configured thresholds in the rule. After some digging in the rule logic I came to a conclusion that the rule logic is flowed and will never fire as due to the “summarize” statement highlighted below the location count will always be equal to 1.

Fixes #

## Proposed Changes
I have updated the summarize statement and then added a join statement to produce the final results
  -
  -
  -


aadelnabil:
Thanks for your review and feedback, I compared the results from the above query with the one I submitted and it is almost the same which is great with one exception. The signin count in the query you suggested is the total signin count for all failed attempts not the signin count for each location. I will play around with the suggested query as it is simpler than the one I submitted to try to figure get the correct signin count for each location and submit an update.

shainw:
@aadelnabil - Wondering if you are able to update your contribution so we can finalize it?  Thanks again for contributing.  Shain

aadelnabil:
@shainw apologies, got busy over the past time. I will set time aside to finalize this.

PR #722, opened 2020-06-03
Fixup missing metadata

shainw:
Fixes #

## Proposed Changes

  -
  -
  -


Detections/CommonSecurityLog/MultiVendor-PossibleDGAContacts.yaml
```diff
@@ -37,7 +37,7 @@ queryPeriod: 6h
 triggerOperator: gt
 triggerThreshold: 0
 tactics:
-  - Command and Control
+  - CommandandControl
```
laithhisham, line 40: CommandAndControl case sensitive

shainw: Done, RareClientFileAccess.yaml has been moved back to Hunting, not sure who\how it got moved, so no need for the query and trigger items for that one.

PR #723, opened 2020-06-03
Create PAN_Parser.csl

happy-jo:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Parsers/PAN_Parser.csl
+++ b/Parsers/PAN_Parser.csl
@@ -1,0 +1,38 @@
+// Palo Alto Networks
+// Last Updated Date: June 3, 2020
+//
+// This parser parses Syslog events for Palo Alto Network event logs. (Mainly the Traffic, Threat, Trap, and OS logs)
+// https://docs.paloaltonetworks.com/pan-os/8-1/pan-os-admin/monitoring/use-syslog-for-monitoring/syslog-field-descriptions.html
+//
+// Parser Notes:
+// 1. This parser assumes logs are collected into the CommonSecurityLog table.
+//
+// Usage Instruction : 
+// Paste below query in log analytics, click on Save button and select as Function from drop down by specifying function name and alias.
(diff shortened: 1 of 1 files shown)


Parsers/PAN_Parser.csl
```diff
@@ -0,0 +1,38 @@
...
+// 1. This parser assumes logs are collected into the CommonSecurityLog table.
+//
+// Usage Instruction : 
+// Paste below query in log analytics, click on Save button and select as Function from drop down by specifying function name and alias.
+// Functions usually takes 10-15 minutes to activate. You can then use function alias from any other queries (e.g. PAN_CL | take 10).
+//
+//
+
+CommonSecurityLog
+| where DeviceVendor == "Palo Alto Networks"
```
shainw, line 17: When running this on PANOS logs, as you state this only appears to parse  Traffic, Threat, Trap and OS logs.  For faster processing, I would suggest that you write this as a function, do the parsing on only rows that have Traffic, Threat, Trap and OS in the Activity field and then union that with unparsed rows that don't have those in the Activity field.  This should make it more performant over larger data sets.

Parsers/PAN_Parser.csl
```diff
@@ -0,0 +1,38 @@
+// Palo Alto Networks
+// Last Updated Date: June 3, 2020
+//
+// This parser parses Syslog events for Palo Alto Network event logs. (Mainly the Traffic, Threat, Trap, and OS logs)
```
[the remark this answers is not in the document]
happy-jo: Thanks Shain. I'll update it a little later this week along with the others I have commited.

Parsers/PAN_Parser.csl
happy-jo: will do and thanks.
Currently that is all that is being forwarded to our ALA workspace, so that is what I'm building the parser off of in hopes that someone will add to what I have committed.

shainw: Will merge for now and if you want to update later, please feel free.  Thanks for the addition!

PR #735, opened 2020-06-08
Fixing output fields name to entity field names

YaronFruchtmann:
Fixes #

## Proposed Changes

  -
  -
  -


Exploration Queries/InputEntity_File/HostwithFile.yaml
```diff
@@ -0,0 +1,30 @@
+﻿Id: 62527635-bc5a-4233-bb93-e4eb4e60bb70
```
shainw, line 1: Looks like you left in the duplicated GUID here from the FilesOnHost.yaml.

PR #757, opened 2020-06-17
Added Isolate-AzureStorageAccount playbook.

rjygraham:
## Proposed Changes

  - Added Isolate-AzureStorageAccount playbook


rjygraham:
@rjygraham updated the foreach setting the storageid variable with recommended concurrency setting.

PR #763, opened 2020-06-18
Performed full deployment and confirmed no errors

9b:
Apologies for the multiple pulls related to this work. The README for the playbooks testing process was not clear and I was unable to follow it. I've since found a work-around using the "deploy to Azure" links and then performing an edit. During this process, I was able to identify the templates were missing a `contentVersion` field that is required for the schema. I deployed several of these end-to-end and confirmed success on an incident run. 

Fixes #

  - Added missing `contentVersion` to the top-level structure. 

## Proposed Changes

  -
  -
  -


9b:
Forgive me @dicolanl! I have a process I can perform locally now, so I won't be cowboy coding from here.

PR #795, opened 2020-06-26
ACN_CD_QualysVM_ConnecterFix

chicduong:
Fixes #

## Proposed Changes

  - Increase default timer trigger to 60 mins
  - Added host.json to increase execution timeout from 5m to 10m
  - Updated the data connector json, azure deploy json, and Azure Function zip accordingly


DataConnectors/Qualys VM/host.json
```diff
@@ -0,0 +1,11 @@
+{
+    "version": "2.0",
+    "functionTimeout": "00:10:00",
+    "managedDependency": {
+      "Enabled": true
+    },
+    "extensionBundle": {
+      "id": "Microsoft.Azure.Functions.ExtensionBundle",
+      "version": "[1.*, 2.0.0)"
```
preetikr, line 9: The opening bracket [  and closing parenthesis  ) - are not matching

DataConnectors/Qualys VM/host.json
```diff
@@ -0,0 +1,11 @@
+{
```
preetikr, line 1: Shouldn't there be corresponding changes in run.ps1 to use the new properties defined here? 
What does version = 2.0 signify here?

chicduong: This is from a MSFT documentation reference.

https://docs.microsoft.com/en-us/azure/azure-functions/functions-host-json

DataConnectors/Qualys VM/host.json
chicduong: This is taken from the sample from the MFST documentation and also compared against out data connectors (Zoom) 

https://docs.microsoft.com/en-us/azure/azure-functions/functions-host-json

PR #818, opened 2020-07-03
Update Office_MailForwarding.yaml

shainw:
fixing fwdDest and also parsing ClientIP and Port updates, along with inclusion of ClientIP_ field that is availalbe in some cases.

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/OfficeActivity/Office_MailForwarding.yaml
+++ b/Detections/OfficeActivity/Office_MailForwarding.yaml
@@ -24,25 +24,41 @@ query: |
   OfficeActivity
   | where TimeGenerated >= ago(timeframe)
   | where Operation =~ "Set-Mailbox"
-  | where Parameters contains "ForwardingSmtpAddress"
+  | where Parameters has "ForwardingSmtpAddress"
   | extend parsed = parse_json(Parameters)
-  | extend parameterName = parsed[1].Name, fwdingDestination = tostring(parsed[1].Value)
+  | mv-expand parsed
+  | where parsed.Name == "ForwardingSmtpAddress"
+  | extend parameterName = tostring(parsed.Name), fwdingDestination = tostring(parsed.Value)
   | where isnotempty(fwdingDestination)
   | extend ClientIPOnly = case( 
-  ClientIP has ".", tostring(split(ClientIP,":")[0]), 
-  ClientIP has "[", tostring(trim_start(@'[[]',tostring(split(ClientIP,"]")[0]))),
+  ClientIP has "." and ClientIP has ':', tostring(split(ClientIP,":")[0]), 
+  ClientIP has "." and ClientIP has '-', tostring(split(ClientIP,"-")[0]), 
+  ClientIP has ']-', tostring(trim_start(@'[[]',tostring(split(ClientIP,"]")[0]))),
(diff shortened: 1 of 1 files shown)


Detections/OfficeActivity/Office_MailForwarding.yaml
```diff
@@ -24,25 +24,41 @@ query: |
...
+  isnotempty(ClientIP_), ClientIP_,
+  "IP Not Available"
   )
-  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), DistinctUserCount = dcount(UserId), UserId = makeset(UserId), 
-  Ports = makeset(Port), EventCount = count() by fwdingDestination, ClientIP = ClientIPOnly 
+  | extend UserId = iff(isempty(UserId), UserId_, UserId)
+  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), DistinctUserCount = dcount(UserId), UserId = make_set(UserId), 
+  Ports = make_set(Port), EventCount = count() by fwdingDestination, ClientIP = ClientIPOnly 
   | where DistinctUserCount > 1
   | mvexpand UserId
```
ashwin-patil, line 61: - can we also do mv-expand Ports since we are doing tostring and distinct afterwards anyways ? for multi-ports it 
- can we change this to mv-expand instead of mvexpand ? although it works, you have used mv-expand throughout the query ?

Detections/OfficeActivity/Office_MailForwarding.yaml
```diff
@@ -24,25 +24,41 @@ query: |
...
-  ClientIP
+  ClientIP has "." and ClientIP has ':', tostring(split(ClientIP,":")[1]), 
+  ClientIP has "." and ClientIP has '-', tostring(split(ClientIP,"-")[1]), 
+  ClientIP has ']-', tostring(split(ClientIP,"]-")[1]), 
+  ClientIP has ']:', tostring(split(ClientIP,"]:")[1]), 
+  isempty(ClientIP) and ClientIP_ has "." and ClientIP_ has ':', tostring(split(ClientIP_,":")[1]), 
+  isempty(ClientIP) and ClientIP_ has "." and ClientIP_ has '-', tostring(split(ClientIP_,"-")[1]), 
+  isempty(ClientIP) and ClientIP_ has ']-', tostring(split(ClientIP_,"]-")[1]),
+  isempty(ClientIP) and ClientIP_ has ']:', tostring(split(ClientIP_,"]:")[1]),
+  isnotempty(ClientIP), ClientIP,
```
ashwin-patil, line 53: Do you want to do this check of ClientIP under case for ClientIP so value IP not available gets assigned to ClientIP instead of Port ? not much difference in functionality.

Detections/OfficeActivity/Office_MailForwarding.yaml
shainw: Done for the mv-expand change, but the context of this is users and if we to mv-expand on ports (which is not a primary entity) then it causes duplicates which we want to avoid.

Detections/OfficeActivity/Office_MailForwarding.yaml
shainw: Good catch, done.

PR #833, opened 2020-07-08
Sentinel UEBA Hunting Queries

itay6588:
Fixes #

## Proposed Changes

  -
  -
  -


itay6588:
@YaronFruchtmann  thank you :)

Hunting Queries/BehaviorAnalytics/Anomalous Account Creation.yaml
```diff
@@ -0,0 +1,26 @@
...
+query: |
+
+ BehaviorAnalytics
+ | where ActionType == "Add user"
+ | where ActivityInsights.FirstTimeUserPerformedAction == True or ActivityInsights.FirstTimeTenantPerformedAction == True
+ | join(
+ AuditLogs
+ ) on $left.SourceRecordId == $right._ItemId
+ | mv-expand TargetResources
+ | extend Target = tostring(TargetResources.userPrincipalName)
```
shainw, line 23: We should fix the parsing of the UserName, UserPrincipalName and TargetUser field for easy connection to additional data types and future use in the Investigation Graph.  UserName comes thru as <username> but in some cases comes thru as <username>_domain.com#EXT#.   You can parse these fields when it has #EXT# in it using this example - | extend TargetUserName = replace("_","@",tostring(split(TargetUserPrincipalName, "#")[0]))

Hunting Queries/BehaviorAnalytics/Anomalous Account Creation.yaml
```diff
@@ -0,0 +1,26 @@
+id: 0454c8d8-d0a6-42a4-8d03-f5b4fdcbd173
+name: Anomalous Account Creation
```
shainw, line 2: We may want to consider a name that includes the type of account created, meaning Windows or Linux or Azure, in this case Azure AD.  The reason is that this specific table doesn't align with a source type being from AD or AAD.

Hunting Queries/BehaviorAnalytics/Anomalous Account Manipulation.yaml
```diff
@@ -0,0 +1,33 @@
...
+    dataTypes:
+      - BehaviorAnalytics
+      - AuditLogs
+tactics:
+  - Persistence
+relevantTechniques:
+  - T1098
+query: |
+
+ let critical = dynamic(['9b895d92-2cd3-44c7-9d02-a6ac2d5ea5c3','c4e39bd9-1100-46d3-8c65-fb160da0071f','158c047a-c907-4556-b7ef-446551a6b5f7','62e90394-69f5-4237-9190-012177145e10','d29b2b05-8046-44ba-8758-1e26182fcf32','729827e3-9c14-49f7-bb1b-9608f156bbb8','966707d0-3269-4727-9be2-8c3a10f19b9d','194ae4cb-b126-40b2-bd5b-6091b380977d','fe930be7-5e62-47db-91af-98c3a49a38b1']);
```
shainw, line 16: A comment here about what these GUIDs represent would be good.  It is not clear how these are derived.

Hunting Queries/BehaviorAnalytics/Anomalous Account Manipulation.yaml
```diff
@@ -0,0 +1,33 @@
...
+ | mv-expand AdditionalDetails
+ | mv-expand TargetResources
+ | where AdditionalDetails.key == "UserPrincipalName"
+ | mv-expand TargetResources
+ | extend RoleId = tostring(TargetResources.modifiedProperties[0].newValue)
+ | extend RoleName = tostring(TargetResources.modifiedProperties[1].newValue)
+ | where RoleId in (critical,high)
+ | extend TargetId = tostring(TargetResources.id)
+ | extend Target = tostring(TargetResources.userPrincipalName)
+ | where isnotempty(RoleId) or isnotempty(RoleName)
```
shainw, line 29: For performance, maybe move this line up in the processing order

Hunting Queries/BehaviorAnalytics/Anomalous Failed Logon.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - T1110.001
+query: |
+
+ BehaviorAnalytics
+ | where ActivityType == "LogOn"
+ | where UsersInsights.BlastRadius == "High"
+ | join (
+ SigninLogs
+ ) on $left.SourceRecordId == $right._ItemId
+ | where Status.errorCode == 50126
```
shainw, line 22: This should be a filter inside the join?

Hunting Queries/BehaviorAnalytics/Anomalous Account Creation.yaml
itay6588: Done

Hunting Queries/BehaviorAnalytics/Anomalous Failed Logon.yaml
itay6588: done

Hunting Queries/BehaviorAnalytics/Anomalous Account Manipulation.yaml
itay6588: done

Hunting Queries/BehaviorAnalytics/Anomalous Account Manipulation.yaml
itay6588: added criteria for each category

Hunting Queries/BehaviorAnalytics/Anomalous Account Creation.yaml
itay6588: added to all relevant queries

Hunting Queries/BehaviorAnalytics/Anomalous Role Assignment.yml
```diff
@@ -0,0 +1,30 @@
...
+    dataTypes:
+       - BehaviorAnalytics
+       - AuditLogs
+tactics:
+  - Persistence
+relevantTechniques:
+  - T1098
+query: |
+
+ let critical = dynamic(['9b895d92-2cd3-44c7-9d02-a6ac2d5ea5c3','c4e39bd9-1100-46d3-8c65-fb160da0071f','158c047a-c907-4556-b7ef-446551a6b5f7','62e90394-69f5-4237-9190-012177145e10','d29b2b05-8046-44ba-8758-1e26182fcf32','729827e3-9c14-49f7-bb1b-9608f156bbb8','966707d0-3269-4727-9be2-8c3a10f19b9d','194ae4cb-b126-40b2-bd5b-6091b380977d','fe930be7-5e62-47db-91af-98c3a49a38b1']);
```
shainw, line 16: The comments you added in Anomalous Account Manipulation.yaml should be added here also.

PR #839, opened 2020-07-09
ACN_CD_TemplateUpdates

chicduong:
Fixes #

## Proposed Changes

  - added examples of code to prevent null exception errors, from empty arrays, when posting to the ALA API


chicduong:
PowerShell files uploaded. Python runs on a Linux kern and do not have these files. Since Python function Apps cannot be created in Azure Functions, only in VSCode or VS, the Azure Functions extensions automatically creates the files needed for the Function. I have checked the creation of the Function App in VSCode as it does not create the files.

chicduong:
I created folders for each template

PR #847, opened 2020-07-09
ACN_CD_QualysVMConnectorUpdates

chicduong:
Fixes #

## Proposed Changes

  - Add additional parameter to allow users to append additional filter parameters to the URI
  - changes the detection_processed_after parameter to vm_processed_after parameter based on Qualys recommendation
  - Accordingly updated data connector page, azure deploy json, and zip package


DataConnectors/Qualys VM/azuredeploy_QualysVM_API_FunctionApp.json
```diff
@@ -24,7 +24,11 @@
           },
           "Uri": {
               "type": "string",
-              "defaultValue": "https://<API Server URL>/api/2.0/fo/asset/host/vm/detection/?action=list&detection_processed_after="
+              "defaultValue": "https://<API Server URL>/api/2.0/fo/asset/host/vm/detection/?action=list&vm_processed_after="
+          },
+          "FilterParameters": {
+              "type": "string",
+              "defaultValue": "&truncation_limit=1000"
```
preetikr, line 31: How can the default value of 1000 be overwritten if it's passed in the filter parameters?

chicduong: They simple just change the 1000 to whatever value they would like to before clicking Purchase (Deploy)

PR #886, opened 2020-07-21
Fixing ARM Template

chboeh:
Connection strings were incorrect; updated everything to the latest version and order.

Fixes #

## Proposed Changes

  -
  -
  -


chboeh:
Confirmed deployment with this template. -Everything works correctly.

image

PR #888, opened 2020-07-22
Replacing ProcessList with new version with more useful information

shainw:
Fixes #

## Proposed Changes

  -
  -
  -


YaronFruchtmann:
1. The file fails 2 of the automatic tests
2. use materialize on "let processEvents" to avoid running twice, and also so that "processes" and "parnts" run on the same dataset.
2+. union: consider using "isfuzzy=true" JIC?
3. summarize: maybe remove the "by Computer". This is a host insight, and baseQuery already filtered by Host. variations in Computer capitalization would result in line duplication

shainw:
> 1. The file fails 2 of the automatic tests
> 2. use materialize on "let processEvents" to avoid running twice, and also so that "processes" and "parnts" run on the same dataset.
> 3. what is the time scope of "let processEvents"?
> 4. union: consider using "isfuzzy=true" JIC?
> 5. summarize: maybe remove the "by Computer". This is a host insight, and baseQuery already filtered by Host. variations in Computer capitalization would result in line duplication

Fixed all.

PR #903, opened 2020-07-24
Fixup parsing

shainw:
## Fixes

  - updating parsing of Computer so we can match properly depending on data
  - made sure that all of custom entity outputs to support bookmarks if so desired

## Proposed Updates

  - added in matching on _ResourceId and SourceComputerId
  - added Activities in where appropriate related to the Base Query available in the Insight

### DIFF
--- a/Insights/InputEntity_Host/ActionsOnAccounts.yaml
+++ b/Insights/InputEntity_Host/ActionsOnAccounts.yaml
@@ -3,6 +3,13 @@ DataTypes:
   - DataType: SecurityEvent
 Type: KQL
 Provider: Sentinel
+RequiredInputFieldsSets: 
+ - - Host_HostName
+   - Host_NTDomain
+ - - Host_HostName
+   - Host_DnsDomain
+ - - Host_AzureID
+ - - Host_OMSAgentID
 BaseQuery: |
   let GetAccountActions = (v_Host_Name:string, v_Host_NTDomain:string, v_Host_DnsDomain:string, v_Host_AzureID:string, v_Host_OMSAgentID:string){
   SecurityEvent
@@ -34,11 +41,6 @@ BaseQuery: |
   | extend timestamp = TimeGenerated, HostCustomEntity = Computer, AccountCustomEntity = TargetAccount
   };
   GetAccountActions('{{Host_HostName}}', '{{Host_NTDomain}}', '{{Host_DnsDomain}}', '{{Host_AzureID}}', '{{Host_OMSAgentID}}')
-RequiredInputFieldsSets: 
- - - Host_HostName
-   - Host_NTDomain
- - - Host_HostName
-   - Host_DnsDomain
 # The queries for the insights.
 Insights:
  Id: e29ee1ef-7445-455e-85f1-269f2d536d61
@@ -127,5 +129,5 @@ Activities:
      title: "An account was deleted on this host"
      content: "On '{{Computer}}' the account '{{TargetAccount}}' was deleted by '{{AddedBy}}'"
      queryDefinitions:
-       filter: where EventID == 4722
+       filter: where EventID == 4726
        summarizeBy: Computer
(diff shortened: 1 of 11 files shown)


Insights/InputEntity_Host/ActionsOnAccounts.yaml
```diff
@@ -4,29 +4,36 @@ DataTypes:
...
+  or v_Host_OMSAgentID == SourceComputerId
+  | project TimeGenerated, EventID, Activity, Computer, TargetAccount, TargetUserName, TargetDomainName, TargetSid, SubjectUserName, SubjectUserSid, _ResourceId, SourceComputerId
   | extend AddedBy = SubjectUserName
   // Future support for Activities
-  | extend timestamp = TimeGenerated, HostCustomEntity = PassedInName, AccountCustomEntity = TargetAccount
+  | extend timestamp = TimeGenerated, HostCustomEntity = Computer, AccountCustomEntity = TargetAccount
   };
-  GetAccountActions('{{Host_HostName}}', '{{Host_NTDomain}}', '{{Host_DnsDomain}}')
+  GetAccountActions('{{Host_HostName}}', '{{Host_NTDomain}}', '{{Host_DnsDomain}}', '{{Host_AzureID}}', '{{Host_OMSAgentID}}')
 RequiredInputFieldsSets:
```
YaronFruchtmann, line 37: This insight would also run if we have
- - Host_HostName
  -  Host_DnsDomain
-  - Host_OMSAgentID
-  - Hoat_AzureID

Maybe worth adding? (why do we need this field? isn't this a directly obtainable from the input fields?)

Insights/InputEntity_Host/ActionsOnAccounts.yaml
```diff
@@ -103,4 +110,22 @@ Insights:
...
-  Query: "project TimeGenerated, EventID, Activity, Computer, TargetAccount, TargetUserName, TargetDomainName, TargetSid, SubjectUserName, SubjectUserSid | order by TimeGenerated desc"
+  Query: "project TimeGenerated, EventID, Activity, Computer, TargetAccount, TargetUserName, TargetDomainName, TargetSid, SubjectUserName, SubjectUserSid, _ResourceId, SourceComputerId, timestamp, HostCustomEntity, AccountCustomEntity | order by TimeGenerated desc"
+
+Activities:
+ enabledByDefault: true
+ items: 
+   - id: 307c85ee-39a2-4da3-952e-4fd79aa46d3a
+     description: Account created on Host
+     title: "An account was created on this host"
+     content: "On '{{Computer}}' the account '{{TargetAccount}}' was created by '{{AddedBy}}'"
```
YaronFruchtmann, line 121: This is a matter of style, but since this is {{Computer}}'s entity page do we need to output the computer?

Insights/InputEntity_Host/ActionsOnAccounts.yaml
```diff
@@ -103,4 +110,22 @@ Insights:
...
+     queryDefinitions:
+       filter: where EventID == 4720
+       summarizeBy: Computer
+   - id: 31529548-dbd2-4d5d-8270-710330cdcec7
+     description: Account deleted on Host
+     title: "An account was deleted on this host"
+     content: "On '{{Computer}}' the account '{{TargetAccount}}' was deleted by '{{AddedBy}}'"
+     queryDefinitions:
+       filter: where EventID == 4722
```
YaronFruchtmann, line 130: I think you mean 4726?

Insights/InputEntity_Host/ActionsOnAccounts.yaml
shainw: I think we have both ways and let customers provide us feedback, we have some private preview customers who will do just that. :)

Insights/InputEntity_Host/ActionsOnAccounts.yaml
shainw: Yes, fixed.

Insights/InputEntity_Host/ActionsOnAccounts.yaml
shainw: Done.

PR #908, opened 2020-07-27
replace DevOps template

chinguyen1:
Fixes #

## Proposed Changes

Helping Crist Romeo push changes in the DevOps template. Replace the old zip with a more updated one.


chinguyen1:
@CrisRomeo - looks like there's an issue with the template

chinguyen1:
@CrisRomeo I think it's due to the fact some of your doc links contains localized "en-us" part. Would you mind fixing it & I'll push the changes again?

PR #916, opened 2020-07-29
Update ForgeRock_CEF.json

javaservlets:
now that we have the ForgeRock parser at https://github.com/Azure/Azure-Sentinel/blob/master/Parsers/ForgeRockParser.txt, we've added a 3rd sample query that leverages it

Fixes #

## Proposed Changes

  -
  -
  -


javaservlets:
I'm retracting this PR given it seems opening it delayed us in the publishing queue

PR #923, opened 2020-07-30
Zscaler (NSS) Parsers

dicolanl:
Fixes #

## Proposed Changes
New parsers for Zscaler NSS logs.


dicolanl:
@ashwin-patil I separated them because customer could send one or the other. which means they may also what to search/query one or the other only and not included results.  Web logs are proxy which would be different use case than FW logs.

ashwin-patil:
> @ashwin-patil I separated them because customer could send one or the other. which means they may also what to search/query one or the other only and not included results. Web logs are proxy which would be different use case than FW logs.

Got it, makes sense. Approved

PR #933, opened 2020-08-03
3AUG

JimmyJamCABD:
Fixes #
Removed TemplateId and Schema from workbook metadata

## Proposed Changes

  -
  -
  -


preetikr:
The relevant changes for this is already approved and merged as PR #975 and #929 so dropping this PR (no action).

PR #941, opened 2020-08-04
DNS detections bug bash fixes.

petebryan:
Minor fixes to DNS based detections as part of bug bash

### DIFF
--- a/Detections/DnsEvents/DNS_HighNXDomainCount_detection.yaml
+++ b/Detections/DnsEvents/DNS_HighNXDomainCount_detection.yaml
@@ -1,5 +1,5 @@
 ﻿id: a0907abe-6925-4d90-af2b-c7e89dc201a6
-name: Potential DGA detected.
+name: Potential DGA detected
 description: |
   'Identifies clients with a high NXDomain count which could be indicative of a DGA (cycling through possible C2 domains
   where most C2s are not live). Alert is generated when a new IP address is seen (based on not being seen associated with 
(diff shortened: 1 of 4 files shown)


Detections/DnsEvents/DNS_HighNXDomainCount_detection.yaml
```diff
@@ -1,8 +1,8 @@
 ﻿id: a0907abe-6925-4d90-af2b-c7e89dc201a6
-name: Rare high NXDomain count
+name: Potential DGA detected.
```
shainw, line 2: removing the trailing period in name, we also may want to consider renaming the file to something like DNS_PotentialDGADetected.yaml.

Detections/DnsEvents/DNS_TorProxies.yaml
```diff
@@ -1,7 +1,7 @@
 ﻿id: a83ef0f4-dace-4767-bce3-ebd32599d2a0
-name: DNS tor proxies
+name: DNS event for ToR proxies
```
shainw, line 2: maybe use "related" instead of "for" in the name.  So, Name would be "DNS events related to ToR proxies".  This would go in line with what we have for the Miners detection also and I think is more accurate.

petebryan:
> Couple of changes, file rename is something we can discuss.

Let me know on the file change. Nothing was done here as the OneNote had this as something that was needed additional PG discussion so was out of scope for now. If we are going to change file names we should update this.

PR #944, opened 2020-08-04
Add Github Action to Validate the WorkbooksMetadata json schema

erant10:
## Proposed Changes
Apply a new github action that triggers upon PR submission which contains changes to the WorkbooksMetadata file

The validation will reflect whether or not the schema is valid, i.e. 
- it does not contain illegal properties 
- is has the proper data type

The  `metadata_schma.json` is the schema that the action will use to verify the json file.

### DIFF
--- a/Workbooks/WorkbooksMetadata.json
+++ b/Workbooks/WorkbooksMetadata.json
@@ -11,7 +11,7 @@
     "templateRelativePath": "AzureActivity.json",
     "subtitle": "",
     "provider": "Microsoft",
-    "someIllegalProperty": "abc"
+    "featureFlag": "MyTestFlag"
   },
   {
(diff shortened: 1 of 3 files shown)


Workbooks/WorkbooksMetadata.json
```diff
@@ -10,7 +10,8 @@
     "title": "Azure Activity",
     "templateRelativePath": "AzureActivity.json",
     "subtitle": "",
-    "provider": "Microsoft"
+    "provider": "Microsoft",
+    "featureFlag": "abc"
```
nazang: What does the feature flag abc mean?

erant10: I removed it for now, but its just for testing..

erant10:
abandoned PR to use pipelines instead of actions

PR #946, opened 2020-08-05
Add new validations for workbooksMetadata

erant10:
Expand the `JsonFileValidator` to validate the schema for `WorkbooksMetadata.json` file

The validation will fail if the schema is not valid, i.e.:
- If WorkbooksMetadata contains illegal properties
- If the data types do not match (int instead of string etc..)


.script/jsonFileValidator.ts
```diff
@@ -1,13 +1,26 @@
...
 export async function IsValidJsonFile(filePath: string): Promise<ExitCode> {
-  JSON.parse(fs.readFileSync(filePath, "utf8"));
+  const json = JSON.parse(fs.readFileSync(filePath, "utf8")); 
+  if (filePath.endsWith('WorkbooksMetadata.json')) {
+    validateWorkbookSchema(json);
+  }
   return ExitCode.SUCCESS;
 }
 
+function validateWorkbookSchema(workbooksMetadataJson: object) {
```
sagamzu, line 15: I think you can change this function to general for all schemas, by adding schema file path as parameter

.script/jsonFileValidator.ts
```diff
@@ -1,13 +1,26 @@
...
+  const schema = JSON.parse(fs.readFileSync('.script/utils/WorkbooksMetadataSchema.json', 'utf8'));
+  var validationResult = new Validator().validate(workbooksMetadataJson, schema);
+  if (!validationResult.valid) {
+    let errorMsg = validationResult.errors.map(err => err.message).join(", ");
+    throw new SchemaError(errorMsg, schema)
```
sagamzu, line 20: What we will see in the logs? both errorMsg & schema or just errorMsg?

erant10: just errorMsg.  for example:
- `additionalProperty "someIllegalProperty" exists in instance when not allowed`
- `requires property "workbookKey"`

.script/utils/schemas/WorkbooksMetadataSchema.json
```diff
@@ -0,0 +1,75 @@
+{
+    "type": "array",
+    "items": [
+        {
+            "type": "object",
+            "properties": {
+                "workbookKey": {
```
Amitbergman, line 7: Is there an option to assert that all the keys are distinct? To make sure no one added a workbook with existing key?

.script/utils/schemas/WorkbooksMetadataSchema.json
```diff
@@ -0,0 +1,75 @@
...
+                },
+                "dataConnectorsDependencies": {
+                    "type": "array",
+                    "items": [
+                        {
+                            "type": "string"
+                        }
+                    ]
+                },
+                "previewImagesFileNames": {
```
Amitbergman, line 32: The image file name must a path to a png image. Can we verify this?

.script/utils/schemas/WorkbooksMetadataSchema.json
```diff
@@ -0,0 +1,75 @@
+{
+    "type": "array",
+    "items": [
+        {
+            "type": "object",
+            "properties": {
+                "workbookKey": {
+                    "type": "string"
+                },
+                "logoFileName": {
```
Amitbergman, line 10: This must be a valid path to an svg file. Can we verify it?

.script/utils/schemas/WorkbooksMetadataSchema.json
```diff
@@ -0,0 +1,75 @@
...
+                        }
+                    ]
+                },
+                "version": {
+                    "type": "string"
+                },
+                "title": {
+                    "type": "string"
+                },
+                "templateRelativePath": {
```
Amitbergman, line 46: This must be a valid path to a .json file. Can we verify it?

.script/utils/schemas/WorkbooksMetadataSchema.json
erant10: From what I understand it's [not possible](https://stackoverflow.com/questions/24763759/make-sure-item-property-in-array-is-unique-in-json-schema) using a json schema. However we can create a custom validation for this (with ts). I listed the validations you mentioned here and I will create a separate validator job specifically for workbooks.

.script/workbooksValidations/previewImageValidator.ts
```diff
@@ -0,0 +1,33 @@
+import { WorkbookValidationError } from "../utils/ValidationError";
+import { WorkbookMetadata } from "../utils/WorkbookMetadata";
+
+const previewImageRegex: RegExp = /(black|white|Black|White)(\d*?).png/;
+const blackPreviewImageRegex: RegExp = /(black|Black)(\d*?).png/;
+const whitePreviewImageRegex: RegExp = /(white|White)(\d*?).png/;
+
+function isAllValidFileNames(previewImagesFileNames: Array<string>): boolean {
+    return previewImagesFileNames.every((previewImageFileName: string) => previewImageRegex.test(previewImageFileName));
```
mpo-sec, line 9: Couldn't you just check for matching the white or black regex here instead of adding a third one?
```
(previewImageFileName: string) => blackPreviewImageRegex.test(previewImageFileName) | whitePreviewImageRegex.test(previewImageFileName)
```

.script/workbooksValidations/previewImageValidator.ts
```diff
@@ -0,0 +1,33 @@
...
+        if (whitePreviewImageRegex.test(filename)) whiteImageCount++;
+        if (blackPreviewImageRegex.test(filename)) blackImageCount++;
+    });
+    return blackImageCount === 0 || whiteImageCount === 0;
+}
+
+export function isValidPreviewImageFileNames(items: Array<WorkbookMetadata>) {
+  items.forEach((workbookMetadata: WorkbookMetadata) => {
+    if (!isAllValidFileNames(workbookMetadata.previewImagesFileNames)) {
+      throw new WorkbookValidationError(`Invalid Preview Images for workbook ${workbookMetadata.workbookKey}. Filename must contain either "Black" or "White" and must end with .png`);
```
mpo-sec, line 25: Technically you can have have Black or White and .png like this errror says and still not match the regexes like `AwsBlackExample.png`
It could be worth putting a more specific message so a 3P knows specifically what to fix.
I guess the format also can include numbers in the filename too like `AwsBlack1.png`


Another option is maybe just split `isAllValidFileNames` into a fewer more simple tests:
- file extension is .png
- filename ends with a number, or nothing
- filename ends with black/white ignoring the number, or just contains black/white.  Not sure the exact requirement

erant10: I realized the requirement is that file names need to *include* either white or black, so I fixed this. also split the png validation to a separate function as you suggested. Thanks for the catch!

.script/workbooksValidations/previewImageValidator.ts
erant10: definitely. however I decided to get rid of the RegExp entirely because the `includes` method actually is good enough for this requirement (see other comment).

.azure-pipelines/workbooksValidations.yaml
```diff
@@ -0,0 +1,12 @@
+jobs:
```
sagamzu, line 1: Please add reference to this validation in "https://github.com/Azure/Azure-Sentinel/blob/master/azure-pipelines.yml" file (the entry point to validation), otherwise this validation won't run.

After adding this change, you will see the new vallidation in "Check" tabs

.azure-pipelines/workbooksValidations.yaml
```diff
@@ -0,0 +1,12 @@
...
+- job: "WorkbooksValidations"
+  pool:
+    vmImage: 'Ubuntu 16.04'
+  steps:
+  - task: Npm@1
+    displayName: 'npm install'
+    inputs:
+      verbose: false
+      command: 'install'
+  - script: 'npm run tsc && node .script/workbooksValidator.js'
```
sagamzu, line 11: we have 2 options to add validation. I think both options are good. Just add this comment FYI (in case you miss it)

1. run all validation in the same script, like you did - we run over the file in one loop
2. add different script foeach validation with different display name - we will run over the files in different loop, but will see the output foreach validation under different titles\jobs (maybe it clearer for users)

.azure-pipelines/workbooksValidations.yaml
erant10: Done. Thanks for the catch!

.azure-pipelines/workbooksValidations.yaml
erant10: Got it. I think I will leave all the Metadata validations under one script, and when we add other workbook validations we will add a separate script file for those validations..

erant10:
@sagamzu I noticed that when a validation fails the message that is displayed is `An error occurred, please open an issue` and only when clicking on `view more details on azure pipelines` it shows the reason for the failure.. (for example the jsonFileValidator below
image
Is this the desired behavior? I would think it makes more sense to show the validation message on the failed check and not in the pipeline log...

.azure-pipelines/workbooksValidations.yaml
```diff
@@ -0,0 +1,12 @@
...
+- job: "WorkbooksValidations"
+  pool:
+    vmImage: 'Ubuntu 16.04'
+  steps:
+  - task: Npm@1
+    displayName: 'npm install'
+    inputs:
+      verbose: false
+      command: 'install'
+  - script: 'npm run tsc && node .script/workbooksMetadataValidator.js'
```
sagamzu, line 11: Please note that you add reference only to "workbooksMetadataValidator.js", means the other validations (like uniqueWorkbookKeyValidator) won't run.

you need to add here links to the other script as well.

erant10: good point. Those are actually utility functions used by the `workbooksMetadataValidator`. I renamed them to `checkers` to prevent future confusion.

.script/utils/schemas/WorkbooksMetadataSchema.json
```diff
@@ -0,0 +1,75 @@
+{
```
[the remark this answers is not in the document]
erant10: fixed

sagamzu:
@erant10 agree with you. good catch 👍 
I never read this message, I always licked on the view details link...

In addition  I think the message "An error occurred, please open an issue" should display only in case the validation failed for some reason we have bug in our validation and not in case the validation works and found a real  bugs that need to fix. 

I think that It is better to fix this issue in anther PR

erant10:
@sagamzu Thanks! I'll open a separate PR for this soon

PR #951, opened 2020-08-07
new playbook bulk close incidents

swiftsolves-msft:
The following new playbook can use a KQL query to selective bulk close or change Azure Sentinel incidents.

Fixes #

## Proposed Changes

  -
  -
  -


swiftsolves-msft:
changed design and name going to use new push req

Issue #925, opened 2020-07-31
OKTA add Domain name column 


Yaniv-Shasha:
Customer wants to ingest for multipla OKTA domains into the same OKTA_CL table 
to be able to differentiate between domains we need to expose OKTA domain name as property.

solution : add the above lines

```powershell
if ($recordCount -gt 0) {
    Write-Output "$recordCount record(s) are avaliable as of $startDate"
    $domain1 = $env:uri
    $domain = $domain1.Replace("api/v1/logs?since=","")
    $response  | Add-Member -MemberType NoteProperty -Name "Domain" -Value  $domain
    $json = $response | ConvertTo-Json -Depth 5

   Write-Output $json
   
    Post-LogAnalyticsData -customerId $customerId -sharedKey $sharedKey -body ([System.Text.Encoding]::UTF8.GetBytes($json)) -logType $LogType  
}
else{

    Write-Output "No new Okta logs are avaliable as of $startDate"

```

chinguyen1:
@Yaniv-Shasha Can you please try this? 

if ($recordCount -gt 0) {
    Write-Output "$recordCount record(s) are avaliable as of $startDate"
    **$domain = [regex]::matches($uri, 'https:\/\/([\w\.]+)\/').captures.groups[1].value
    $response | Add-Member -MemberType NoteProperty -Name "Domain" -Value $domain**
    #$domain1 = $env:uri
    # $domain = $domain1.Replace("api/v1/logs?since=","")
    # $response  | Add-Member -MemberType NoteProperty -Name "Domain" -Value  $domain
    $json = $response | ConvertTo-Json -Depth 5
    Post-LogAnalyticsData -customerId $customerId -sharedKey $sharedKey -body ([System.Text.Encoding]::UTF8.GetBytes($json)) 
          -logType $LogType  
}
else{

Yaniv-Shasha:
Hi @chinguyen1  from my testing your sugested solution doesnt give me the domain feild.
let take it offline

chinguyen1:
Yaniv and I synced offline and it worked with these lines.

```
$domain = [regex]::matches($uri, 'https:\/\/([\w\.\-]+)\/').captures.groups[1].value
$responseObj = $response | ConvertFrom-Json
$responseObj | Add-Member -MemberType NoteProperty -Name "Domain" -Value $domain
```

@chicduong, would you mind adding this to your next PR?

chinguyen1:
[...]

PR #953, opened 2020-08-09
Version 2.01

cabberley:
## 2.0.1


cabberley:
Fixes issues #925 and #957

PR #959, opened 2020-08-12
Fix | Query

razmarom:
Fixes #

## Proposed Changes

  -
  -
  -


YaronFruchtmann:
Json dot notation doesn't work well with a dash symbol in key name. The fix replaced the dot notation with brackets

PR #966, opened 2020-08-13
Insecure Protocols Workbook update for CVE-2020-1472

briandelmsft:
Added Vulnerable Netlogon Secure Channel (for CVE-2020-1472)
Added Legacy Authentication to Summary
Fixed reporting of Weak Kerberos Cipher in summary
Updated Insecure Protocols workbook version

### DIFF
--- a/Workbooks/InsecureProtocols.json
+++ b/Workbooks/InsecureProtocols.json
@@ -49,7 +49,8 @@
             },
             "timeContextFromParameter": "TimeRange",
             "queryType": 1,
-            "resourceType": "microsoft.resourcegraph/resources"
+            "resourceType": "microsoft.resourcegraph/resources",
+            "value": null
           },
           {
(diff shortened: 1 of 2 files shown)


sudhamani85:
Is this condition correct for **insecure protocol summary**? As per windows documentation https://docs.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4769 thye ask us to monitor other than 0x11 and 0x12 Ticket Encryption Type
`(EventID == 4768 or EventID == 4769) and Level == 8 and (TicketEncryptionType != "0x12" and TicketEncryptionType != "0x11")`

briandelmsft:
Yes, that filter logic covers that scenario.  It looks for TGT (4768) or TGS (4769) requests, that are successful (Level == 8) and unsuccessful requests do not have an encryption type since no ticket is issued, and only if the encryption type is not 0x12 or 0x11 ( != is not equals).  If you are seeing something different please provide additional detail on the issue.

PR #969, opened 2020-08-14
Added pagination support and flatten the data.

pbankar28:
# Fixes

- Pagination support is added which was missing earlier.
- Data is flattened such that there will be one record per detection per host.

## Proposed Changes

  - URI parameter is simplified to accept only API Server URL, the endpoint route is moved to PS script. User need not provide the actual endpoint. 
  - Added some more API error handling. (Concurrency limit, reached, authentication handling, session creation error handling)
  - Updated the Object names used in parsing logic for better understanding.


pmgupte:
@preetikr The PR has been updated after you requested changes. I wanted to know if there is any other change required in this regard. If no other change is required, can you confirm if this PR is good to merge, so that we can unblock the users awaiting this change?

preetikr (changes requested):
Thanks, for incorporating feedback.
1. Is it possible to please drop the changes from the old / existing files and just have the v2 changes included in this PR? 
2. Also .zip of the v2 file is needed too.
3. host, requirements, profile files I believe can be as-is. Please let me know if you think otherwise.
4. This would leave only the change to point to this v2 azure deploy file in the QualysVM_API_FunctionApp.json file and updating all locations to point to this v2 run.ps1 file etc. Would be great if you can include this change as well.

We are also getting validations on this to ensure the v2 works E2E. Thanks.

pranaybqualys:
@preetikr , 

1. May I know whether to update the file `QualysVM_API_FunctionApp.json` to `QualysVM_API_FunctionApp_v2.json`?
2. Should we create the zip from `AzureFunctionQualysVMv2.0` directory?

pmgupte:
@preetikr tagging to bring attention to this PR. Is there any other change required?

preetikr:
@v-jayakal - please merge the changes so that this can be done as Qualys v2 data connector as there's a PR in flight for that one. Thanks.

v-jayakal:
Closing this PR as this has conflicts and a new PR has been raised: https://github.com/Azure/Azure-Sentinel/pull/1839

PR #976, opened 2020-08-15
Malicious Link Clicked 

secops-and-hops:
Fixes #
Forked malware rule and included the phish classification
## Proposed Changes

  -  where classification_s =~ "phish" or classification_s =~ "malware"


aprakash13:
@secops-and-hops When you have some time would you please make the suggested changes as per my previous comment and submit a PR. Thanks again for all your work here.

shainw:
> Thanks Anthony for submitting this .
> 
> It feels like it might be a good idea to have a separate query for phishing instead of merging them in one. Wondering, if you could please create separate query for phish. Couple of other minor suggestions:
> 
> description : In the description section we should probably start with “This query identifies”. Please see query style guide here: https://github.com/Azure/Azure-Sentinel/wiki/Query-Style-Guide
> 
> severity: We normally have Low, Medium and High severity detections in Azure Sentinel . Might be helpful to avoid ‘Informational’
> This is how we think about the severity level of the detections:
> • Low – Could potentially cause noise or is a detection that would need additional detections to raise the overall severity.
> • Medium – These are generally quiet but may require some additional investigation to verify the impact of the attack.
> • High – These are high confidence, rare detections that are generally guaranteed to indicate compromise or a high level impact of an attack.
> 
> Thanks again.

@secops-and-hops - Please have a look at the feedback and make changes when you have a chance.  We are happy to get this included in the repo.  Thanks, Shain

PR #987, opened 2020-08-20
Fix | EntityMappings for github detections

razmarom:
Fixes #

## Proposed Changes

  -
  -
  -


Detections/SigninLogs/GitHub Activities from Infrequent Country.yaml
```diff
@@ -3,6 +3,10 @@ name: GitHub Activites from a New Country
 description: |
   'Detect activities from a location that was not recently or was never visited by the user or by any user in your organization.'
 severity: Medium
+requiredDataConnectors:
```
shainw, line 6: This should not be included, Detections do not support CustomConnectors, it must be from the list or left out if it does not exist - [Query Style guide - Data connectors](https://github.com/Azure/Azure-Sentinel/wiki/Query-Style-Guide#requireddataconnectors)

shainw (changes requested):
This file should not be changed to SigninLogs, that is the that datatype it uses, please move it back - Hunting Queries/SigninLogs/GitHub Activities from Infrequent Country.yaml

shainw (changes requested):
All good, except this one should be moved back to the GitHub folder as it doesn't use SigninLogs - Detections/SigninLogs/GitHub Activities from Infrequent Country.yaml

shainw (approved):

PR #990, opened 2020-08-20
Update ForgeRock queries, CyberArk EPV, CEF template

matthew-ouellette:


DataConnectors/ForgeRock_CEF.json
```diff
@@ -7,30 +7,30 @@
     {
       "metricName" : "Total data received",
       "legend" : "ForgeRock",
-      "baseQuery" : "CommonSecurityLog | where DeviceVendor in (\"ForgeRock Inc\") | where DeviceProduct in (\"IDM\") // can be ANY ForgeRock product (AM, IG, Etc) "
+      "baseQuery" : "CommonSecurityLog\n| where DeviceVendor == \"ForgeRock Inc\" and DeviceProduct == \"IDM\"\n"
     }
   ],
   "sampleQueries" : [
     {
       "description" : "Total failed login attempts to the ForgeRock Platform",
```
preetikr, line 15: The device product values can be different, hence we should leave that comment in the sample queries, so that customers can accordingly provide values and filter records. One value is IDM. Others can be AM, IG. Please retain this comment in other queries as well. It helps to provide that clarity to customers to minimize support issues in the future.

PR #997, opened 2020-08-21
Okta sso function v2

cabberley:
## 2.1.1
- Moved version 2.1.0 to sub folder AzureFunctionOktaSSO_V2 under Azure-Sentinel/DataConnectors/Okta Single Sign-On/
- Created V2 versions of Azure Deploy and Connector Json files to enable both versions to co-exist
- Changed zip file reference to use 'https://aka.ms/sentineloktaazuredeployv2' in Azure Deploy json file
- Changed run.ps1 reference to use 'https://aka.ms/sentineloktaazurefunctioncodev2' in Connector json file

## 2.1.0
- Modified Okta Event log tracking to use next URI link from header
- Fixed Total record counter to record correctly the total number of records ingested

## 2.0.2
- Added  ACN_CD_OktaIssue925 resolution to add domain to output records

## 2.0.1


cabberley:
> Thanks, for creating the Azure Functions as a separate version for this change. The data connector json file Connector_REST_API_FunctionApp_Okta_V2.json need not be created as a separate version. It's a UX file and we can just update the existing data connector UX file to "deploy to azure" and functions link (manual) to the new v2 version once validations are complete.

The content of this json file V2 in the Merge Request has changes to the v1 to reflect the removal of the "Time" parameter so how do we manage that?

cabberley:
@preetikr  I have removed the Json file as requested.

PR #1009, opened 2020-08-25
Add TAXII data source to analytics and hunting queries

pemontto:
## Proposed Changes

- Currently analytic templates will look as if they don't have the required source ThreatIntelligenceIndicator data if nothing is coming specifically from the Graph Security API. However these analytics are still valid and work if TI data is coming from the TAXII connector.

Screen Shot 2020-08-25 at 10 20 31

- Minor change to the "(Preview) TI map Domain entity to DnsEvent" to bring it inline with the other Domain entity analytics so the filter doesn't rely on an arbitrary ID but rather the presence of a domain entity.


shainw (approved):
Thanks for contributing and fixing the bug in the Domain+DNS events detection.

PR #1044, opened 2020-09-08
Fix computer field unmapped bug

NoamLandress:


NoamLandress (commented):
In this PR I inform clients that their syslog filter mapping is fault, and will not map the Computer field from their hostname.
I suggest a fix as well for them to run.

morshabi (approved):

PR #1051, opened 2020-09-09
Updated the BYO ML Notebooks and Utilities for Public Preview

fafadavi:
Fixes #

## Proposed Changes

  -
  -
  -


ianhelle:
@fafadavi - couple of questions:
- instead of uploading Python wheels have you considered publishing these packages on PyPI? Seems like a more accessible place for anyone that wants to use them
- We have a separate sub-repo for notebooks - github/com/Azure/Azure-Sentinel-Notebooks - would this be a better place for the notebooks? We have a mechanism to show notebooks here in the Sentinel UI and clone them to an AzML notebook workspace (although I realize that these are Databricks libs/notebooks so the latter part may not be 100% appropriate).

fafadavi:
Hi Ian,

I had a chance to talk with Anil, who wrote these notebooks and he felt that it wasn’t the right time to put the Python Wheels on PyPi.

As well, since BYOML does not yet support Azure ML and our private preview customers are use to seeing the Notebooks in Azure/Azure-Sentinel I think it is best to keep it in the Azure-Sentinel repo. In GA when we do support PyPi and (potentially) Azure ML, I think Azure/Azure-Sentinel-Notebooks will be the right place.

I have created a new PR with an updated version of the Notebooks and wheel file. Can you please ignore this PR and review the new one?

Thanks,
Faraz

PR #1083, opened 2020-09-15
fix fault notice for RH firewall check

NoamLandress:
Will check if an exception for the OMS agent exists, and if not- suggests a command for users to run.

### DIFF
--- a/DataConnectors/CEF/cef_troubleshoot.py
+++ b/DataConnectors/CEF/cef_troubleshoot.py
@@ -128,10 +128,15 @@ def check_red_hat_firewall_issue():
 
 
 def validate_rh_firewall_exception():
+    '''
+    Validating that a firewall rule with the agents port as an exception exists
+    :return: True if exception was found, False otherwise
+    '''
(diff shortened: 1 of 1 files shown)


DataConnectors/CEF/cef_troubleshoot.py
```diff
@@ -110,38 +110,41 @@ def check_red_hat_firewall_issue():
...
-    '''
-    print("Checking for exception for omsagent")
-    print_notice(firewall_d_exception_configuration_file)
-    firewall_status = subprocess.Popen(["sudo", "cat", firewall_d_exception_configuration_file], stdout=subprocess.PIPE)
-    o, e = firewall_status.communicate()
-    if e is not None:
-        print_error("Error: could not get /etc/firewalld/zones/public.xml file holding firewall exceptions")
-    print_command_response(str(o))
-    return agent_port in str(o)
+def validate_rh_firewall_exception():
```
ofshomro, line 130: Can you add some docs about what the function is doing and how?

NoamLandress: Done.

morshabi (approved):

PR #1084, opened 2020-09-15
Update Zscaler-LowVolumeDomainRequests.yaml

thmcelro:
Fixes 

- Expanded supported script extensions and made them easier to customise.
- Added additional check for POST request or GET request with parameters to remove FP's where no data was submitted to the server. Use the max summarisation function as we only care about there being one set of activity where data is transferred.
- Added additional check to ensure that no referrer was provided, this check defeats most FP's caused by wordpress's admin ajax
- Added basic defeat to remove email click tracking activity, while this may increase FN's it should significantly reduce FP's.

### DIFF
--- a/Detections/CommonSecurityLog/Zscaler-LowVolumeDomainRequests.yaml
+++ b/Detections/CommonSecurityLog/Zscaler-LowVolumeDomainRequests.yaml
@@ -30,7 +30,7 @@ query: |
   | where SentBytes > 0 and ReceivedBytes > 0
   // Extract the Domain
   | extend Domain = iff(countof(DestinationHostName,'.') >= 2, strcat(split(DestinationHostName,'.')[-2], '.',split(DestinationHostName,'.')[-1]), DestinationHostName)
-  | extend GetData=iff(RequestURL in "?", 1, 0)
+  | extend GetData=iff(RequestURL == "?", 1, 0)
   | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), makelist(RequestURL), makelist(DestinationIP), makelist(SourceIP), numOfConnections = count(), make_set(RequestMethod), max(GetData), max(RequestContext) by Domain
   // Determine the number of URIs that have been visited for the domain
   | extend destinationURI = arraylength(list_RequestURL)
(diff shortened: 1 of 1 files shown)


Detections/CommonSecurityLog/Zscaler-LowVolumeDomainRequests.yaml
```diff
@@ -19,25 +19,31 @@ relevantTechniques:
...
-  // Only look at connections that were allowed through the web proxy. Comment line out for all events including those which were blocked
+  // Only look at connections that were allowed through the web proxy
   | where DeviceVendor =~ "Zscaler" and DeviceAction =~ "Allowed"
-  // Only look where some data was exchanged. Comment out to see 0 byte connections
+  // Only look where some data was exchanged.
   | where SentBytes > 0 and ReceivedBytes > 0
   // Extract the Domain
   | extend Domain = iff(countof(DestinationHostName,'.') >= 2, strcat(split(DestinationHostName,'.')[-2], '.',split(DestinationHostName,'.')[-1]), DestinationHostName)
-  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), makelist(RequestURL), makelist(DestinationIP), makelist(SourceIP), numOfConnections = count() by Domain
+  | extend GetData=iff(RequestURL in "?", 1, 0)
```
shainw, line 33: If you are using "in", then this must be in parenthesis - iff(RequestURL in **("?")**, 1, 0).  However, you are only matching on "?" and so you should use "=" instead of "in".

Detections/CommonSecurityLog/Zscaler-LowVolumeDomainRequests.yaml
```diff
@@ -19,25 +19,31 @@ relevantTechniques:
...
   | extend destinationURI = arraylength(list_RequestURL)
   | where destinationURI <= uriThreshold
-  | where tostring(list_RequestURL) has_any (".php",".aspx")
-  ;
-  suspiciousURLs
+  | where tostring(list_RequestURL) has_any(scriptExtensions)
+  //Remove matches with referer
+  | where max_RequestContext == ""
+  //Keep requests where data was trasferred either in a GET with parameters or a POST
+  | where set_RequestMethod in~ "POST" or max_GetData == 1
```
shainw, line 42: Same here - this needs parenthesis - set_RequestMethod in~ **("POST")**.  Again, you probably should be using "has" here as this is a set.

shainw (changes requested):

shainw (approved):

PR #1090, opened 2020-09-17
Bug fix failed logon attemptswithin10m

aprakash13:
Fixes #

## Proposed Changes

  -
  -
  -


aprakash13:
Modifying this detection query so as to restrict it to looking for Failed logon attempts by valid accounts only.  Filtering for this was done using SubStatus == '0xc0000064'  which stands for  'Account name does not exist'.

Will modify the below detection query so that it also looks for failed logon attempts from unknown users on Windows machine . Currently the below query is only targeted to Linux devices.
 https://github.com/Azure/Azure-Sentinel/blob/bd4b2c5947b6248a5b77a55663bce85f6f0ff7a5/Detections/Syslog/FailedLogonAttempts_UnknownUser.yaml

shainw (approved):

PR #1095, opened 2020-09-21
Parsers sept-2020

ashwin-patil:
Fixes #

## Proposed Changes

  -  Parser for Cisco ISE (Identity Services Engine) Syslog
  - Sysmon KQL Parser for all versions irrespective of schema (Less performant than other versions)
  - Sysmon v 12.0 Parser - autogenerated from OSSEM tool - ossemSysmonKQLParser.py

### DIFF
--- a/Parsers/Cisco_ISEParser.txt
+++ b/Parsers/Cisco_ISEParser.txt
@@ -1,0 +1,43 @@
+// KQL Parser for Cisco ISE (Identity Services Engine) Syslog
+// Last Updated Date: Sept 20, 2020
+//
+// Parser Notes: 
+// 1. Extracts common fields based on Regex
+//      TimeStamp, hostname, ise_type, sessionkey, Segments,Segnumber
+// 2. Common Json Property bag to hold all the remaining key value pairs. 
+//     - (Easily extractable using mv-expand and parse_json on demand)
+//     - Some of the keys are duplicate in logs hence we can`t use directly make_bag, using make_list. 
+// 3. Regular expression in this parser assumes field names in alphanumeric format (a-zA-Z0-9- ) and values in alphanumeric 
+//      and some special characters format (a-zA-Z0-9-_:/@.#{};= ) which covers most cases per initial testing. 
+//      To include any additional special chars modify regex accordingly.
+// 4. Parser expects Syslog to be of below format in the same order and in a field - RawData. 
+//    For any changes, adjust the regex accordingly or submit an issue of Github with the format you are trying.
+//    Expected Format - (SYSLOGTIMESTAMP:timestamp} {HOSTNAME:hostname} {DATA:cisco_ise_type} {INT:session_key} {INT:total_segments} {INT:seg_number} %{GREEDYDATA:message- key-value pairs}
+// 5. Reference - Documentation Cisco ISE PSN Logs Remote Format: 
+//  https://www.cisco.com/c/en/us/td/docs/security/ise/2-6/Cisco_ISE_Syslogs/Cisco_ISE_Syslogs/Cisco_ISE_Syslogs_chapter_00.html#id_101580
+// 
+//
+// Usage Instruction : 
+// Paste below query in log analytics, click on Save button and select as Function from drop down by specifying function name and alias (e.g. CommonSecurityLogs_Parsed).
+// Function usually takes 10-15 minutes to activate. You can then use function alias from any other queries (e.g. CommonSecurityLogs_Parsed | take 10).
+// Reference :
+// Using functions in Azure monitor log queries : https://docs.microsoft.com/azure/azure-monitor/log-query/functions
+// Tech Community Blog on KQL Functions : https://techcommunity.microsoft.com/t5/Azure-Sentinel/Using-KQL-functions-to-speed-up-analysis-in-Azure-Sentinel/ba-p/712381
+//
+Cisco_ISE_PSN_CL
+| extend  TimeStamp = extract( "(\\w+\\s\\d{2}\\s\\d{2}:\\d{2}:\\d{2}) (.*)",1,RawData)
+| extend  HostName = extract( "(\\w+\\s\\d{2}\\s\\d{2}:\\d{2}:\\d{2}) (\\S+) (.*)",2,RawData)
+| extend  ISE_Type = extract( "(\\w+\\s\\d{2}\\s\\d{2}:\\d{2}:\\d{2}) (\\S+) (\\S+) (.*)",3,RawData)
+| extend  SessionKey = extract( "(\\w+\\s\\d{2}\\s\\d{2}:\\d{2}:\\d{2}) (\\S+) (\\S+) (\\d{10}) (.*)",4,RawData)
+| extend  TotalSegments = extract( "(\\w+\\s\\d{2}\\s\\d{2}:\\d{2}:\\d{2}) (\\S+) (\\S+) (\\d{10}) (\\d{1,2}) (.*)",5,RawData)
+| extend  SegmentNumber = extract( "(\\w+\\s\\d{2}\\s\\d{2}:\\d{2}:\\d{2}) (\\S+) (\\S+) (\\d{10}) (\\d{1,2}) (\\d{1,2}) (.*)",6,RawData)
+| extend  Remainder = extract( "(\\w+\\s\\d{2}\\s\\d{2}:\\d{2}:\\d{2}) (\\S+) (\\S+) (\\d{10}) (\\d{1,2}) (\\d{1,2}) (.*)",7,RawData)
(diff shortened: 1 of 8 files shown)


Parsers/Cisco_ISEParser.txt
```diff
@@ -0,0 +1,43 @@
...
+//     - (Easily extractable using mv-expand and parse_json on demand)
+//     - Some of the keys are duplicate in logs hence we can`t use directly make_bag, using make_list. 
+// 3. Regular expression in this parser assumes field names in alphanumeric format (a-zA-Z0-9- ) and values in alphanumeric 
+//      and some special characters format (a-zA-Z0-9-_:/@.#{};= ) which covers most cases per initial testing. 
+//      To include any additional special chars modify regex accordingly.
+// 4. Parser expects Syslog to be of below format in the same order and in a field - RawData. 
+//    For any changes, adjust the regex accordingly or submit an issue of Github with the format you are trying.
+//    Expected Format - (SYSLOGTIMESTAMP:timestamp} {HOSTNAME:hostname} {DATA:cisco_ise_type} {INT:session_key} {INT:total_segments} {INT:seg_number} %{GREEDYDATA:message- key-value pairs}
+// 5. Reference - Documentation Cisco ISE PSN Logs Remote Format: 
+//  https://www.cisco.com/c/en/us/td/docs/security/ise/2-6/Cisco_ISE_Syslogs/Cisco_ISE_Syslogs/Cisco_ISE_Syslogs_chapter_00.html#id_101580
```
shainw, line 17: Remove the locale specific info in the URL.

Parsers/Sysmon/Sysmon-AllVersions_Parser.txt
```diff
@@ -0,0 +1,41 @@
...
+| where Source == "Microsoft-Windows-Sysmon"
+| extend RenderedDescription = tostring(split(RenderedDescription, ":")[0])
+| extend EventData = parse_xml(EventData).DataItem.EventData.Data
+| mv-expand bagexpansion=array EventData
+| evaluate bag_unpack(EventData)
+| extend Key=tostring(['@Name']), Value=['#text']
+| evaluate pivot(Key, any(Value), TimeGenerated, Source, EventLog, Computer, EventLevel, EventLevelName, EventID, UserName, RenderedDescription, MG, ManagementGroupName, Type, _ResourceId)
+| parse Hashes with * 'SHA1=' SHA1 ',' * 'MD5=' MD5 ',' * 'SHA256=' SHA256 ',' * 'IMPHASH=' IMPHASH
+| parse RuleName with * 'technique_id=' TechniqueId ',' * 'technique_name=' TechniqueName
+| project TimeGenerated, Source, EventLog, Computer, EventLevel, EventLevelName, EventID, UserName, RenderedDescription, MG, ManagementGroupName, Type, _ResourceId,
```
shainw, line 35: Would it make more sense to do a project-away here?  Maybe a shorter list of items?

Parsers/Sysmon/Sysmon-AllVersions_Parser.txt
```diff
@@ -0,0 +1,41 @@
...
+//
+Event
+| where Source == "Microsoft-Windows-Sysmon"
+| extend RenderedDescription = tostring(split(RenderedDescription, ":")[0])
+| extend EventData = parse_xml(EventData).DataItem.EventData.Data
+| mv-expand bagexpansion=array EventData
+| evaluate bag_unpack(EventData)
+| extend Key=tostring(['@Name']), Value=['#text']
+| evaluate pivot(Key, any(Value), TimeGenerated, Source, EventLog, Computer, EventLevel, EventLevelName, EventID, UserName, RenderedDescription, MG, ManagementGroupName, Type, _ResourceId)
+| parse Hashes with * 'SHA1=' SHA1 ',' * 'MD5=' MD5 ',' * 'SHA256=' SHA256 ',' * 'IMPHASH=' IMPHASH
```
shainw, line 33: Just want to confirm this parse on Hashes does not fail with an error if Hashes field is not there?  if so, we should maybe think about using column_ifexists.

Parsers/Sysmon/Sysmon-AllVersions_Parser.txt
```diff
@@ -0,0 +1,41 @@
...
+| parse RuleName with * 'technique_id=' TechniqueId ',' * 'technique_name=' TechniqueName
```
shainw, line 34: Just want to confirm this parse on RuleName does not fail with an error if RuleName field is not there?  if so, we should maybe think about using column_ifexists.

Parsers/Sysmon/Sysmon-v12.0.txt
```diff
@@ -0,0 +1,310 @@
...
+    | project-away EventDetail  ;
+    processEvents;
+    
+};
+(union isfuzzy=true  SYSMON_ERROR_255, SYSMON_CREATE_PROCESS_1, SYSMON_FILE_TIME_2, SYSMON_NETWORK_CONNECT_3, SYSMON_SERVICE_STATE_CHANGE_4, SYSMON_PROCESS_TERMINATE_5, 
+SYSMON_DRIVER_LOAD_6, SYSMON_IMAGE_LOAD_7, SYSMON_CREATE_REMOTE_THREAD_8, SYSMON_RAWACCESS_READ_9, SYSMON_ACCESS_PROCESS_10, SYSMON_FILE_CREATE_11, SYSMON_REG_KEY_12, 
+SYSMON_REG_SETVALUE_13, SYSMON_REG_NAME_14, SYSMON_FILE_CREATE_STREAM_HASH_15, SYSMON_SERVICE_CONFIGURATION_CHANGE_16, SYSMON_CREATE_NAMEDPIPE_17, SYSMON_CONNECT_NAMEDPIPE_18, 
+SYSMON_WMI_FILTER_19, SYSMON_WMI_CONSUMER_20, SYSMON_WMI_BINDING_21, SYSMON_DNS_QUERY_22, SYSMON_FILE_DELETE_23, SYSMON_CLIPBOARD_24)
+| extend Details = column_ifexists("Details", ""), RuleName = column_ifexists("RuleName", ""), PreviousCreationUtcTime=column_ifexists("PreviousCreationUtcTime", "")
+| project EventID, UtcTime, ID, Description, RuleName, ProcessGuid, ProcessId, Image, FileVersion, Product, Company, OriginalFileName, CommandLine, CurrentDirectory,
```
shainw, line 304: Same comment here, does it make more sense to use project-way?

shainw (changes requested):

Parsers/Cisco_ISEParser.txt
ashwin-patil: It seems Cisco does not support global locale like docs.microsoft.com . I tried removing en/us but throws 404.

Have added instruction in the notes, if user wants to change.. Also build is passing so i think we can keep this URL.

Parsers/Sysmon/Sysmon-AllVersions_Parser.txt
ashwin-patil: added `column_ifexists` check for the field.

Parsers/Sysmon/Sysmon-AllVersions_Parser.txt
ashwin-patil: Explicit project of all columns required for function to auto-populate the field names with intellisense. 
I tried with project-away and does not provide full listing of derived fieldnames.

Parsers/Sysmon/Sysmon-v12.0.txt
ashwin-patil: Explicit project of all columns required for function to auto-populate the field names with intellisense. 
I tried with project-away and does not provide full listing of derived fieldnames.

ashwin-patil (commented):
Made some changes.

Added some new additions:
- Dynamic handling of Hashes to create Json object when customer has enabled one or all 4 types of hashes.
- Can not use same hashes logic in all version as extract_all fails when field is empty. It works in other parser as we have per event Id parsing.

Parsers/Sysmon/Sysmon-AllVersions_Parser.txt
ashwin-patil: Added `column_ifexists` check for Hashes

Parsers/Cisco_ISEParser.txt
shainw: Yeah, the check for locale is not done on parsers, but good practice to have no locale any where that we can.  Understood that Cisco does not have it available.

Parsers/Sysmon/Sysmon-AllVersions_Parser.txt
shainw: Got it, thanks for checking.

Parsers/Sysmon/Sysmon-AllVersions_Parser.txt
shainw: Thanks

shainw (approved):

PR #1137, opened 2020-09-30
Update PulseConnectSecure.txt

FlyingBlueMonkey:
Removed extra "," from end of query.  Separated the two versions of the parser namely that the first parser section works with the default PulseSecure format and the second is expecting WebTrends Enhanced Log Format (WELF)

Fixes #
Removed stray comment
## Proposed Changes
Clarify format of logs and parser sections
  -
  -
  -

### DIFF
--- a/Parsers/PulseConnectSecure/PulseConnectSecure.txt
+++ b/Parsers/PulseConnectSecure/PulseConnectSecure.txt
@@ -29,7 +29,7 @@
 //
 Syslog
 | where Computer in ("datasource") and Facility == "local7"
-//Version 8.0R7 and below
+//Version 8.0R7 and below using the Standard format
 | extend Parser = extract_all(@'^(\d{4}\-\d{2}-\d{2})\s(\d{2}\:\d{2}:\d{2})\s(\S+)\s(\S+)\s(\S+)\s\[(\S+)\]\s(\S+)\((.*)?\)\[(.*)\]\s\-\s(.*)',dynamic([1,2,3,4,5,6,7,8,9,10]),SyslogMessage)
 | mv-expand Parser
 | extend LogTime = todatetime(strcat(tostring(Parser[0]),'T',tostring(Parser[1]))), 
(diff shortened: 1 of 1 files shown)


Parsers/PulseConnectSecure/PulseConnectSecure.txt
```diff
@@ -40,7 +40,9 @@ Syslog
     EventID = tostring(Parser[8]), 
     Messages = tostring(Parser[9])
 | project-away Parser
-//Version 8.0R7 and above
+
```
shainw, line 43: This extra line has to be removed, otherwise the parser will fail to run the 'extend' below.

shainw (changes requested):

Parsers/PulseConnectSecure/PulseConnectSecure.txt
FlyingBlueMonkey: The line break is actually there for a reason.  If you run the combined query, it will rewrite the values from the first parser.  Unless I am mistaken, the parsers are / were originally intended to be for two different log formats.  The first one (that uses "Parser") is for the PulseConnect Standard log format.  The second one (labeled as "8.097 and above") would only work with the WebAnalytics Extended Log Format (WELF) format.  If you're using Standard log format, the second section overwrites the first block of parsed content.

shainw: You could have a look at [this parser](https://github.com/Azure/Azure-Sentinel/blob/master/Parsers/Sysmon/Sysmon-AllVersions_Parser.txt) which uses syntax related to column_ifexists to allow for differing columns, otherwise this will have to be broken into 2 parsers, the line break will cause the parser to fail to load.

shainw (changes requested):

Parsers/PulseConnectSecure/PulseConnectSecure.txt
shainw: The original author, @chicduong may be able to comment further, but I would suggest for now that you remove the extra line as this will cause failure on import as a function and only remove the trailing comma.  @chicduong - can you validate the parser works, I don't have a current setup of PulseConnectSecure data.

FlyingBlueMonkey: What if we simply comment out the second block with a note about the version differences?  Leaving the second block active will cause the parser to return empty / null values because it re-uses the same column names and if the values don't exist then it fails.  Another option could be to use an IIF around the values, no?

chicduong: This is correct the "comma" will need to be removed. I agree, commenting out the second block would be the easiest approach and allow the customers to un-comment the block if they're using that format.

shainw: Sounds good to me.  Once you make the changes, I will approve and complete. ;)

shainw:
Closing this as I published a new PR against latest master and based on the discussion - https://github.com/Azure/Azure-Sentinel/pull/1294

PR #1140, opened 2020-10-01
Okta sso function v2

cabberley:
## 2.1.2
- fixed a bug with the extraction and insertion of the domain name into the record set to ingest into Log Analytics
- Changed the value used for the RowKey to a config name instrad of the Okta API, Okta API can be changed on demand by admins

## 2.1.0
- Modified Okta Event log tracking to use next URI link from header
- Fixed Total record counter to record correctly the total number of records ingested

## 2.0.2
- Added  ACN_CD_OktaIssue925 resolution to add domain to output records

## 2.0.1


cabberley:
@preetikr Created new pull Request with the Domain issue fixed

preetikr (approved):

PR #1155, opened 2020-10-08
Update To Connector_Json

cabberley:
Updates to Connector_REST_API_FunctionApp_Okta.json to reflect change in Okta V2 Connector.

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/DataConnectors/Okta Single Sign-On/Connector_REST_API_FunctionApp_Okta.json
+++ b/DataConnectors/Okta Single Sign-On/Connector_REST_API_FunctionApp_Okta.json
@@ -50,7 +50,16 @@
                     "read": true,
                     "delete": true
                 }
-            }
+            },
+            {	
+                "provider": "Microsoft.OperationalInsights/workspaces/sharedKeys",	
+                "permissionsDisplayText": "read permissions to shared keys for the workspace are required. [See the documentation to learn more about workspace keys](https://docs.microsoft.com/azure/azure-monitor/platform/agent-windows#obtain-workspace-id-and-key)",	
+                "providerDisplayName": "Keys",	
+                "scope": "Workspace",	
+                "requiredPermissions": {	
+                    "action": true
+                }
+             }
         ],
         "customs": [
             {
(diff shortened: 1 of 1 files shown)


cabberley:
PR for update to Connector JSON file

DataConnectors/Okta Single Sign-On/Connector_REST_API_FunctionApp_Okta.json
```diff
@@ -1,6 +1,6 @@
 {
-    "id": "OktaSSO",
-    "title": "Okta Single Sign-On (Preview)",
+    "id": "OktaSSOV2",
```
preetikr, line 2: Thanks, @cabberley - Let's keep the id same as before and also title too. We are swapping out the earlier version and replacing with v2 Azure Function. Changing id and title makes this a new data connector and just will cause confusion and people on two different connector paths. 

On a slightly related note, I just saw a couple of CSS / customer requests on the Okta data connector (v1) which is like this. Want to find out is this scenario is handled in v2 so that we can ensure works for this case too. 
"I configured the Data Connector for Okta single sign-on, connected to Okta, and the data display worked. However, if you configure another Okta tenant in the same way, it will be deployed with an override and you will not be able to log the first tenant you configured. Is there a way to add multiple Okta tenants with the same connector? 
If you manually configure Azure Function, you can get logs from three tenants, but you don't know which logs are from which tenant. We have three Okta tenants and we want to collect three logs in Sentinel in an identifyable way, but what should we do?

preetikr (changes requested):

cabberley:
Hi @preetikr updated as requested. 
On the other query, the V2 connector with its inclusion of the "domain" field in the results before they are ingested, will identify which records are from which OKTA instance, if they follow these instructions https://github.com/cabberley/AS-Okta-SSO/wiki/Deploying-the-Okta-V2-Connector-for-multiple-Okta-Accounts

DataConnectors/Okta Single Sign-On/Connector_REST_API_FunctionApp_Okta.json
```diff
@@ -50,15 +50,6 @@
                     "read": true,
                     "delete": true
                 }
-            },
```
preetikr, line 53: This section  is needed for building the data connector - please add this back.

preetikr (changes requested):

preetikr (approved):

PR #1156, opened 2020-10-08
Create UserLoginIPAddressTeleportation.yaml

thmcelro:
Analytic to detect users that have successfully logged in from two countries within a 20 minute window of the previous login. This analytic will detect VPN activity, a variable is included to allow known ranges to be defeated.


shainw (approved):
Suggested addition to the end of the query to aggregate a bit and reduce the row return count, but not required.

PR #1173, opened 2020-10-14
Palo Alto data connector fixes

erant10:
## Changes 
  - added PAN connector JSON (exported from portal)
  - replaces DeviceProduct string operator from `==` to `contains` for both workbooks and connector


morshabi:
@erant10 did you also updated the version of the workbook in the workbook metadata?

PR #1178, opened 2020-10-14
MFA push deny detection

secops-and-hops:
Fixes #

## Proposed Changes
Detection created for MFA push deny
  -
  -
  -

### DIFF
--- a/Detections/SigninLogs/ExplicitMFADeny.yaml
+++ b/Detections/SigninLogs/ExplicitMFADeny.yaml
@@ -1,0 +1,24 @@
+id: a22740ec-fc1e-4c91-8de6-c29c6450ad00
+name: Explicit MFA Deny
+description: |
+  'User explicitly denies MFA push, indicating that login was not expected and the account's password may be compromised.'
+severity: Medium
+requiredDataConnectors:
+  - connectorId: AzureActiveDirectory
+    dataTypes:
+      - SigninLogs
+queryFrequency: 1h
+queryPeriod: 1h
+triggerOperator: gt
+triggerThreshold: 0
+tactics:
+  - CredentialAccess
+relevantTechniques:
+  - T1110
+query: |
+  SigninLogs 
+  | where Status contains "MFA denied; user declined the authentication"
+  | project TimeGenerated,Identity,IPAddress,AppDisplayName,ClientAppUsed,AlternateSignInName,AuthenticationDetails,Status,DeviceDetail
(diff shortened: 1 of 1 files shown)


Detections/SigninLogs/ExplicitMFADeny.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - CredentialAccess
+relevantTechniques:
+  - T1110
+query: |
+  SigninLogs
+  | where ResultType == 500121
+  | where Status has "MFA Denied; user declined the authentication"
+  | extend AccountCustomEntity = AlternateSignInName
+  | extend IPCustomEntity = IPAddress
+  | extend URLCustomEntity = ClientAppUsedlet LearningPeriod = 7d;
```
shainw, line 24: You have a let statement trailing here that should be removed - "let LearningPeriod = 7d;"

Detections/SigninLogs/ExplicitMFADeny.yaml
```diff
@@ -0,0 +1,24 @@
+id: a22740ec-fc1e-4c91-8de6-c29c6450ad00
+name: Explicit MFA Deny
+description: |
+  'User explicitly denies MFA push, indicating that login was not expected and the account's password may be compromised.'
+severity: Medium
+requiredDataConnectors:
+  - connectorId: AzureActiveDirectory
+    dataTypes:
+      - SigninLogs
+queryFrequency: 1h
```
shainw, line 10: We prefer default to be 1d and allow for users to change as they see fit in the UX.  Not required, but suggested.

Detections/SigninLogs/ExplicitMFADeny.yaml
```diff
@@ -0,0 +1,24 @@
...
+name: Explicit MFA Deny
+description: |
+  'User explicitly denies MFA push, indicating that login was not expected and the account's password may be compromised.'
+severity: Medium
+requiredDataConnectors:
+  - connectorId: AzureActiveDirectory
+    dataTypes:
+      - SigninLogs
+queryFrequency: 1h
+queryPeriod: 1h
```
shainw, line 11: We prefer default to be 1d and allow for users to change as they see fit in the UX.  Not required, but suggested.

shainw (changes requested):
Fix the trailing "let" statement, the others are not required but recommended.  Thanks much for contributing!

shainw (approved):

PR #1180, opened 2020-10-15
Bugfix/fix portal auto sync

NoamLandress:
Alert from potential auto sync with the portal which will override configuration updates made by the user to avoid duplicated data (syslog and cef)


NoamLandress:
@morshabi

morshabi (approved):

PR #1181, opened 2020-10-15
Adding new version of ThreatIntelligence Workbook

DonFu99:
Fixes #

Added updated workbook

  -
  -
  -


Amitbergman (commented):
You need to also change the version of the workbook in the workbooksMetadata file, otherwise customers who already saved this workbook will not see new version you are publishing.

shainw:
Very old PR, closing

PR #1188, opened 2020-10-19
New detection rule

KennethMLdk:
Fixes #

## Proposed Changes
Creates incident when a source IP has more than 1 registered deny action in Azure Firewall
  -
  -
  -

### DIFF
--- a/Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
+++ b/Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
@@ -1,0 +1,34 @@
+id: 
+name: Several deny actions registered
+description: |
+  'This Creates an incident when a source has more than 1 registered deny action in Azure Firewall.'
+severity: Medium
+requiredDataConnectors:
+  - connectorId: AzureFirewall
+    dataTypes: 
+      - AzureDiagnostics
+queryFrequency: 1h
+queryPeriod: 1h
+triggerOperator: gt
+triggerThreshold: 1
+tactics:
+  - InitialAccess
+relevantTechniques:
+  - 
+query: |
+
+let timeframe = ago(1h);
+let threshold = 2;
+AzureDiagnostics
+    | where TimeGenerated >= timeframe
+    | where OperationName in ('AzureFirewallApplicationRuleLog','AzureFirewallNetworkRuleLog')
+    | extend msg_s_replaced0 = replace(@'\s\s',@' ',msg_s)
+    | extend msg_s_replaced1 = replace(@'\.\s',@' ',msg_s_replaced0)
+    | extend msg_a = split(msg_s_replaced1,' ')
+    | extend srcAddr_a = split(msg_a[3],':') , destAddr_a = split(msg_a[5],':')
+    | extend protocol = tostring(msg_a[0]), srcIp = tostring(srcAddr_a[0]), srcPort = tostring(srcAddr_a[1]), destIp = tostring(destAddr_a[0]), destPort = tostring(destAddr_a[1]), action = tostring(msg_a[7])
+    | where action == 'Deny'
+    | extend url = iff(destIp matches regex "\\d+\\.\\d+\\.\\d+\\.\\d+","",destIp)
+    | summarize StartTime = min(TimeGenerated), count() by srcIp, destIp, url
+    | where count_ >= ['threshold']
+    | extend URLCustomEntity = url, IPCustomEntity = srcIp
--- a/Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
+++ b/Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
@@ -22,7 +22,7 @@ relevantTechniques:
 query: |
 
 let timeframe = ago(1h);
-let threshold = 2;
+let threshold = 1;
 AzureDiagnostics
     | where TimeGenerated >= timeframe
     | where OperationName in ('AzureFirewallApplicationRuleLog','AzureFirewallNetworkRuleLog')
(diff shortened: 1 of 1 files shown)


Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
```diff
@@ -0,0 +1,34 @@
+id: 
+name: Several deny actions registered
+description: |
+  'This Creates an incident when a source has more than 1 registered deny action in Azure Firewall.'
```
shainw, line 4: I think it would be good to explain what type of security issue this is.  It would be unclear to users what the attack might be based on this.

Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
```diff
@@ -0,0 +1,34 @@
...
+severity: Medium
+requiredDataConnectors:
+  - connectorId: AzureFirewall
+    dataTypes: 
+      - AzureDiagnostics
+queryFrequency: 1h
```
shainw, line 10: Generally, we want defaults to be 1d on frequency and period.  You can leave it as 1h, but this may cause many results per day.

Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
```diff
@@ -0,0 +1,34 @@
...
+queryPeriod: 1h
+triggerOperator: gt
+triggerThreshold: 1
+tactics:
+  - InitialAccess
```
shainw, line 15: I think this is more like CommandAndControl.  This is a deny on an outbound connection, correct?  To me, this would indicate the system has executed a payload and attempting to reach back to a malicious website.  I am open to a discussion, just want to make sure we are hitting the mark on the Tactic.

Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
```diff
@@ -0,0 +1,34 @@
...
+relevantTechniques:
+  -
```
shainw, line 17: A technique is required, once there is consensus on the Tactic, this will allow selection on the Technique.  If you agree on CommandAndControl, then potentially use [T1071](https://attack.mitre.org/tactics/TA0011/)

Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
```diff
@@ -0,0 +1,34 @@
...
+    | where TimeGenerated >= timeframe
+    | where OperationName in ('AzureFirewallApplicationRuleLog','AzureFirewallNetworkRuleLog')
+    | extend msg_s_replaced0 = replace(@'\s\s',@' ',msg_s)
+    | extend msg_s_replaced1 = replace(@'\.\s',@' ',msg_s_replaced0)
+    | extend msg_a = split(msg_s_replaced1,' ')
+    | extend srcAddr_a = split(msg_a[3],':') , destAddr_a = split(msg_a[5],':')
+    | extend protocol = tostring(msg_a[0]), srcIp = tostring(srcAddr_a[0]), srcPort = tostring(srcAddr_a[1]), destIp = tostring(destAddr_a[0]), destPort = tostring(destAddr_a[1]), action = tostring(msg_a[7])
+    | where action == 'Deny'
+    | extend url = iff(destIp matches regex "\\d+\\.\\d+\\.\\d+\\.\\d+","",destIp)
+    | summarize StartTime = min(TimeGenerated), count() by srcIp, destIp, url
```
shainw, line 32: I would recommend you bring thru protocol and action into the result so it is clear when receiving the alert.
Example - | summarize StartTime = min(TimeGenerated), count() by srcIp, destIp, url, **action, protocol**

Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
```diff
@@ -0,0 +1,34 @@
...
+    | where count_ >= ['threshold']
+    | extend URLCustomEntity = url, IPCustomEntity = srcIp
```
shainw, line 34: timestamp value is required, please add timestamp = StartTime.
Example - | extend **timestamp = StartTime**, URLCustomEntity = url, IPCustomEntity = srcIp

shainw (changes requested):
Some requested changes and thanks much for contributing.  Happy to discuss any recommendations.

Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
KennethMLdk: I think it will take too long time to discover the attack pattern if this is only run with 1d frequency. 1h is the right frequency for this. It shouldn't create a lot of incidents as the firewall rules should be configured correct and this indicates an outgoing attack.

Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
KennethMLdk: Agree. Edited.

Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
KennethMLdk: I agree that the tactics isn't really InitialAccess. Edited to CommandAndControl (for outbound connections), Discovery (for discovering resources) and LateralMovement (for moving between resources)

Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
KennethMLdk: Relevant techniques added.

Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
KennethMLdk: Good point. Added.

Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
KennethMLdk: Ok, added.

KennethMLdk:
Learned something on YAML formatting, apologies for the many commits.

Detections/AzureFirewall/SeveralDenyActionsRegistered.yaml
```diff
@@ -0,0 +1,38 @@
+id: 
+name: Several deny actions registered
+description: |
+  'Identifies attach pattern when attacker tries to move, or scan, from resource to resource on the network and creates an incident when a source has more than 1 registered deny action in Azure Firewall.'
```
shainw, line 4: Minor typo - "attach" should be "attack" :)

shainw (approved):
Minor typo and then good to go.

shainw:
> Learned something on YAML formatting, apologies for the many commits.

All good, thanks for the fixes.  Minor typo fix and all good for me to complete.  Thanks again for contributing!  Shain

PR #1196, opened 2020-10-20
Triton Fix

lior-tamir:
Fixes #

## Proposed Changes

  -change file name: **A**zuredeploy to **a**zuredeploy so deployment link would work
  -Change playbookname default value without spaces
  -


Yaniv-Shasha (approved):
reviewed and tested

PR #1215, opened 2020-10-26
Create WindowsSystemTimeChange.yaml

shainw:
Adding in Time Change event check for Hunting.

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Hunting Queries/SecurityEvent/WindowsSystemTimeChange.yaml
+++ b/Hunting Queries/SecurityEvent/WindowsSystemTimeChange.yaml
@@ -1,0 +1,19 @@
+id: 9fd6f61d-2cc3-48de-acf5-7194e78d6ea1
+name: Windows System Time changed on hosts
+description: |
+  'Identifies when the system time was changed on a Windows host which can indicate potential timestomping activities.'
+requiredDataConnectors:
+  - connectorId: SecurityEvents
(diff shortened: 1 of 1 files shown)


Hunting Queries/SecurityEvent/WindowsSystemTimeChange.yaml
```diff
@@ -0,0 +1,19 @@
+id: 9fd6f61d-2cc3-48de-acf5-7194e78d6ea1
+name: Windows System Time changed on hosts
+description: |
+  'Identifies when the system time was changed on a Windows host which can indicate potential timestomping activities.'
```
ashwin-patil, line 4: Looks good.

You may want to add a note on requirement on 4616 event id. It is not present in common or minimal so All Events mode must be selected for this event id while collecting Windows Events via connector and auditing must be enabled for this event id.
https://docs.microsoft.com/en-us/azure/sentinel/connect-windows-security-events

ashwin-patil (changes requested):

Hunting Queries/SecurityEvent/WindowsSystemTimeChange.yaml
shainw: Done

PR #1250, opened 2020-11-01
Add Intelligence workbook for Intsights

ShirSabag:
Add workbook for insights IOC overview


Workbooks/IntsightsIOCWorkbook.json
```diff
@@ -0,0 +1,657 @@
+{
```
Amitbergman, line 1: It is not working for me in a workspace I tried it in, the fiels "ReportingFeeds" does not exist.


image

Workbooks/WorkbooksMetadata.json
```diff
@@ -1049,5 +1049,18 @@
...
+  },
+  {
+    "workbookKey": "IntsightsIOCWorkbook",
+    "logoFileName": "IntSights_logo.svg",
+    "description": "",
+    "dataTypesDependencies": [ "ThreatIntelligenceIndicator", "SecurityAlert" ],
+    "dataConnectorsDependencies": [ "ThreatIntelligenceTaxii" ],
+    "previewImagesFileNames": [ "IntsightsIOCWhite.png", "IntsightsMatchedWhite.png", "IntsightsMatchedBlack.png", "IntsightsIOCBlack.png"],
+    "version": "1.0",
+    "title": "IntSights IOC Dashboard",
```
Amitbergman, line 1061: Consider changing to "workbook".

Workbooks/IntsightsIOCWorkbook.json
```diff
@@ -0,0 +1,657 @@
...
+        "comparison": "isEqualTo",
+        "value": "correlation_overview"
+      },
+      "name": "Correlation Overview"
+    }
+  ],
+  "fallbackResourceIds": [
+    "/subscriptions/df5ebcb7-3cc5-4549-b82d-2ee2a62cf4de/resourcegroups/resource_group_for_sentinel_clint/providers/microsoft.operationalinsights/workspaces/azure-sentinel-clint"
+  ],
+  "fromTemplateId": "sentinel-IntsightsIOCDashboard",
```
Amitbergman, line 655: Consider changing to "workbook".

Workbooks/IntsightsIOCWorkbook.json
ShirSabag: mmm yes you are right, in case that you don't have any reporting_feeds fields you will get an error, I can found a solution for that (like a default value to 'All' = [])
basically, the only problem is that we see an error instead of 'no data to show'
if there is any way to fix that please let me know, but I don't think it's a problem because there is no data to show anyway

the ReportingFeed field exists only if you will add a new 'Threat intelligence - TAXII' connection with our URL, you can use that to see how the workbook should be look like:

url: https://devapi.intsights.com/tip/integrations/taxii2/5fb405d635869500073c525a
collection id: 5fb405fd4c6c8e000738e16c
username: 5d70fcf999293c0007fb427d
password: d08da09cedf09f152f029fe041714b0208269de1bc7a1f6cb789edabbabce83c5b752539e5a5f1b8d19b26a24110ba5bb9166a6d1c8527d991eee23499f49960

 Basically, it should only be used by our customers - Intsights company

Amitbergman: What you can do is:
1. Making the parameter reporting feeds required:
image

2. Add a default value to all, for example - All

image

Then it will look like this:
image

Amitbergman: It is still not perfect but I think it is good enough, the goal is that customers who do not have this field will have an empty experience and not broken experience.

ShirSabag: 1. I updated the reporting_feeds to be required
2. About the 'Select All value' - it kind of weird if 'All' filter will be select nothing, instead of select all.
it's not a default value, this is the value he receives in case that we have at least one option or nothing

Amitbergman (approved):

ShirSabag:
Just want to know when this PR will be merged? and when can we told our customers that they can use it

Amitbergman (approved):

PR #1258, opened 2020-11-03
Fix tcpdump command not found check

NoamLandress:
Fix the check in case "Tcpdump" command is not installed.


DataConnectors/CEF/cef_troubleshoot.py
```diff
@@ -260,11 +260,6 @@ def handle_tcpdump_line(line, incoming_port, ok_message):
             "Notice: To tcp dump manually execute the following command - \'tcpdump -A -ni any port " + incoming_port + " -vv\'")
         time.sleep(1)
         return True
-    # Handle command not found
```
ofshomro, line 263: do we use handle_tcpdump_line in another place? maybe its better to leave it here as well

morshabi (approved):

PR #1279, opened 2020-11-09
Recorded Future - Azure Playbooks Set 1

AdrianRecFut:
## Proposed Changes
  - Adding 3 new playbooks for importing various datasets from Recorded Future into tiIndicators table from Azure Sentinel
  - Adding 1 new playbook for Enrichment of IPs from Azure Sentinel Alerts


sarah-yo (changes requested):
Can you please remove the metadata property from all of your ARM templates? This is causing our JSON validation checks to fail.

e.g.

"metadata": {
    "comment": "This playbook leverages the Recorded Future API and automatically imports the Actively Communicating C&C Server IP RiskList, as Threat Intelligence Indicators, for detection purposes in Azure Sentinel",
    "author": "Adrian Porcescu, Recorded Future"
  }

Thanks!

AdrianRecFut:
@sarah-yo - i have made all the requested changes so please let me know if now those files can be committed or there are any additional aspects that need to be corrected.
I also have 1 question in regards to the changes you asked for. I noticed that other committed playbooks do have that information added (actually taking them as a model lead me to adding some in mine). Is there anything that changed meanwhile, as that metadata seems to be present in some of the published ones.
Thank you,
Adrian

sarah-yo (changes requested):
The JSON file check is still failing as it looks like you have a BOM on every UTF-8 file.

You can fix this with an extension in VS - https://marketplace.visualstudio.com/items?itemName=SergeyVlasov.FixFileEncoding 

Re: the metadata comment, you're absolutely right that other templates have it, I need to check why it doesn't like this one. If you have time it might be worth removing the BOM from the .json files and adding the metadata back in to see if the BOM was affecting both

AdrianRecFut:
@sarah-yo - I have changed the encoding of all files to UTF-8 (no BOM) and added the metadata back.
It seems that the checks have been passed now, so please do let me know if there is anything else that I should be changing or if i have any next steps in order to finalise the process.
Thank you,
Adrian

PR #1289, opened 2020-11-11
Barium related IP/Domain IOC's

aprakash13:
Fixes #

## Proposed Changes

  -
  -
  -


Detections/MultipleDataSources/BariumDomainIOC.yaml
```diff
@@ -0,0 +1,125 @@
+id: 70b12a3b-4899-42cb-910c-5ffaf9d7997d 
+name: Known Barium domains  
+description: | 
+  'Identifies a match across various data feeds for domains IOCs related to the Barium activity group.' 
+   References: https://www.justice.gov/opa/pr/seven-international-cyber-defendants-including-apt41-actors-charged-connection-computer'
```
petebryan, line 5: Don't think you need the ' at the end of the first line - worried this might break something.

Detections/MultipleDataSources/BariumIPIOC.yaml
```diff
@@ -0,0 +1,141 @@
+id: 6ee72a9e-2e54-459c-bc9a-9c09a6502a63 
+name: Known Barium IP 
+description: | 
+  'Identifies a match across various data feeds for IP IOCs related to the Barium activity group.  
+References: https://www.justice.gov/opa/pr/seven-international-cyber-defendants-including-apt41-actors-charged-connection-computer'
```
petebryan, line 5: Is References meant to be part of the description field rather than its own heading?

aprakash13: Made it part of description now. Thanks..

PR #1291, opened 2020-11-11
Updating Product Parsers

briandelmsft:
Product parsers used long datatype for EventCount, changed to int to comply with schema


shainw (approved):
Approved, but can we fix the aka.ms link as it goes to the generic Azure Sentinel page - https://docs.microsoft.com/azure/sentinel/ versus to the schema normalization.
The network normalization schema link is - https://docs.microsoft.com/azure/sentinel/normalization-schema

briandelmsft:
I emailed the owner of that aka.ms link to ask for the change... I'm suggesting https://docs.microsoft.com/en-us/azure/sentinel/normalization as the destination for the link as it includes more setup information as well as a link to the detailed schema.

PR #1301, opened 2020-11-12
Added missing fields in Better MTD sample data

aymenim:
Fixes missing fields in Better MTD sample data


aymenim:
Thanks @nazang  for the quick response.

PR #1303, opened 2020-11-12
Adding Onapsis Connector and Workbook

cbiguet:
Creation of Onapsis Connector and Workbook. I have a datatable that is meant to be saved as a Kusto function (used to enrich events). The only folder where I saw functions being stored was in "Parsers", this is where I added mine. This is probably not ideal since this function isn't actually a parser, so if there is a better place for me to put it please let me know!


DataConnectors/OnapsisOP.json
```diff
@@ -0,0 +1,240 @@
+{
+
+    "id": "OnapsisOP",
+
+    "title": "Onapsis OP",
```
v-jayakal, line 5: Can you give full form in the title (OP)? Is the appliance name OSP or OP?

v-jayakal (changes requested):
OnapsisLookup - Can you fix the usage instructions in this parser, it has a reference about Infoblox NIOS. Also, can you give the usage of the incident_lookup function in OnapsisAlarmsOveriview, could not find it in the json.

DataConnectors/OnapsisOP.json
cbiguet: Currently in the logs, the appliance is defined as OSP but our appliance is called OP (Onapsis Platform). Should I change it to "Onapsis OSP' or "Onapsis Platform"?

cbiguet: - Will fix the usage instructions in the parser, thanks for pointing that out
- The incident_lookup function is used at line 322 in OnapsisAlarmsOverview.json

cbiguet: Currently in the logs, the appliance is defined as OSP but our appliance is called OP (Onapsis Platform). Should I change it to "Onapsis OSP' or "Onapsis Platform"?

v-jayakal: Onapsis Platform will be meaningful without abbreviations.

cbiguet: Just committed the changes, let me know what you think

v-jayakal: Could you please provide the usage of "incident_lookup"- Its not used in workbook as mentioned in the Kusto function definition. Where is this function used?

cbiguet: If you take a look at line 322 of the workbook file (OnapsisAlarmsOverview.json) - incident_lookup is used in a query - it's goal being to provide context into what the incident means.

Workbooks/OnapsisAlarmsOverview.json
```diff
@@ -0,0 +1,384 @@
...
+        "queryType": 0,
+        "resourceType": "microsoft.operationalinsights/workspaces"
+      },
+      "name": "query - 7"
+    },
+    {
+      "type": 3,
+      "content": {
+        "version": "KqlItem/1.0",
+        "query": "incident_lookup | where IncidentName == '{incident_name}'",
```
cbiguet, line 322: Here is the query I am referring to

v-jayakal: Thank you.

v-jayakal (approved):

nazang (dismissed):

PR #1316, opened 2020-11-16
Bugfix/fix multiple mock data sent

NoamLandress:
Because sending the mock data was inside a while loop it would generate sometimes hundreds of requests.
I've entered the mock sending into a for loop and limited the mock data sending to 5 sends maximum.


DataConnectors/CEF/cef_troubleshoot.py
```diff
@@ -291,8 +293,9 @@ def incoming_logs_validations(incoming_port, ok_message, mock_message=False):
     poll_obj = select.poll()
     poll_obj.register(tcp_dump.stdout, select.POLLIN)
     while (end_seconds - start_seconds) < tcpdump_time_restriction:
```
ofshomro, line 295: Do we still need this while? why not change all the logic to "for 5 times"?

NoamLandress: Because we have 2 previous tcpdumps (on port 514 and port 25226).
These tcpdumps check if data is flowing in (without any mock generation)
they need those 60 seconds while loop of polling to see if any data is received.

morshabi (approved):

PR #1319, opened 2020-11-16
Feature/lahisham/migrate scheduled templates to new entity mapping

laithhisham:
# Proposed Changes
  - Migrate all the detection scheduled templates to support the new entity mapping model.
  - This is part of the new custom entity mapping in the Alert Enrichment feature we are currently working on.

I will send a detailed mail on the the new format and instructions to use it.


.script/tests/detectionTemplateStructureValidation/Models/ValidEntityMappingsAttribute.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        protected override ValidationResult IsValid(object value, ValidationContext validationContext)
+        {
+            if (value == null)
+            {
+                return ValidationResult.Success;
+            }
+
+            var entityMappings = (List<EntityMapping>)value;
```
ehudk-msft, line 33: consider try catch for casting exception

.script/tests/detectionTemplateStructureValidation/Models/ValidEntityMappingsAttribute.cs
```diff
@@ -0,0 +1,76 @@
...
+            var fieldName = validationContext.MemberName;
```
ehudk-msft, line 34: should you validate that validationContext is not null?

ehudk-msft (changes requested):
looking good

.script/tests/detectionTemplateStructureValidation/Models/ScheduledTemplateInternalModel.cs
```diff
@@ -38,6 +39,10 @@ public class ScheduledTemplateInternalModel : AnalyticsTemplateInternalModelBase
         [JsonProperty("triggerThreshold", Required = Required.Always)]
         [Range(0, 10000)]
         public int TriggerThreshold { get; set; }
+
+        [JsonProperty("entityMappings", Required = Required.Default, NullValueHandling = NullValueHandling.Ignore)]
+        [ValidEntityMappings(entityMappingsMinLength: 1, entityMappingsMaxLength: 5, fieldMappingsMinLength: 1, fieldMappingsMaxLength: 3)]
```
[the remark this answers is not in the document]
laithhisham: no, it needs to be at least one, otherwise it is invalid

.script/tests/detectionTemplateStructureValidation/Models/ValidEntityMappingsAttribute.cs
```diff
@@ -0,0 +1,76 @@
...
+    public class ValidEntityMappingsAttribute : ValidationAttribute
+    {
+        private readonly int _entityMappingsMinLength;
+        private readonly int _entityMappingsMaxLength;
+        private readonly int _fieldMappingsMinLength;
+        private readonly int _fieldMappingsMaxLength;
+        private readonly int _laColumnNameMaxLength = 500; // 500 is the max length of a column name in LA
+
+        // field mapping column name should start with a letter or an underscore and contain only alphanumeric English characters (i.e. [a-zA-Z0-9_])
+        private readonly Regex _laColumnNameRegex = new Regex("^[a-zA-Z_]+\\w*$");
```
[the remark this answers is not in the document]
laithhisham: correct, it can contain numbers but not in the beginning. the \w gives us exactly that. so 'SHA1' is a valid column name

.script/tests/detectionTemplateStructureValidation/Models/ScheduledTemplateInternalModel.cs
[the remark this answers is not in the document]
ehudk-msft: can't I have a rule without entity mapping?

.script/tests/detectionTemplateStructureValidation/Models/ValidEntityMappingsAttribute.cs
laithhisham: we won't reach here if the deserialization fails. So we are sure it casts.
Even if it doesn't we would throw anyway in the catch since we want to fail the validation in that case.. so it won't make any difference

.script/tests/detectionTemplateStructureValidation/Models/ValidEntityMappingsAttribute.cs
laithhisham: it won't be. it gets called by the runtime and it will always be passed

Detections/CommonSecurityLog/MultiVendor-PossibleDGAContacts.yaml
```diff
@@ -111,4 +111,9 @@ query: |
...
     on Name
     | project StartTime, EndTime, Name, DGADomain, SourceIP, DestinationIP, DataSource
-    | extend timestamp=StartTime, IPCustomEntity=SourceIP
\ No newline at end of file
+    | extend timestamp=StartTime, IPCustomEntity=SourceIP
+entityMappings:
+  - entityType: IP
+    fieldMappings:
+      - identifier: Address
+        columnName: IPCustomEntity
```
shainw, line 119: Are we ready to add in the DNS.DomainName entity yet?  It would be great if we could include this -
`| extend DNSCustomEntity = Name`
  \- entityType: DNS
    fieldMappings:
         - identifier: DomainName
       columnName: DNSCustomEntity

shainw (approved):

Detections/CommonSecurityLog/MultiVendor-PossibleDGAContacts.yaml
laithhisham: great point @shainw 
Sure, this should be available in the new mapping model. However, it won't work in the old one, but it is totally fine. we will be moving to the new mapping model slowly until we are fully using it. For now, users with the new model would benefit from it, and it won't have an effect on the old one, so feel free to use it. More details about all the supported entity types in the mail I sent.

In the meantime, since we will live with both mappings for a while, we would still need to maintain both mapping models. So we still want to have the **CustomEntity in the query. Exactly as you did above. Once we are only working with the new mapping model (it will take a while), we can use any column names we want for the entity fields :)

ehudk-msft (approved):

PR #1332, opened 2020-11-18
ACN_CD_NetskopeParser01

chicduong:
## Proposed Changes

  - Netskope Parser Function


shainw (approved):
We prefer to have a log sample included, but the parser appears to be good.

chicduong:
@shainw I'm including the sample in the connector. I believe I mistakenly submitted the other way around for this connector.

PR #1335, opened 2020-11-19
Removed BIN as this can miss bruteforces spread across BIN windows

orhanmoye:
Fixes #
Removed the BIN, and replaced this with a check to see if the bruteforce occurred during a 20 min window.  This is now performed using datetime math instead.  
## Proposed Changes
In some instance, a bruteforce activity can occur that is of a short duration (< 20 mins) but it spans two BIN windows.  Because it spans two BIN windows, the detection rule will not fire.

For Example:
User A attempt 5 failed logins from 1:19 to 1:20.  This is across a duration of 1 minute but because of the BIN function, the 5 events will be split across two windows.  1:00 and 1:20.  Because of this, the threshold is not met.  
  -
  -
  -

### DIFF
--- a/Detections/SigninLogs/SigninBruteForce-AzurePortal.yaml
+++ b/Detections/SigninLogs/SigninBruteForce-AzurePortal.yaml
@@ -1,4 +1,4 @@
-id: 28b42356-45af-40a6-a0b4-a554cdfd5d8a
+﻿id: 28b42356-45af-40a6-a0b4-a554cdfd5d8a
 name: Brute force attack against Azure Portal
 description: |
   'Identifies evidence of brute force activity against Azure Portal by highlighting multiple authentication failures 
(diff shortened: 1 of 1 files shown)


shainw:
@orhanmoye - Apologies for the delay in responding on this one.  We have had some other high priority items over the last several weeks that kept us away from review items.  We don't want to removing the bin concept in this, but agree that there is opportunity to miss.  For that, we think it best to use either [Window functions](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/windowsfunctions) or [User Analytics](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/useranalytics).  Over the next few weeks we will investigate and share back.

orhanmoye:
No problem.  Thanks for the update.

shainw:
Closing this as it is out of date now, we have made changes and are looking at one more change to help with the bin() size causing misses.  This will come in the next week.

PR #1354, opened 2020-11-23
Update ProcessEntropy.yaml

shainw:
Improving Process Entropy performance, removing additional common processes to speed execution, improving ratio's to reduce common processes further and handling Weight calculation slightly differently.  Lower Weight is still indicative of High Entropy or Rarity.

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Hunting Queries/SecurityEvent/ProcessEntropy.yaml
+++ b/Hunting Queries/SecurityEvent/ProcessEntropy.yaml
@@ -16,77 +16,112 @@ tactics:
   - Execution
 query: |
 
-  // May need to reduce the number of days if the environment is very large.  Try 3-5 days, but less accuracy.
-  let end = startofday(now());
-  let start = end - 7d;
-  let Exclude = SecurityEvent
-  // Timeframe is set so that results do not change during the same day (UTC time)
-  | where TimeGenerated >= start and TimeGenerated <= end
-  | where EventID == 4688
-  | summarize ExcludeCompCount = dcount(Computer),  ExcludeProcCount = count() by Process 
-  // Removing noisy processes for an environment, adjust as needed
-  | where ExcludeProcCount >= 2000 and ExcludeCompCount > 2
-  ;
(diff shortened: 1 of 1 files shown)


Hunting Queries/SecurityEvent/ProcessEntropy.yaml
```diff
@@ -16,77 +16,112 @@ tactics:
...
+  // Remove or increase value to see processes with low entropy, meaning more common. 
+  | where Weight <= 100
+  | project Computer, Process, Weight , ProcessEntropy, AllHostsProcessCount, ProcessCountOnHost, DistinctHostsProcessCount; 
+  // Join back full entry 
+  Results 
+  | join kind= inner ( 
+  SecEvents
+  | project TimeGenerated, EventID, Computer, SubjectUserSid, Account, AccountType, Process, NewProcessName, CommandLine, ParentProcessName, _ResourceId, SourceComputerId 
+  ) on Computer, Process 
+  | project TimeGenerated, EventID, Computer, SubjectUserSid, Account, AccountType, Weight, ProcessEntropy,
```
ashwin-patil, line 123: Is there a way to group final results (specially  timestamp, commandline) so that final output has 1 row per rare process without loosing context.

Sample output showed multiple rows (40-50) for single process  (e.g. MOMPerfSnapshotHelper.exe,  MicrosoftEdgeUpdate.exe, WmiApSrv.exe, TiWorker.exe, gpupdate.exe) with same weight and parent but different timestamp and same commandline or commadline with minor cardinality.

ashwin-patil (commented):
This was giant query, looks good.
minor comment to group outputs per process, otherwise ok to merge.

Hunting Queries/SecurityEvent/ProcessEntropy.yaml
shainw: Done

PR #1359, opened 2020-11-24
Darktrace connector

SunRift:
Fixes #
NA
## Proposed Changes

  adding relevant files so that the Darktrace Connector can be approved for Preview:
  - logo svg
  - connector json
  - sample data (sanitised)


preetikr:
@SunRift - Thanks, for your submission. Please sign the CLA so that the PR can be approved and merged.

Logos/Darktrace.svg
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0"?>
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 89.76 99.09">
+  <defs>
+    <style>.cls-1,.cls-4,.cls-5{opacity:0.5;}.cls-1{fill:url(#linear-gradient);}.cls-2{fill:url(#linear-gradient-2);}.cls-3{fill:#f49c14;}.cls-4{fill:url(#linear-gradient-3);}.cls-5{fill:url(#linear-gradient-4);}</style>
```
v-jayakal, line 4: Invalid logo, Can you please correct it.

The company logo has to adher to the following requirements:
Ensure raw file of logo does not have any of the following:
cls and style formats
embedded png formats
xmlns:xlink
data-name
Do not use xlink:href - use inline instead
Do not use title tag
If some properties in the logo have IDs (for e.g. <g id="layer0"...), then set these IDs as GUIDs so that these are uniquely identifiable across all Azure logo assets
Please refer guidelines here:  https://github.com/Azure/Azure-Sentinel/blob/e1665d4c4442b430d6847adc65f4b9c25f7b071f/DataConnectors/ReadMe.md.

v-jayakal (changes requested):

Logos/Darktrace.svg
SunRift: I see, removed the xmlns:xlink

Logos/Darktrace.svg
```diff
@@ -0,0 +1,7 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 90.79 99.09">
```
v-jayakal, line 1: The logo is not clear, since it does not have colors, only orange color dot is emphasized. Can you confirm the output (look and feel) of the logo? 
DarkTrace

v-jayakal (changes requested):

Logos/Darktrace.svg
v-jayakal: Can you please verify the attached file and confirm if it is as expected. If it is, please update the logo here.
[Darktrace-New.zip](https://github.com/Azure/Azure-Sentinel/files/5646216/Darktrace-New.zip)

v-jayakal (changes requested):

Logos/Darktrace.svg
nazang: All we ask is to change  fill="#fff"  ==>   fill="#2f3a56"
Because the logo background is white and it is not visible.

nazang: @SunRift, could you help to verify the proposed logo. We want to start the deployment process on our side ASAP.

SunRift: Yes that looks great, I've updated my PR with the new .svg

SunRift:
The logo should be updated now.

We support both posting CEF and the custom log's JSON on our end: would it be better for me to add the additional log type now, into the connector JSON? 
Currently for simplicity I was going to submit with just the CEF, as most connectors support only this, and once approved edit the connector JSON with the additional custom log format. 

Which way would be quicker to process on your end?
 We are expecting the CEF ( ie Common Security Event) format to be more popular, but the custom log solution is easier for customers to set up, and so we want to provide both as options.

Thanks

v-jayakal (approved):

nazang:
> The logo should be updated now.
> 
> We support both posting CEF and the custom log's JSON on our end: would it be better for me to add the additional log type now, into the connector JSON?
> Currently for simplicity I was going to submit with just the CEF, as most connectors support only this, and once approved edit the connector JSON with the additional custom log format.
> 
> Which way would be quicker to process on your end?
> We are expecting the CEF ( ie Common Security Event) format to be more popular, but the custom log solution is easier for customers to set up, and so we want to provide both as options.
> 
> Thanks

@SunRift, different type of connectors will in two connector files. But, I noticed the current connector title is called Darktrace, If you are planning to add another connector with Rest API, you may need to think of modifying the Title to reduce the confusion. 

@preetikr, do you have more comments on this?

DataConnectors/Darktrace.json
```diff
@@ -0,0 +1,114 @@
+{
+    "id": "DarktraceDarktrace",
+    "title": "Darktrace",
+    "publisher": "Darktrace",
+    "descriptionMarkdown": "Common Event Format (CEF) is an industry standard format on top of Syslog messages, used by many security vendors to allow event interoperability among different platforms. By connecting your Darktrace config to Azure Sentinel, you can take advantage of search & correlation, alerting, and threat intelligence enrichment for each log.",
```
preetikr, line 5: Please update the description to something more specific for Darktrace, in terms of log types being pulled in, etc.

preetikr:
> > The logo should be updated now.
> > We support both posting CEF and the custom log's JSON on our end: would it be better for me to add the additional log type now, into the connector JSON?
> > Currently for simplicity I was going to submit with just the CEF, as most connectors support only this, and once approved edit the connector JSON with the additional custom log format.
> > Which way would be quicker to process on your end?
> > We are expecting the CEF ( ie Common Security Event) format to be more popular, but the custom log solution is easier for customers to set up, and so we want to provide both as options.
> > Thanks
> 
> @SunRift, different type of connectors will in two connector files. But, I noticed the current connector title is called Darktrace, If you are planning to add another connector with Rest API, you may need to think of modifying the Title to reduce the confusion.
> 
> @preetikr, do you have more comments on this?

@SunRift - Including both data connectivity options in the same data connector would be confusing customer experience. I'd recommend you have a separate data connector for the REST API connectivity. Particularly because the data from this REST API connectivity will land in custom table. You might also want to cross reference the two connectors from each other as well (in the description so that customers can choose to connect with one as needed). Another aspect from customer experience and your maintenance standpoint would be to have a parser / Kusto Function created so that the content you develop can be the same for both data streams and a better approach would be to even transform the columns to normalized schema as well. Please reach out to azuresentinelpartner_at_microsoft_dot_com so that we can schedule a meeting to talk about this. For now, let's complete the CEF data connector and get this PR approved and merged.

nazang (approved):

v-jayakal (approved):

v-jayakal (approved):

nazang (approved):

PR #1364, opened 2020-11-24
ACN_CD_SymantecDLP_Parser01

chicduong:
## Proposed Changes

  - Symantec DLP Parser
  - Symantec DLP Sample Data


Parsers/SymantecDLP/SymantecDLP.txt
```diff
@@ -0,0 +1,51 @@
...
+// 
+// LOG SAMPLES:
+// This parser assumes the raw log are formatted as follows:
+//
+//<13>Sep 5 08:22:01 data.example.com CEF:0|Symantec|DLP|12.5.0|ruleID|Policy|5|BLOCKED=Passed INCIDENT_ID=204529 INCIDENT_SNAPSHOT=https://main.example.com/Path/Address MATCH_COUNT=3 PROTOCOL=SMTP RECIPIENTS=email@example.com SENDER=sender@example.com SUBJECT=Sensitive Data (attachment included) SEVERITY=1:High FILE_NAME=myfile.xml
+//
+// 
+CommonSecurityLog
+| where DeviceVendor == "Symantec" and DeviceProduct == "DLP"
+| extend Parser = extract_all(@"(^|;)(?P<key>[^=]+)\=(?P<value>[^;]+)(;|$)", dynamic(["key","value"]), AdditionalExtensions)
```
shainw, line 27: I don't think this is working as expected on the AdditionalExtensions field.  When I simply use the sample entry you have above and put that in my own table with a column named AdditionalExtensions, it doesn't appear to parse it properly.

image

If we then do the mv-apply and projecting of columns, they are all empty -

image

shainw (changes requested):

Parsers/SymantecDLP/SymantecDLP.txt
chicduong: @shainw - I uploaded a new parser, reverting back to the old approach.

shainw (approved):
Looks like it works with the strings in the included JSON, but still does not work on your Log Sample from the Parser.txt file itself.  This looks to be due to the separator being a space instead of a semi-colon which is what is used in the JSON file.

PR #1379, opened 2020-11-29
ACN_CD_Netskope_Connector04

chicduong:
**Checkpoint File was not being created in ARM template deployments**

## Proposed Changes

 The current code, creates the Checkpoint file in the working directory (C:\home\site\wwwroot\AzureFunctionNetskope). For ARM template deployment, anything from the C:\home\site\wwwroot and below is linked directly to contents in the Azure Github for Netskope, which means this directory is read-only. The Checkpoint File cannot write to this directory since it is a static directory and not unique to each Azure Function App deployment.

The proposed changed moves the Checkpoint file directory to C:\home\site, where it can be written.


nazang (changes requested):
Is c:\ drive always the root for the function?

The value listed here is d:\ drive.

https://github.com/Azure/azure-functions-host/wiki/Retrieving-information-about-the-currently-running-function

Consider to use the environment variable FunctionAppDirectory or  (get-location).Drive.Root 
instead of hard-code the drive.

chicduong:
A majority of my Function Apps were C: but I did see some D: , I wasn't able to identify why one was selected over the other, so I just updated the code to take the root drive letter of the current working directory

nazang (approved):

PR #1382, opened 2020-11-30
Adding ARM template samples for Connectors and Analytics Rules

javiersoriano:
Adding sample ARM templates for the following:

- Analytics Rules (Scheduled, ML Behavior Analytics, Fusion, Microsoft)

- Connectors (AATP, MDATP, MCAS, O365, ASC)


javiersoriano:
JSON and link validations are expected to fail as we have included links to appropriate documentation within the templates. Templates work fine when deployed.

PR #1402, opened 2020-12-04
New low severity daily detection for AAD StsRefresh modification

itsreallynick:
Adding this initial anomalous StsRefreshToken modification rule that may accompany and aid in triaging other suspicious AAD & OAuth detections

TODO: properly extract LastDirSyncTimeNew as dateTime and compare this value to now() to highlight extremely long refresh tokens to highlight extraordinarily long refresh token validation timeframes

### DIFF
--- a/Detections/AuditLogs/StsRefreshTokenModification.yaml
+++ b/Detections/AuditLogs/StsRefreshTokenModification.yaml
@@ -31,8 +31,10 @@ query: |
   | extend InitiatingUserOrApp = iff(isnotempty(InitiatedBy.user.userPrincipalName),tostring(InitiatedBy.user.userPrincipalName), tostring(InitiatedBy.app.displayName))
   | extend targetUserOrApp = TargetResources[0].userPrincipalName
   | extend eventName = tostring(TargetResourcesModProps.displayName)
-  | extend LastDirSyncTimeOld = todatetime(parse_json(tostring(TargetResourcesModProps.oldValue))[0])
-  | extend LastDirSyncTimeNew = todatetime(parse_json(tostring(TargetResourcesModProps.newValue))[0])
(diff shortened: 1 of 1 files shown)


shainw (approved):
@itsreallynick - I made some changes to improve efficiency and to fix the validations errors, basically Log Analytics is much more forgiving and will let you get away with mv-expand and calling those expanded items with underscores, the validation doesn't let that pass.  Check it out, should still do exactly what you need.

itsreallynick:
Looks great, thanks for cleaning up the variable names for the mv-expand and for fixing the dateTime values.
I also added two helpful time deltas to allow users to more easily identify anomalies.
I tested current version in several environments. Should be good to ship. Thanks!

timbMSFT (dismissed):

shainw (approved):

EmilienCourt:
Hi,

I was checking the backlog regarding this rule, to understand it a bit more.

Could you specify which field (out of tokenMinutesRemaining and tokenMinutesAdded) should be considered important to look when hunting for malicious activity, and why ?

How can be tokenMinutesRemaining != 0 ?
Knowing that, according to [Microsoft documentation](https://docs.microsoft.com/en-us/troubleshoot/azure/active-directory/federated-users-forced-sign-in#powershell) "StsRefreshTokenValidFrom will appear to accept any date, but it will always set to the current date and time.". Which means that TimeGenerated == newStsRefreshValidFrom, so tokenMinutesRemaining = 0, right ?

Leaving tokenMinutesRemaining out, we can only look for anomalies in tokenMinutesAdded.

However, could you precise what would a high tokenMinutesAdded value indicate ? An app which renewed a very old Refresh Token ?



Thanks a lot in advance for your answer,

Regards

@shainw @itsreallynick

PR #1410, opened 2020-12-04
ACN_CD_CylancePROTECT_Parser01

chicduong:
## Proposed Changes

  - Blackberry CylancePROTECT Parser
  - Sample Data


Parsers/CylancePROTECT/CylancePROTECT.txt
```diff
@@ -0,0 +1,144 @@
...
+        PolicyName = extract(@"Policy Name: ([^,]+)(,|$)",1,SyslogMessage);
+let AuditLog = LogHeader
+| where EventType == "AuditLog"
+| extend Provider = extract(@"Provider: ([^,]+),",1,SyslogMessage),
+        PolicyChanged = extract(@"Policy Changed: ([^,]+),",1,SyslogMessage),
+        ZoneNames = split(extract(@"Zone: ([^,]+)(,|$)",1,SyslogMessage),","),
+        UserName = extract(@"User: ([^,]+)(,|$)",1,SyslogMessage),
+        DeviceName = extract(@"Devices?:\s([\w\d\-\_]+)\,",1,SyslogMessage),
+        SrcIpAddr = split(extract(@"Source IP: ([\w\.\:]+)\,",1,SyslogMessage),",");
+let AllOtherLogs = LogHeader
```
shainw, line 92: May want to consider populating the EventType field here with something like "OtherType' if it is empty.  It may help query efficiency later.

shainw (dismissed):
One recommended change, but not required.

Parsers/CylancePROTECT/CylancePROTECT.txt
chicduong: @shainw added Events that do not have a Event Type listed as "Other. Also added the capture of other Event Types that may not be defined

shainw (approved):

PR #1413, opened 2020-12-05
ACN_CD_QualysKB_Parser01

chicduong:
## Proposed Changes

  - Qualys KB Parser
  + Sample Data


shainw (approved):
**Not required, but recommended** - If Detections are added later, validations will fail on all the Detection YAML's because the table schema cannot be identified since this is a Custom Log. While not required here, to prevent any issues with future Detection addition, you may want to add a Custom Table schema here - https://github.com/Azure/Azure-Sentinel/tree/master/.script/tests/KqlvalidationsTests/CustomTables. They are simple JSON format. The name of the table goes at that top in the name field, in your case it would be "QualysKB_CL". Then the schema follows in the Properties key value pair section.

shainw:
> **Not required, but recommended** - If Detections are added later, validations will fail on all the Detection YAML's because the table schema cannot be identified since this is a Custom Log. While not required here, to prevent any issues with future Detection addition, you may want to add a Custom Table schema here - https://github.com/Azure/Azure-Sentinel/tree/master/.script/tests/KqlvalidationsTests/CustomTables. They are simple JSON format. The name of the table goes at that top in the name field, in your case it would be "QualysKB_CL". Then the schema follows in the Properties key value pair section.

@chicduong - do you want to include this or do later?

chicduong:
@shainw 

> 
> 
> > **Not required, but recommended** - If Detections are added later, validations will fail on all the Detection YAML's because the table schema cannot be identified since this is a Custom Log. While not required here, to prevent any issues with future Detection addition, you may want to add a Custom Table schema here - https://github.com/Azure/Azure-Sentinel/tree/master/.script/tests/KqlvalidationsTests/CustomTables. They are simple JSON format. The name of the table goes at that top in the name field, in your case it would be "QualysKB_CL". Then the schema follows in the Properties key value pair section.
> 
> @chicduong - do you want to include this or do later?

@shainw I added the file, but I'm still unclear on what property key value pair should be added here. Should I add all the ones we're expecting?

shainw:
> @shainw
> 
> > > **Not required, but recommended** - If Detections are added later, validations will fail on all the Detection YAML's because the table schema cannot be identified since this is a Custom Log. While not required here, to prevent any issues with future Detection addition, you may want to add a Custom Table schema here - https://github.com/Azure/Azure-Sentinel/tree/master/.script/tests/KqlvalidationsTests/CustomTables. They are simple JSON format. The name of the table goes at that top in the name field, in your case it would be "QualysKB_CL". Then the schema follows in the Properties key value pair section.
> > 
> > 
> > @chicduong - do you want to include this or do later?
> 
> @shainw I added the file, but I'm still unclear on what property key value pair should be added here. Should I add all the ones we're expecting?

@chicduong - Sorry for not being clear.  Yes, you want to include the entire schema, so yes all of them.  Within reason.  If someone submits a detection with a field that is new after this, then the validation will fail, but we can simply add that into the JSON when that occurs.  Make sense?

chicduong:
@shainw updates made to the validation file.

nazang (approved):

nazang:
> @shainw updates made to the validation file.

@shainw, the new added schema looks good to me. But, please verify as well. Thank you1

shainw (approved):

PR #1418, opened 2020-12-07
Timeseries Performance fixes 

ashwin-patil:
Fixes #

## Proposed Changes

  -  Created new temp list of time windows to use for filtering
  - Added Filtering on the base data to improve filtering
  -


shainw (approved):

pemontto:
This PR introduced non-breaking whitespaces which work OK if we copy paste in the GUI, however when uploading the text directly to the Sentinel API you'll get a 500 error without any detail.

https://github.com/wortell/AZSentinel/issues/158

PR #1423, opened 2020-12-08
Updated Oauth hunting query

petebryan:
## Proposed Changes
Added additional filtering and summarization to Oauth hunting query as by default it is very noisy and difficult to hunt through.


shainw (approved):
Typo to fix, but approved for the query

PR #1450, opened 2020-12-11
Using has_any will miss values

orhanmoye:
Current query using has_any is looking for AntiPhish, SafeAttachment and will miss values like: 

AntiPhishingRule, 
AntiPhishingPolicy
SafeAttachmentPolicy
Etc.

We need to either use the full value (which I don't have and may change) or use a contains check to cater for the various values.  I have implemented the contains approach.

Fixes #

## Proposed Changes

  -use of contains for value checking. 
  -
  -

### DIFF
--- a/Detections/OfficeActivity/office_policytampering.yaml
+++ b/Detections/OfficeActivity/office_policytampering.yaml
@@ -21,13 +21,20 @@ relevantTechniques:
   - T1089
 query: |
   let timeframe = 1d;
+  let opList = OfficeActivity 
+  | where TimeGenerated >= ago(timeframe) 
+  | summarize by Operation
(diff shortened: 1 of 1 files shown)


shainw (commented):
@orhanmoye - we really want to try to stay away from using contains, especially on data types that can have large volume.  I am going to modify this slightly so that we get the list of available values first and then run it thru via the in~ operator.

shainw (approved):
@orhanmoye - Take a look at my updates, they should do what you want with better perf on large data sets.

orhanmoye:
> @orhanmoye - Take a look at my updates, they should do what you want with better perf on large data sets.

Looks good.  thanks

PR #1472, opened 2020-12-16
Adding in Schema support for M365 tables

shainw:
In order to support detections that are being included for M365, we need to temporarily add the schema manually.

Fixes #

## Proposed Changes

  -
  -
  -


shainw:
Completing so we can support some current critical detections coming in, can rollback at a later date when the schema is fully recognized.

PR #1490, opened 2020-12-16
Update GitHub hunting query for correct column name.

BenMcGarry:
Column name is Country instead of Location in latest updates to the GH Function connector.

Fixes #

## Proposed Changes

  - Swap Location > Country to be compatible with latest GH function changes.


sreedharande:
@BenMcGarry - verified against the Parser https://github.com/Azure/Azure-Sentinel/blob/master/Parsers/GitHub/GitHubAuditLog.txt.

PR #1495, opened 2020-12-17
Validate workspace given as parameter

NoamLandress:
Adding a check in the Troubleshooting script to validate the agent is indeed connected to the workspace given as parameter from the customer. The check is copied from a similar one in the installation script called- "check_multi_homing" (fixed documentation typo as well)


shainw (approved):

shainw:
Very old PR, closing

PR #1510, opened 2020-12-19
New queries and some fixes

petebryan:
New queries:
 - Suspected export of mail archive via OWA
 - ReadMail permissions added to an application hunting
 - Host activity exporting on-prem mailboxes then covering tracks
 - ADFS key export detection with Sysmon

Minor fixes elsehwere


shainw (approved):
All checks passed except Workbook which there are none here and the failure is because the NPM failed, not any actual check.  Approving and completing.

PR #1524, opened 2020-12-23
Update MFADisable.yaml

KustoKing:
If an adversary is able to disable MFA then I would like to get notified within an hour.

In some cases I set this rule to 5 min.

Fixes #

## Proposed Changes

  -
  -
  -


shainw (approved):
Seems reasonable

PR #1530, opened 2020-12-24
Fix permissions function decoding

NoamLandress:
Hotfix for permissions check function in the cef_troubleshooting script.
Fixes the support for python 3 and doesn't affect python 2


YaronFruchtmann (approved):
Approved by Ofir Shomron

PR #1541, opened 2020-12-30
Update FirstAppOrServicePrincipalCredential.yaml

KustoKing:
The used dash is unicode 2013 which results in : â€“

Fixes #

## Proposed Changes

  -
  -
  -


shainw (approved):
Thanks for spotting.

PR #1543, opened 2020-12-30
Create System_Shutdown_Reboot(Sysmon)

chihebchebbi:
Fixes #
First Try/Practice


shainw (changes requested):
@chihebchebbi - First, thanks for contributing! Please check out the [Sysmon parser](https://github.com/Azure/Azure-Sentinel/tree/master/Parsers/Sysmon) and I recommend you use that in this query versus doing your own parsing, although not required.  Additionally, this doesn't meet the bar to be a detection on its own as it will create many hits.  Please move to the Hunting folder.  Once moved, I will make comments on the query itself.  Thanks, Shain

shainw:
Closing, new version submitted - #1691

PR #1548, opened 2021-01-04
Data Connector validator checks

v-rucdu:
Fixes #

## Proposed Changes

  - Generic (Required attributes) schema check for connectors
  - Space check in connector ID , Data Type and File name
  -  Permissions match with the permissions in the respective template


.script/utils/dataConnectorCheckers/dataTypeChecker.ts
```diff
@@ -0,0 +1,20 @@
+import { DataType } from "../dataConnector";
+import { DataConnectorValidationError } from "../validationError";
+
+export function isValidDataType(dataTypes: Array<DataType>): boolean {
```
erant10, line 4: where is this function called?

.script/dataConnectorValidator.ts
```diff
@@ -0,0 +1,38 @@
...
+let fileKinds = ["Added", "Modified"];
+let CheckOptions = {
+  onCheckFile: (filePath: string) => {
+    return IsValidDataConnectorSchema(filePath);
+  },
+  onExecError: async (e: any, filePath: string) => {
+    console.log(`Data Connector Validation Failed. File path: ${filePath}. Error message: ${e.message}`);
+  },
+  onFinalFailed: async () => {
+    logger.logError("An error occurred, please open an issue");
```
erant10, line 34: FYI there is a bug where this message is thrown for every validation error (all of them).
see comment - https://github.com/Azure/Azure-Sentinel/pull/946#issuecomment-675134655
however the fix should be a separate work item and PR

.script/utils/dataConnector.ts
```diff
@@ -0,0 +1,189 @@
+export interface DataConnectorModel {
```
erant10, line 1: what are all these definitions for? are they part of the validation?

erant10 (changes requested):

.script/tests/dataConnectorValidatorTest/dataConnectorValidator.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+chai.use(chaiAsPromised);
+
+describe("dataConnectorValidator", () => {
+  it("should pass when dataConnectorSchema.json is valid", async () => {
+    await checkValid(".script/tests/dataConnectorValidatorTest/testFiles/validDataConnectorSchema.json");
+  });
+
+  it("should skip .json file when missing ID or connectivityCriterias attributes", async () => {
+    await checkValid(".script/tests/dataConnectorValidatorTest/testFiles/inValidDataConnectorSchema.json");
```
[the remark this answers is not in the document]
v-maudan: There would be multiple json files in data connector folder. There is no specific file naming convention by which we can identify data connector json, so we are checking ID and connectivityCriterias to identify the same and skipping the other json files.

.script/utils/dataConnector.ts
v-maudan: These are part of data connector model and we would needed for validation, as we add more test cases.

.script/utils/dataConnectorCheckers/dataTypeChecker.ts
v-maudan: updated.
following are the CEF and Syslog data connector which is having spaces in Datatype.
https://github.com/Azure/Azure-Sentinel/blob/master/DataConnectors/ThycoticSecretServer_CEF.json    
https://github.com/Azure/Azure-Sentinel/blob/master/DataConnectors/WireXsystemsNFP(1b).json     - CEF
https://github.com/Azure/Azure-Sentinel/blob/master/DataConnectors/CiscoISE/Connector_Cisco_ISE.json  - Syslog

.script/utils/dataConnector.ts
erant10: OK. If possible try to avoid adding classes and type definitions pre-emptively for future test cases, since it is unclear why they are part of the current PR.

.script/tests/dataConnectorValidatorTest/dataConnectorValidator.test.ts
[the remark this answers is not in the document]
erant10: I see. It sounds like a potential bug. If these 2 fields are missing or misspelled in a connector json file, the test would skip the file and pass validation even though it is invalid. however I don't see an easy way to identify connector json (other than a naming convention).
I'm fine to live with this bug but in needs to be more clear. see my other notes in the validator.

.script/dataConnectorValidator.ts
```diff
@@ -0,0 +1,40 @@
+import fs from "fs";
+import { runCheckOverChangedFiles } from "./utils/changedFilesValidator";
+import { ExitCode } from "./utils/exitCode";
+import { isValidSchema } from "./utils/jsonSchemaChecker";
+import { isValidId } from "./utils/dataConnectorCheckers/idChecker";
+import { isValidDataType } from "./utils/dataConnectorCheckers/dataTypeChecker";
+import * as logger from "./utils/logger";
+
+export async function IsValidDataConnectorSchema(filePath: string): Promise<ExitCode> {
+  let dataConnector = JSON.parse(fs.readFileSync(filePath, "utf8"));
```
erant10, line 10: variable name is misleading since it may not be a connector json. rename to generic jsonFile.

.script/dataConnectorValidator.ts
```diff
@@ -0,0 +1,40 @@
...
+  let schema = JSON.parse(fs.readFileSync(".script/utils/schemas/DataConnectorSchema.json", "utf8"));
+if(typeof dataConnector.id != "undefined" && typeof dataConnector.connectivityCriterias != "undefined")
```
erant10, line 12: Please declare a helper function for clarity
```
if (isPotentialConnectorJson(jsonFile) {
// ...
}
```

.script/dataConnectorValidator.ts
```diff
@@ -0,0 +1,40 @@
...
+{
+  isValidSchema(dataConnector, schema);
+  isValidId(dataConnector.id);
+  isValidDataType(dataConnector.dataTypes);
+}
+else{
+  console.log(`Skipping File path: ${filePath}`);
```
erant10, line 19: please change to `console.warn` and make message more informative.
```
console.warn(`Could not identify json file as a connector. Skipping File path: ${filePath}`)
```

.script/utils/dataConnectorCheckers/dataTypeChecker.ts
erant10: based on the example it seems that if there is a space inside the braces of the data type then it's ok. It seems we don't need to invalidate in that case. is that correct?

v-maudan: Yes, We are checking there should not be spaces in data type of Rest API and Azure function data connector.

nazang (approved):

.script/utils/dataConnectorCheckers/permissionsChecker.ts
```diff
@@ -0,0 +1,115 @@
...
+            "description": "Read and write permissions to Azure Functions to create a Function App is required. [See the documentation to learn more about Azure Functions](https://docs.microsoft.com/azure/azure-functions/)."
+        }
+    ]
+};
+
+export function isValidPermissions(permissions: RequiredConnectorPermissions, connectorCategory: any) {   
+    
+    switch(connectorCategory)
+    {
+        case ConnectorCategory.CEF:
```
erant10, line 80: why is this case empty?

.script/dataConnectorValidator.ts
```diff
@@ -0,0 +1,83 @@
...
+  {
+    return ConnectorCategory.SysLog;
+  }
+  else if(dataTypes[0].name.endsWith("_CL"))
+  {
+    let fileDirectory:string = filePath.substr(0, filePath.lastIndexOf('/'));
+    let isAzureFunction:boolean = false;
+
+    // reads all the file names in the directory to verify the existence of zip folder
+    fs.readdirSync(fileDirectory).forEach(file => {
```
erant10, line 55: This logic is a bit fragile. Is there a better way to check if it is an azure function? if the PR author doesn't include the zip in the folder, this will assume it is not an azure function and check the wrong permissions.

erant10 (dismissed):

.script/dataConnectorValidator.ts
v-rucdu: Modified logic to check if there is Deploy to Azure button in JSON, than consider it as Azure Function connector

erant10 (dismissed):

v-rucdu:
@nazang : We updated the logic to identify the Azure Function Connectors, hence your vote got reset. Can you please approve the PR again? Thanks!!!

v-jayakal (dismissed):

PR #1551, opened 2021-01-05
ACN_CD_CrowdStrikeFalcon_Parser01

acnccd:
## Proposed Changes

+ CrowdStrike Falcon Endpoint Protection Parser
+ Sample Data

### DIFF
--- a/Parsers/CrowdStrikeFalconEventStream/CrowdstrikeFalconEventStream.txt
+++ b/Parsers/CrowdStrikeFalconEventStream/CrowdstrikeFalconEventStream.txt
@@ -1,0 +1,75 @@
+// Title:           CrowdStrike Falcon Endpoint Protection Parser
+// Author:          Microsoft
+// Version:         1.0
+// Last Updated:    01/04/21
+// Comment:         Initial Release
+//  
+// DESCRIPTION:
+// This parser takes raw CrowdStrike Falcon Event Stream logs from a Syslog (CEF) stream and parses the logs into a normalized schema.
+//
+// USAGE:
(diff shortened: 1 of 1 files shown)


v-jayakal (dismissed):

shainw (changes requested):
@acnccd - Apologies on the delay here, a bit behind on reviews.   I cannot test the parser as the sample data set doesn't have the columns needed.  For example, you are parsing DeviceCustomNumber*, DeviceCustomDate*, DeviceCustomString*, but these are not included.  I only see DeviceCustomIPv6Address* and DeviceCustomFloatingPoint*.  The AdditionalExtensions parsing seems to work fine, but there is no example of SessionStartTime or SessionEndTime in the field, so I cannot confirm that specific part functions properly.

acnccd:
> 
> 
> @acnccd - Apologies on the delay here, a bit behind on reviews. I cannot test the parser as the sample data set doesn't have the columns needed. For example, you are parsing DeviceCustomNumber*, DeviceCustomDate*, DeviceCustomString*, but these are not included. I only see DeviceCustomIPv6Address* and DeviceCustomFloatingPoint*. The AdditionalExtensions parsing seems to work fine, but there is no example of SessionStartTime or SessionEndTime in the field, so I cannot confirm that specific part functions properly.

@shainw, sample data was usually submitted for data exports post-parsing. I will try to generate more sample data to export pre-parsed data.

acnccd:
@shainw new sample data uploaded

acnccd:
@shainw any updates on this?

shainw (approved):

PR #1559, opened 2021-01-07
New Playbook Added to Synchronize Compromised Passwords to the Azure AD Banned Password List using the paid DeHashed API

SCStelz:
Added new Playbook for Synchronizing Compromised Passwords to the Azure AD Banned Password List using the paid DeHashed API.


sreedharande:
@SCStelz - Instead of requesting KeyVault URI for BIGDATACLOUD.COM API KEY as parameter, requesting Name of the Secret/Key/Certificate would be the best practice

SCStelz:
How would only the secret name allow secret retrieval without using the native Key Vault connector? I'm using standard HTTP connectors for simplicity and secret retrieval via URI parameter just as in the example here - https://github.com/Azure-Samples/azure-ad-conditional-access-apis/blob/main/03-deploy/readme.md.

PR #1571, opened 2021-01-12
New rules and queries for Cisco Umbrella, Cisco ISE and Proofpoint POD

vu-socprime:
Added new detection rules and hunting queries for Cisco Umbrella, Cisco ISE and Proofpoint POD

### DIFF
--- a/Detections/ProofpointPOD/ProofpointPODHighRiskNotDiscarded.yaml
+++ b/Detections/ProofpointPOD/ProofpointPODHighRiskNotDiscarded.yaml
@@ -14,11 +14,11 @@ triggerThreshold: 0
 tactics:
   - InitialAccess
 query: |
-let lbtime = 10m;
-ProofpointPOD
-| where TimeGenerated > ago(lbtime)
-| where EventType == 'message'
-| where NetworkDirection == 'inbound'
-| where FilterDisposition !in ('reject', 'discard')
-| where FilterModulesSpamScoresOverall == '100'
-| project SrcUserUpn, DstUserUpn
+  let lbtime = 10m;
+  ProofpointPOD
+  | where TimeGenerated > ago(lbtime)
+  | where EventType == 'message'
+  | where NetworkDirection == 'inbound'
+  | where FilterDisposition !in ('reject', 'discard')
+  | where FilterModulesSpamScoresOverall == '100'
+  | project SrcUserUpn, DstUserUpn
--- a/Detections/ProofpointPOD/ProofpointPODHighRiskNotDiscarded.yaml
+++ b/Detections/ProofpointPOD/ProofpointPODHighRiskNotDiscarded.yaml
@@ -1,0 +1,24 @@
+id: c7cd6073-6d2c-4284-a5c8-da27605bdfde
+name: ProofpointPOD - High risk message not discarded
+description: |
+  'Detects when email with high risk score was not rejected or discarded by filters.'
+severity: Low
+requiredDataConnectors:
+  - connectorId: ProofpointPOD
+    dataTypes:
+      - ProofpointPOD_message_CL
+queryFrequency: 10m
+queryPeriod: 10m
+triggerOperator: gt
+triggerThreshold: 0
+tactics:
+  - InitialAccess
+query: |
+let lbtime = 10m;
+ProofpointPOD
+| where TimeGenerated > ago(lbtime)
+| where EventType == 'message'
+| where NetworkDirection == 'inbound'
+| where FilterDisposition !in ('reject', 'discard')
+| where FilterModulesSpamScoresOverall == '100'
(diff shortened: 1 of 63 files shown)


vu-socprime:
During KqlValidation there is an error "The name 'ProofpointPOD' does not refer to any known table, tabular variable or function". But all queries are based on functions (ProofpointPOD is the name of function) located in "Parsers" directory. Could you please tell if it is possible to pass KqlValidation in this situation?

shainw:
> During KqlValidation there is an error "The name 'ProofpointPOD' does not refer to any known table, tabular variable or function". But all queries are based on functions (ProofpointPOD is the name of function) located in "Parsers" directory. Could you please tell if it is possible to pass KqlValidation in this situation?

@vu-socprime - You will need to create a CustomTable mapping for these as these connectors are not built into the build system at this point.  Instructions are here - https://github.com/Azure/Azure-Sentinel#pull-request-kql-validation-check - specifically this -

`If you are using custom logs table (a table which is not defined on all workspaces by default) you should verify your table schema is defined in json file in the folder Azure-Sentinel\.script\tests\KqlvalidationsTests\CustomTables`

vu-socprime:
> > During KqlValidation there is an error "The name 'ProofpointPOD' does not refer to any known table, tabular variable or function". But all queries are based on functions (ProofpointPOD is the name of function) located in "Parsers" directory. Could you please tell if it is possible to pass KqlValidation in this situation?
> 
> @vu-socprime - You will need to create a CustomTable mapping for these as these connectors are not built into the build system at this point. Instructions are here - https://github.com/Azure/Azure-Sentinel#pull-request-kql-validation-check - specifically this -
> 
> `If you are using custom logs table (a table which is not defined on all workspaces by default) you should verify your table schema is defined in json file in the folder Azure-Sentinel\.script\tests\KqlvalidationsTests\CustomTables`

@shainw, json files for custom tables are already in Azure-Sentinel\.script\tests\KqlvalidationsTests\CustomTables. The issue is that we use functions (Azure-Sentinel/Parsers/) instead of tables because we need normalized fields to make queries. What should we do in this situation?

shainw:
@vu-socprime - You can include it in the CustomTable, but it would be a JSON with the schema of the Parser versus the Table.  At least I believe this should work.  Give it a try and see if it allows passing of the validation.

vu-socprime:
@shainw - Ok, I will try it. Thank you.

shainw:
> @shainw - Ok, I will try it. Thank you.

@vu-socprime - I think you did the wrong one.  I see you submitted for Cisco_Umbrella but the KQL validation error is on the ProofPointPOD missing a customtable entry.

vu-socprime:
> > @shainw - Ok, I will try it. Thank you.
> 
> @vu-socprime - I think you did the wrong one. I see you submitted for Cisco_Umbrella but the KQL validation error is on the ProofPointPOD missing a customtable entry.

@shainw - There were errors regarding Cisco_Umbrella, ProofPointPOD and CiscoISEEvent. I've also added schemas for  ProofPointPOD and CiscoISEEvent in last commit.

Detections/ProofpointPOD/ProofpointPODHighRiskNotDiscarded.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - InitialAccess
+query: |
+  let lbtime = 10m;
+  ProofpointPOD
+  | where TimeGenerated > ago(lbtime)
+  | where EventType == 'message'
+  | where NetworkDirection == 'inbound'
+  | where FilterDisposition !in ('reject', 'discard')
+  | where FilterModulesSpamScoresOverall == '100'
+  | project SrcUserUpn, DstUserUpn
```
shainw, line 24: Missing the Entity Mapping section on this one.

Detections/ProofpointPOD/ProofpointPODWeakCiphers.yaml
```diff
@@ -0,0 +1,19 @@
...
+queryFrequency: 1h
+queryPeriod: 1h
+triggerOperator: gt
+triggerThreshold: 0
+query: |
+  let lbtime = 1h;
+  let tls_ciphers = dynamic(['RC4-SHA', '​DES-CBC3-SHA']);
+  ProofpointPOD
+  | where EventType == 'message'
+  | where TlsCipher in (tls_ciphers)
```
shainw, line 19: Missing the Entity Mapping section

shainw (changes requested):
Looks like referencing the parser in the CustomTable worked.  Additionally, a couple of required changes for Detections with the Entity Mapping.  Also, to support a future improvement we want to include Entity Mapping on Hunting queries also now.  This is not required, but when it does become required we would need you to come back and do it then, so may want to do it now.

shainw (approved):

PR #1580, opened 2021-01-14
Delete StsRefreshTokenModification.yaml

itsreallynick:
Original purpose of the low priority, hunting-only query was to demonstrate an event that has token validity time periods in it (we don't expose SAML token lifetimes in Sentinel or other products currently), and demonstrate how one could monitor for anomalous/edited tokens. This event will only fire on the manual expiration of the StsRefreshToken by an admin (or the user) and has caused some confusion. In the spirit of the original contribution, people have learned more about token visibility ;)
Recommend retiring this and removing from Sentinel hunting workbook/blog.

Fixes #

## Proposed Changes

  -
  -
  -


itsreallynick:
Or... just remove from hunting blog/notebook, and leave here as hunting so people can explore/modify.

PR #1588, opened 2021-01-15
oktagovchages

v-jayakal:
Okta Gov Changes


shainw:
@v-jayakal - your CLA is not completed, you need to do this in order for PR to pass all checks.

DataConnectors/Okta Single Sign-On/Connector_REST_API_FunctionApp_Okta.json
```diff
@@ -131,7 +131,7 @@
...
...pp Name and select **Configuration**.\n2. In the **Application settings** tab, select **+ New application setting**.\n3. Add each of the following four (4) application settings individually, with their respective string values (case-sensitive): \n\t\tapiToken\n\t\tworkspaceID\n\t\tworkspaceKey\n\t\turi\n\t\tlogAnalyticsUri\n - Use the following schema for the `uri` value: `https://<OktaDomain>/api/v1/logs?since=` Replace `<OktaDomain>` with your domain. [Click here](https://developer.okta.com/docs/reference/api-overview/#url-namespace) for further details on how to identify your Okta domain namespace. There is no need to add a time value to the URI, the Function App will dynamically append the inital start time of logs to UTC 0:00 for the current UTC date as time value to the URI in the proper format.\n - Note: If using Azure Key Vault secrets for any of the values above, use the`@Microsoft.KeyVault(SecretUri={Security Identifier})`schema in place of the string values. Refer to [Key Vault references documentation](https://docs.microsoft.com/azure/app-service/app-service-key-vault-references) for further details. \n4. Once all application settings have been entered, click **Save**."
```
nazang, line 134: logAnalyticsUri, please add an example how to set the value, a sample will be helpful.

DataConnectors/Okta Single Sign-On/azuredeploy_OktaSingleSignOn_API_FunctionApp_V2.json
```diff
@@ -193,7 +195,11 @@
                         },
                         {
                             "name": "WEBSITE_RUN_FROM_PACKAGE",
-                            "value": "https://aka.ms/sentineloktaazurefunctioncodev2"
+                            "value": "https://github.com/v-jayakal/Azure-Sentinel/blob/v-jayakal/oktaGovTest/DataConnectors/Okta%20Single%20Sign-On/AzureFunctionOktaSSO_V2/AzureFunctionOktaSSO.zip?raw=true"
```
nazang, line 198: just a note, we need to modify this back after customer tested with the private change.

DataConnectors/Okta Single Sign-On/AzureFunctionOktaSSO_V2/AzureFunctionOktaSSO/run.ps1
```diff
@@ -144,14 +145,18 @@ do {
             -method $method `
             -contentType $contentType `
             -resource $resource
-        $LAuri = "https://" + $customerId + ".ods.opinsights.azure.com" + $resource + "?api-version=2016-04-01"
+		if ([string]::IsNullOrEmpty($logAnalyticsUri))
```
sreedharande, line 148: This check is not required, as we are dynamically generating LA URI in ARM Template and send it as environment variable to Azure function. The value will not be null or empty

DataConnectors/Okta Single Sign-On/azuredeploy_OktaSingleSignOn_API_FunctionApp_V2.json
```diff
@@ -204,19 +210,6 @@
                    "alwaysOn": true
               }
           },
-          {
```
sreedharande, line 207: This will create a Web App and expose it through Private Endpoint. 
Parameterize "name": "[concat(variables('FunctionName'), '/', variables('FunctionName'), parameters('webapp_dns_name'))]",
if environment.portal is .com use .net
else use the extension like (.us)

nazang: Sreedhar, we remove this dns settings purposely, because this is an azure function working as a worker role and it doesn't expose any endpoint. Plus, the code is adding the default host name binding which is in place. We are also planning to remove this in the template. Do you know any previous reason why this is added in the first place?

DataConnectors/Okta Single Sign-On/AzureFunctionOktaSSO_V2/AzureFunctionOktaSSO/run.ps1
nazang: The package will be shared by new installation and existing installation, we will do an in-place upgrade. This check is for backward compatibility, for existing customer with already installed azure function, the current function doesn't have this environment variable defined.

v-jayakal: Resolving as the change is needed.

DataConnectors/Okta Single Sign-On/azuredeploy_OktaSingleSignOn_API_FunctionApp_V2.json
v-jayakal: No change is needed here, hence resolving.

v-jayakal:
> should it be \.ods\.opinsights\.azure ?

Modified, it was workign even without escape.

PR #1597, opened 2021-01-18
IntSights Workbook | New version | Add filters and IOCs list

ShirSabag:
## Proposed Changes

  - Add filters by severity for each tab
  - Add IOCs list with search & export CSV
  - Update the screenshots


Workbooks/IntsightsIOCWorkbook.json
```diff
@@ -606,7 +439,7 @@
...
                     "size": 1,
                     "title": "Top Malwares Linked with Matched Indicators",
```
shainw, line 444: Probably should be Malware and not Malwares

shainw (approved):
Nothing blocking, but probably want to avoid using the term Malwares and just go with Malware.  Will approve and merge as is though since we are a bit behind on reviews and this has been waiting.

PR #1607, opened 2021-01-20
Query to identify attempts to modify registry ACL to evade security s…

aprakash13:
…olutions

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/MultipleDataSources/SecurityServiceRegistryACLModification.yaml
+++ b/Detections/MultipleDataSources/SecurityServiceRegistryACLModification.yaml
@@ -11,8 +11,7 @@ description: |
      - https://docs.microsoft.com/windows/security/threat-protection/auditing/audit-registry
      - https://docs.microsoft.com/windows/security/threat-protection/auditing/event-4670
      - https://github.com/OTRF/Set-AuditRule 
-     - https://docs.microsoft.com/en-us/dotnet/api/system.security.accesscontrol.registryrights?view=dotnet-plat-ext-5.0
-     For the above event to be created the audit policy for the registry must have auditing enabled for Write DAC and/or Write Owner'
+     - https://docs.microsoft.com/dotnet/api/system.security.accesscontrol.registryrights?view=dotnet-plat-ext-5.0'
 severity: High
 requiredDataConnectors:
   - connectorId: SecurityEvents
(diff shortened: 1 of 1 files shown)


Detections/MultipleDataSources/SecurityServiceRegistryACLModification.yaml
```diff
@@ -0,0 +1,86 @@
...
+  (union isfuzzy=true
+  (
+  SecurityEvent
+  | where TimeGenerated >= ago(timeframe)
+  | where EventID == 4670
+  | where ObjectType == 'Key'
+  | where ObjectName has_any (servicelist)
+  | parse EventData with * 'OldSd">' OldSd "<" *
+  | parse EventData with * 'NewSd">' NewSd "<" *
+  | extend Reason =
```
shainw, line 52: Using [case()](https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/casefunction) here would be better

Detections/MultipleDataSources/SecurityServiceRegistryACLModification.yaml
```diff
@@ -0,0 +1,86 @@
...
+     The detection leverages Security Event as well as MDE data to identify when specific security services registry permissions are modified. 
+     Only some portions of this detection are related to Solorigate, it also includes coverage for some common tools that perform this activity. 
+     Reference on guidance for enabling registry auditing:
+     - https://docs.microsoft.com/windows/security/threat-protection/auditing/advanced-security-auditing-faq
+     - https://docs.microsoft.com/windows/security/threat-protection/auditing/appendix-a-security-monitoring-recommendations-for-many-audit-events
+     - https://docs.microsoft.com/windows/security/threat-protection/auditing/audit-registry
+     - https://docs.microsoft.com/windows/security/threat-protection/auditing/event-4670
+     - https://github.com/OTRF/Set-AuditRule 
+     - https://docs.microsoft.com/en-us/dotnet/api/system.security.accesscontrol.registryrights?view=dotnet-plat-ext-5.0
+     For the above event to be created the audit policy for the registry must have auditing enabled for Write DAC and/or Write Owner'
```
shainw, line 15: This should go with the event-4670 line not under all the references.

Detections/MultipleDataSources/SecurityServiceRegistryACLModification.yaml
```diff
@@ -0,0 +1,86 @@
...
+     The detection leverages Security Event as well as MDE data to identify when specific security services registry permissions are modified. 
+     Only some portions of this detection are related to Solorigate, it also includes coverage for some common tools that perform this activity. 
+     Reference on guidance for enabling registry auditing:
+     - https://docs.microsoft.com/windows/security/threat-protection/auditing/advanced-security-auditing-faq
+     - https://docs.microsoft.com/windows/security/threat-protection/auditing/appendix-a-security-monitoring-recommendations-for-many-audit-events
+     - https://docs.microsoft.com/windows/security/threat-protection/auditing/audit-registry
+     - https://docs.microsoft.com/windows/security/threat-protection/auditing/event-4670
+     - https://github.com/OTRF/Set-AuditRule 
+     - https://docs.microsoft.com/en-us/dotnet/api/system.security.accesscontrol.registryrights?view=dotnet-plat-ext-5.0
```
shainw, line 14: Need to remove the /en-us/ from the URL, this is why the validations are failing.

Detections/MultipleDataSources/SecurityServiceRegistryACLModification.yaml
```diff
@@ -0,0 +1,86 @@
...
+  - DefenseEvasion
+relevantTechniques:
+  - T1562
+tags: 
+  - Solorigate
+query: |
+ 
+  let servicelist = dynamic(['Services\\HealthService', 'Services\\Sense', 'Services\\WinDefend', 'Services\\MsSecFlt', 'Services\\DiagTrack', 'Services\\SgrmBroker', 'Services\\SgrmAgent', 'Services\\AATPSensorUpdater' , 'Services\\AATPSensor', 'Services\\mpssvc']);
+  let filename = dynamic(["subinacl.exe",'SetACL.exe']);
+  let parameters = dynamic (['/deny=SYSTEM', '/deny=S-1-5-18', '/grant=SYSTEM=r', '/grant=S-1-5-18=r', 'n:SYSTEM;p:READ', 'n1:SYSTEM;ta:remtrst;w:dacl']);
```
shainw, line 38: This is good for now, but could be fragile.  After this version is published verify that a regex match is not needed to handle if there are spaces include in the parameters being passed in.  Such as "/deny = SYSTEM" or "n:SYSTEM; p:READ"

shainw (changes requested):

Detections/MultipleDataSources/SecurityServiceRegistryACLModification.yaml
aprakash13: Thanks Shain. Done now

Detections/MultipleDataSources/SecurityServiceRegistryACLModification.yaml
aprakash13: Done now. Thanks

Detections/MultipleDataSources/SecurityServiceRegistryACLModification.yaml
aprakash13: Ack. Will work on it

Detections/MultipleDataSources/SecurityServiceRegistryACLModification.yaml
aprakash13: Done Now. Thanks.

shainw (dismissed):

PR #1609, opened 2021-01-20
Updating and fixing parsers

YaronFruchtmann:
Fixes #

## Proposed Changes

  - Bugs in ZScaler,PA, Empty
  - Changes to schema data types 
  -


shainw (approved):
Do we want to change these to YAML format as discussed or wait for the next go round?

YaronFruchtmann:
No, let's not wait since the table is already available. We will reformat all parser in a later merge

PR #1620, opened 2021-01-21
Add non ascii validations

oshvartz:
This is will validate detection file to verify they do not include any non-ascii chars that will not display correctly


YaronFruchtmann:
This is an important validation, as we saw stray chars added to files.
Can you apply the validation to all other types too?
Specifically: Expansion queries, Insights, Activities, Hunting, Parsers

oshvartz:
added more folders validation

PR #1621, opened 2021-01-21
Cyberpion detection

cyberpion-yotam:
Added a detection using the Cyberpion connector


cyberpion-yotam:
@v-rucdu As requested, here is the detection.

Detections/CyberpionSecurityLogs/HighUrgencyActionItems.yaml
```diff
@@ -0,0 +1,38 @@
...
+     | where TimeGenerated > ago(timeframe)
+     | summarize max(bin(TimeGenerated, time_generated_bucket))
+     );
+  CyberpionActionItems_CL
+   | where TimeGenerated > ago(timeframe) and is_open_b == true
+   | where bin(TimeGenerated, time_generated_bucket) == maxTimeGeneratedBucket
+   | where urgency_d >= min_urgency
+   | extend timestamp = opening_datetime_t
+entityMappings:
+  - entityType: DNS
```
aprakash13, line 35: The entity has to be extracted from your query results and should generally be present in the last extend statement. Could you please check if that can be done here. Also, looking at the column name field "host_s" sounds like this should be a HostCustomEntity with a entity type of Host vs DNS. If that is the case would you please alter this section accordingly.
Here are some more details regarding this: https://github.com/Azure/Azure-Sentinel/wiki/Query-Style-Guide#entitymappings for general guidance regarding this. 
Thanks.

cyberpion-yotam: Hey, thanks for the reply.
host_s isn't a "Host, Computer, System, Device" in the connector. It is a domain or subdomain (any of the following - www.example.com / www.mail.example.com / example.com). 
Is there a field other than DomainName I can map it to?

Regarding the extension of the DNS mapping, I'd be happy to extend the host_s field name to be something like DNSCustomEntity just for readability. Is what you mean?

Thanks!

cyberpion-yotam: Hello?

aprakash13: Thanks Yotam for the explanation around the host_s field. That makes sense. If there is a link to a schema reference, we could probably also include that in the description section of the query. 

For the DNS Entity type the identifier is DomainName. 
Yes. If you could  please extend the host_s field name to DNSCustomEntity and have the entity mapping to somewhat like below that would be great. Thanks again.

- entityType: DNS
    fieldMappings:
      - identifier: DomainName
        columnName: DNSCustomEntity

cyberpion-yotam: Done.
About the schema reference, we don't have one right now and I think that by extending host_s = DNSCustomEntity like you said, it becomes clear.

By the way, I think maybe you should add the new step of adding the Connector to "ValidConnectorIds.json" to the steps in the documentation (if that is what I should have done).

Thanks again!

aprakash13: Thanks Yotam.

cyberpion-yotam: So can I get an approval?

cyberpion-yotam:
@aprakash13 Do you need anything further from me to approve this detection?

cyberpion-yotam:
Helloooo?

shainw:
> Helloooo? @aprakash13 @shainw @NoamLandress

@cyberpion-yotam - Apologies on the missed follow up on this.  Simply slipped out of view for Ajeet, but he is on it today and will get a response out.

aprakash13 (approved):
The changes look good.

PR #1639, opened 2021-01-22
office timegenerated fix

ashwin-patil:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/OfficeActivity/MailItemsAccessedTimeSeries.yaml
+++ b/Detections/OfficeActivity/MailItemsAccessedTimeSeries.yaml
@@ -49,9 +49,17 @@ query: |
       | where TimeGenerated > ago(2d)
       | extend DateHour = bin(TimeGenerated, 1h)
       | where OfficeWorkload=~ "Exchange" and Operation =~ "MailItemsAccessed" and ResultStatus =~ "Succeeded"
-  ) on $left.TimeGenerated == $right.DateHour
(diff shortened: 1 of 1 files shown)


Detections/OfficeActivity/MailItemsAccessedTimeSeries.yaml
```diff
@@ -46,9 +46,10 @@ query: |
...
   | join (
       OfficeActivity
-  | where TimeGenerated > ago(2d)
-  | where OfficeWorkload=~ "Exchange" and Operation =~ "MailItemsAccessed" and ResultStatus =~ "Succeeded"
-  ) on TimeGenerated
+      | where TimeGenerated > ago(2d)
+      | extend DateHour = bin(TimeGenerated, 1h)
+      | where OfficeWorkload=~ "Exchange" and Operation =~ "MailItemsAccessed" and ResultStatus =~ "Succeeded"
+  ) on $left.TimeGenerated == $right.DateHour
 entityMappings:
```
shainw, line 53: Think we should include the legacy entity mapping items as those are still needed and IPCustomEntity is not mapped to anything so we need that at a minimum - | extend timestamp = TimeGenerated, AccountCustomEntity = UserId, IPCustomEntity = ClientIP.  Also, ClientIP includes the Port in the field, you will want to remove that so entity mapping can occur properly.

shainw (changes requested):

shainw:
@ashwin-patil - Reminder to have a look at my suggestion. :)

shainw (approved):

PR #1642, opened 2021-01-24
Update WorkbooksMetadata.json

thbanasi:
Fixes #

## Proposed Changes

  -
  -
  -


Amitbergman (approved):
Consider changing to Microsoft and changing the version to 1.0 as it is not deployed yet.

PR #1663, opened 2021-01-28
ioc queries

jannieli:
Fixes #

## Proposed Changes

  - ZINC IOC queries

### DIFF
--- a/Detections/MultipleDataSources/ZincJan272021IOCs.yaml
+++ b/Detections/MultipleDataSources/ZincJan272021IOCs.yaml
@@ -36,7 +36,7 @@ tactics:
   - CommandAndControl
   - Execution
 relevantTechniques:
-  - T1102
+  - T1071.001
   - T1204
 query: |
   let timeframe = 1d;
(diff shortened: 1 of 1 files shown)


Detections/MultipleDataSources/ZincJan272021IOCs.yaml
```diff
@@ -0,0 +1,120 @@
...
+      - DeviceFileEvents
+  - connectorId: MicrosoftThreatProtection
+    dataTypes:
+      - DeviceNetworkEvents
+queryFrequency: 1d
+queryPeriod: 1d
+triggerOperator: gt
+triggerThreshold: 0
+tactics:
+  - CommandAndControl
```
shainw, line 36: T1204 is part of the Execution tactic, should CommandAndControl be here?  If so, that Technique is associated with it

shainw (changes requested):
In general, we are not doing a project on each of these datatypes, so the results set could have many columns, we should consider reducing to only required columns from each datatype and it may may sense to rename similar column names for ease of reading results.  Also, project Type for each so that the user knows what data type it came form.

shainw (approved):

PR #1678, opened 2021-01-28
🐛 Fix query and more NBSPs

pemontto:
Fix more NBSPs and update `ADFSDomainTrustMods.yaml` query so period/frequency match the configured/updated `auditLookback`.

Also updates the conditional in `ExternalUserAddedRemovedInTeams.yaml` to match the description. That is where an account is removed **after** it was added.

Removes detection `CorrelateIPC_Unfamiliar-Atypical` which was an older version of `CorrelateIPC_Unfamiliar-Atypical.yaml`.


shainw:
@pemontto - please merge master, there are many changes and some of your items have been fixed in other PR's.

shainw (approved):

PR #1684, opened 2021-01-29
Add projects and datatype

jannieli:
Fixes #

## Proposed Changes

  - Add project statements to each table to show only most essential columns
  - Add DataType column to disclose the data source

### DIFF
--- a/Detections/MultipleDataSources/ZincJan272021IOCs.yaml
+++ b/Detections/MultipleDataSources/ZincJan272021IOCs.yaml
@@ -50,21 +50,24 @@ query: |
   | parse Message with * '(' DNSName ')' * 
   | where isnotempty(FileHash)
   | where FileHash in~ (SHA256Hash) or DNSName in~ (DomainNames)
-  | extend Account = SourceUserID, Computer = DeviceName, IPAddress = SourceIP
+  | extend DataType = "CommonSecurityLog", Account = SourceUserID, Computer = DeviceName, IPAddress = SourceIP
+  | project DataType, TimeGenerated, Computer, Account, IPAddress, FileHash, DNSName
   ),
   (DnsEvents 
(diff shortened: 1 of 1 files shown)


Detections/MultipleDataSources/ZincJan272021IOCs.yaml
```diff
@@ -50,21 +50,24 @@ query: |
   | parse Message with * '(' DNSName ')' * 
   | where isnotempty(FileHash)
   | where FileHash in~ (SHA256Hash) or DNSName in~ (DomainNames)
-  | extend Account = SourceUserID, Computer = DeviceName, IPAddress = SourceIP
+  | extend DataType = "CommonSecurityLog", Account = SourceUserID, Computer = DeviceName, IPAddress = SourceIP
```
shainw, line 53: You can simply use Type here, it is a field available on all data types and provides you with the name of it without having to manually map it.

Detections/MultipleDataSources/ZincJan272021IOCs.yaml
```diff
@@ -76,33 +79,40 @@ query: |
   | where isnotempty(Hashes)
   | parse Hashes with * 'SHA256=' SHA256 ',' * 
   | where SHA256 in~ (SHA256Hash) 
-  | extend Account = UserName
+  | extend DataType = "Microsoft-Windows-Sysmon", Account = UserName, FileHash = Hashes
```
shainw, line 82: This one you can use the Source and map it to Type.  I would recommend that it is clear what DataType it comes from, so a concat is probably best -  | extend Type = strcat(Type, ": ", Source)

shainw (changes requested):
Minor changes

shainw (approved):

PR #1693, opened 2021-02-01
Invalid grep command

cliff-wakefield-acurus:
The grep regex used contains a backslash which escapes the pipe symbol, the pipe symbol is supposed to be a logical OR between the keywords CEF and ASA.

This regex causes zero matches to be found.

Fixes #
Remove the offending backslash so the regex works as intended.


shainw:
Very old PR, closing

cliff-wakefield-acurus:
Shame this has been closed, the bug is still in the code, I submitted a quick simple fix and then the project team just ignored it.

PR #1717, opened 2021-02-04
Changes to the Query based on changes in the Log schema and Cx feedback.

aprakash13:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/MultipleDataSources/MalformedUserAgents.yaml
+++ b/Detections/MultipleDataSources/MalformedUserAgents.yaml
@@ -11,6 +11,9 @@ requiredDataConnectors:
   - connectorId: Office365
     dataTypes:
       - OfficeActivity
+  - connectorId: AzureActiveDirectory
+    dataTypes:
+      - SigninLogs
 queryFrequency: 1d
 queryPeriod: 1d
 triggerOperator: gt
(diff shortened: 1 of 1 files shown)


Detections/MultipleDataSources/MalformedUserAgents.yaml
```diff
@@ -30,13 +33,23 @@ query: |
...
+  ),
   (AzureDiagnostics
   | where TimeGenerated >= ago(endtime)
   | where ResourceType =~ "APPLICATIONGATEWAYS" 
   | where OperationName =~ "ApplicationGatewayAccess" 
-  | extend ClientIP = columnifexists("clientIP_s", "None"), UserAgent = columnifexists("userAgent_s", "None")))
+  | extend ClientIP = columnifexists("clientIP_s", "None"), UserAgent = columnifexists("userAgent_s", "None")
+  ),
+  (SigninLogs
+  | where TimeGenerated >= ago(endtime)
```
shainw, line 50: There are cases with SigninLogs events don't have UserAgent popluated, we should exclude rows right away where UserAgent is empty.

Detections/MultipleDataSources/MalformedUserAgents.yaml
```diff
@@ -30,13 +33,23 @@ query: |
   (OfficeActivity | where TimeGenerated >= ago(endtime) | where UserAgent != ""),
```
shainw, line 33: Since we are doing a union between several tables, for all tables in the query, we should consider projecting only the most desired columns and we should also consider mapping similar column names to the same name to reduce the number of columns in the result.

Detections/MultipleDataSources/MalformedUserAgents.yaml
```diff
@@ -30,13 +33,23 @@ query: |
...
   | where ResourceType =~ "APPLICATIONGATEWAYS" 
   | where OperationName =~ "ApplicationGatewayAccess" 
-  | extend ClientIP = columnifexists("clientIP_s", "None"), UserAgent = columnifexists("userAgent_s", "None")))
+  | extend ClientIP = columnifexists("clientIP_s", "None"), UserAgent = columnifexists("userAgent_s", "None")
+  ),
+  (SigninLogs
+  | where TimeGenerated >= ago(endtime)
+  | extend ClientIP = IPAddress )
+  )
   // Likely artefact of hardcoding
```
shainw, line 53: Look at adding in AADNonInteractiveUserSignInLogs and it might make sense to join TI logs based on the UserAgent so we have a row that indicates a TI connection where applicable.

shainw (changes requested):

Detections/MultipleDataSources/MalformedUserAgents.yaml
```diff
@@ -30,13 +33,23 @@ query: |
   (OfficeActivity | where TimeGenerated >= ago(endtime) | where UserAgent != ""),
   (OfficeActivity
   | where TimeGenerated >= ago(endtime)
-  | where RecordType in ("AzureActiveDirectoryAccountLogon", "AzureActiveDirectoryStsLogon")
-  | extend OperationName = Operation, UserAgent = extractjson("$[0].Value", ExtendedProperties, typeof(string))),
+  | where RecordType in ("AzureActiveDirectory", "AzureActiveDirectoryStsLogon")
+  | extend OperationName = Operation
```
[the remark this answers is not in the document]
aprakash13: Done. Thanks Shain

Detections/MultipleDataSources/MalformedUserAgents.yaml
aprakash13: Done now. Thanks Shain.

Detections/MultipleDataSources/MalformedUserAgents.yaml
aprakash13: Done. Thanks Shain

Detections/MultipleDataSources/MalformedUserAgents.yaml
aprakash13: Added AADNonInteractiveUserSignInLogs. Adding TI Logs with join was getting interesting results. Will probably investigate in later iteration.

aprakash13: Thanks Shain.

shainw (approved):

PR #1723, opened 2021-02-04
updated empty connector, moved Teams queries into OfficeActivity, upd…

shainw:
…ated some entity mappings

Fixes #

## Proposed Changes

  -
  -
  -


Hunting Queries/OfficeActivity/NewBotAdded.yaml
```diff
@@ -0,0 +1,35 @@
...
+query: |
+
+  // If you have more than 14 days worth of Teams data change this value
+  let data_date = 14d;
+  let historical_bots = (
+  OfficeActivity
+  | where TimeGenerated > ago(data_date)
+  | where OfficeWorkload =~ "MicrosoftTeams"
+  | where isnotempty(AddonName)
+  | project AddonName);
```
ashwin-patil, line 24: project can be changed to distinct to have unique list to use with in operator later.

ashwin-patil (approved):
Looks good. just one minor suggestion on hunting query.

Hunting Queries/OfficeActivity/NewBotAdded.yaml
shainw: Yeah, not making any substantive changes unless required by DataType change.  Will leave for now and we can change later if needed.

PR #1729, opened 2021-02-06
InfobloxCloudDataConnector

sschuur:
## Proposed Changes

Initial files for the Infoblox Cloud Data Connector data connector. 
  - Added data connector template json file
  - Added sample data csv file
  - Note: Logo svg already exists from previous NIOS data connector


v-admahe:
@sschuur : fallbackResourceIds  in the workbook should be empty. So could you please empty the fallbackResourceIds.

Thanks,
Aditi Maheshwari

v-admahe:
@sschuur : Preview images( InfobloxCDCB1TDBlack.png, InfobloxCDCB1TDWhite.png) should be same only background should be different.
Thanks,
Aditi Maheshwari

v-admahe:
@sschuur : Some of the query in the workbook are not syntactically correct. So could you please make required changes in it.
Please find the document  below in which we have mention list of queries with screenshot.


[InfoboxWorkboxQuery.zip](https://github.com/Azure/Azure-Sentinel/files/6050058/InfoboxWorkboxQuery.zip)

Thanks,
Aditi Maheshwari

v-admahe:
@sschuur : preview image file InfobloxCDCB1TDWhite.png is not available in Workbooks/Images/Preview directory.

Thanks,
Aditi Maheshwari

sschuur:
@v-admahe : I got rid of the fallbackResourceIds and reuploaded the broken white image. Would it be okay if we keep the images different? I really like having two images that show off two different cool parts of the workbook. Totally understand if it is preferred to keep things uniform though. 

Also, about the errors. Here's my issue: I see all the issues in the .odt are because the “parameters Threat Class, Feed are not set.” This only occurs when there is not a single nonempty Threat Class field in any logs in the time period selected. When “empty” is the only option when selecting the parameter, it’s as if Sentinel treated it more like “not a parameter” than “an empty parameter.” Queries with this empty field filter as expected when there is another parameter option: you can select N/A, other options, both or one or the other and it will filter completely as expected. But when the empty field (N/A)  is the only selectable option, Sentinel doesn’t seem to treat it as being an option in “All” inside the queries. This was an issue I was unable to fully get around. Many logs show the Threat Class field as empty, which in Infoblox means they are displayed as N/A, that is, not having a Threat Class. I decided that being able to filter Threat Classes was too valuable to a customer in exchange for the issue that is conditional anyway.

The problem goes away when the user simply selects the “N/A” parameter instead of “All” for the Threat Class. Since Sentinel tells the user right away that the issue is with the parameter selection, I imagine he would figure out the issue pretty quickly and just reselect the param. HOWEVER. The problem still occurs and becomes unfixable altogether when “N/A” (empty) is not a parameter option (by using isnotempty() in the parameter query) – the empty field is the issue. When this is the case and you select “All” you still get the “please set params” issue and now can’t fix it. See the photos attached. Any suggestions for getting around this? Maybe there is a super easy solution I am not seeing.

Feel free to reach out to me and I can show you the problem over video. Simply select N/A for Threat Class and the problem disappears.

image
--
image
--
image

v-maudan:
> @v-admahe : I got rid of the fallbackResourceIds and reuploaded the broken white image. Would it be okay if we keep the images different? I really like having two images that show off two different cool parts of the workbook. Totally understand if it is preferred to keep things uniform though.
> 
> Also, about the errors. Here's my issue: I see all the issues in the .odt are because the “parameters Threat Class, Feed are not set.” This only occurs when there is not a single nonempty Threat Class field in any logs in the time period selected. When “empty” is the only option when selecting the parameter, it’s as if Sentinel treated it more like “not a parameter” than “an empty parameter.” Queries with this empty field filter as expected when there is another parameter option: you can select N/A, other options, both or one or the other and it will filter completely as expected. But when the empty field (N/A) is the only selectable option, Sentinel doesn’t seem to treat it as being an option in “All” inside the queries. This was an issue I was unable to fully get around. Many logs show the Threat Class field as empty, which in Infoblox means they are displayed as N/A, that is, not having a Threat Class. I decided that being able to filter Threat Classes was too valuable to a customer in exchange for the issue that is conditional anyway.
> 
> The problem goes away when the user simply selects the “N/A” parameter instead of “All” for the Threat Class. Since Sentinel tells the user right away that the issue is with the parameter selection, I imagine he would figure out the issue pretty quickly and just reselect the param. HOWEVER. The problem still occurs and becomes unfixable altogether when “N/A” (empty) is not a parameter option (by using isnotempty() in the parameter query) – the empty field is the issue. When this is the case and you select “All” you still get the “please set params” issue and now can’t fix it. See the photos attached. Any suggestions for getting around this? Maybe there is a super easy solution I am not seeing.
> 
> Feel free to reach out to me and I can show you the problem over video. Simply select N/A for Threat Class and the problem disappears.
> 
> ## image
> ## image
> image


Hi @sschuur , 

If All option selected from the dropdown, the default value of selection will be null, producing an empty value for the variables used in the kql query. In order to fix this you need to add default value in all places for Feed and Threat Class dropdown.  
**"selectAllValue": "All",** [PFB screenshots]

Feed
Threat Class
I have updated and attached workbook file. Can you please try this and let us know if you still face any issue. 

[Updated_InfobloxCDCB1TDWorkbook.zip](https://github.com/Azure/Azure-Sentinel/files/6076280/Updated_InfobloxCDCB1TDWorkbook.zip)

sschuur:
@v-maudan  : That worked perfectly!! Thank you so much, that had been bothering the heck out of me. I reuploaded the workbook gallery json.

v-jayakal:
Hi @sschuur, one more point, you can add multiple images, you need to add them in both in black and white. As you wanted to add 2 images, just add 2 black images and 2 white images. Also, can you please resolve the merge conflicts? Thanks.

sschuur:
Hi @v-jayakal, I added the same image in black and white for simplicity. All merge conflicts have been resolved from what I can see.

v-jayakal (approved):

preetikr:
@v-jayakal - even though this is merged it shows up as review needed in the PR queue - can you see what's going on and if the merge was cleanly done? Thanks.

PR #1735, opened 2021-02-09
Fixing passed parameter

YaronFruchtmann:
Fixes #

## Proposed Changes

  -
  -
  -


itay6588:
The change request is correct, fixing passed parameter from AadObjectId to AadUserId

PR #1754, opened 2021-02-11
Sysmon 12 Parser Fixes

lostInSpaceSomewhere:
# Sysmon 12 Parser Fixes
Current parser is not working
Tested proposed changes and working in large production environment

##  Proposed Changes

- There was certain format errors with blank lines which breaks the KQL query/function
- The "extend Hashes" creates issues when extending Hashes to Hashes before creating the MV field of ParsedHashes
- Fixed projected field Hashes since the true extended field was actually ParsedHashes
- Added Original projected fields from let EventData query to the final projected fields (TimeGenerated, Computer, UserName, etc)
- Made it user friendly (Legible)

## Needs further evaluation

- Final extend for Hashes = column_ifexists("Hashes", "") because ParsedHashes is used instead
- Need to evaluate if ParsedHashes fits normalization schema based on OSSEM

### DIFF
--- a/Parsers/Sysmon/Sysmon-v12.0.txt
+++ b/Parsers/Sysmon/Sysmon-v12.0.txt
@@ -1,6 +1,6 @@
 // KQL Sysmon Event Parser
 // Last Updated Date: Sept 20,  2020
-// Sysmon Version: 12.0, Binary Version : 11.0, Schema Version: 4.40
+// Sysmon Version: 12.0, Binary Version : 11.0, Schema Version: 4.40 
 //
 // Sysmon Instructions:
 // If you want to print configuration schema definition of sysmon. Execute below command from command shell or powershell terminal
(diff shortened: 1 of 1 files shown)


shainw:
@ashwin-patil - please review the suggested changes here.

ashwin-patil:
Thank you @lostInSpaceSomewhere  for the PR. Since this parser is auto-generated via automated script written by @Cyb3rWard0g . I have raised an issue on OSSEM repo to update the template used by script to get these changes.

OSSEM Issue: https://github.com/OTRF/OSSEM/issues/99

For the evaluation section - lets start discussion on OSSEM for the best path forward.

Cyb3rWard0g:
Hello @ashwin-patil and @lostInSpaceSomewhere , I pushed a PR with the changes I made to the scripts and files I used in OSSEM to automate the creation of the KQL Sysmon Parser. 

OSSEM Project PR (ONLY FOR REFERENCE - NOTHING TO DO) : https://github.com/OTRF/OSSEM/pull/100

# How does it work?

1. We deploy a Windows VM via ARM templates
2. We download latest Sysmon Binary (https://docs.microsoft.com/en-us/sysinternals/downloads/sysmon) 
3. In a Windows box, we run `sysmon.exe -s` to get the schema and save it as an XML file. We currently save them here for future references: https://github.com/OTRF/OSSEM/tree/AzureCloudDD/resources/schemas
4. We run this Python script (https://github.com/OTRF/OSSEM/blob/AzureCloudDD/resources/tools/ossemSysmonKQLParser.py) to process the XML file and use this JINJA template (https://github.com/OTRF/OSSEM/blob/AzureCloudDD/resources/tools/templates/kql/sysmon_parser.txt) to automate the creation of the KQL Sysmon parser.
5. We save those KQL parsers in OSSEM in the following folder: https://github.com/OTRF/OSSEM/tree/AzureCloudDD/resources/parsers
6. We do 1-5 when there is a new Sysmon release and then push a PR to this repo (Azure-Sentinel) with the new parser.

# What I did based on your feedback can be seen here:
https://github.com/OTRF/OSSEM/pull/100

# Two things I added/updated to initial ideas

1. In the case of Sysmon EventID 1 as an example, I removed the `ParsedHashes` and simply replaced the original `Hashes` value with the results of running the following:

```
| extend Hashes = extract_all(@"(?P<key>\w+)=(?P<value>[a-zA-Z0-9]+)", dynamic(["key","value"]), tostring(EventDetail.[17].["#text"]))| mv-apply Hashes on (summarize Hashes = make_bag(pack(tostring(Hashes[0]), tostring(Hashes[1]))))
```

2. I added support for Sysmon Event ID 15. It also has a field named `Hash` which is the same output format as `Hashes` in events like Event ID 1, 7, etc.

https://github.com/OTRF/OSSEM/blob/AzureCloudDD/resources/tools/templates/kql/sysmon_parser.txt#L29-L33

Jinja Template handling the same way how it is handled for `Hashes`

```
{% for field in event['events'] -%}
    {% if field['name'] in ['Hashes','Hash'] -%}
        {%- raw %}| extend {% endraw %}{{field['name']}}{% raw %} = extract_all(@"(?P<key>\w+)=(?P<value>[a-zA-Z0-9]+)", dynamic(["key","value"]), tostring(EventDetail.[{% endraw %}{{field['index']}}{% raw %}].["#text"])){% endraw %}
        {%- raw %}| mv-apply {% endraw %}{{field['name']}}{% raw %} on (summarize {% endraw %}{{field['name']}}{% raw %} = make_bag(pack(tostring({% endraw %}{{field['name']}}{% raw %}[0]), tostring({% endraw %}{{field['name']}}{% raw %}[1])))){% endraw %}
{% endif %}{% endfor %}
```

# What do I share all this?

One reason is to share some context about how we handle the creation of the KQL Sysmon parser, and also to get your feedback about the results. This is the final KQL Sysmon parser that is in a PR for review for OSSEM. 

https://github.com/OTRF/OSSEM/blob/AzureCloudDD/resources/parsers/SysmonKQLParserV13.01.txt

If the parser aligns with what you had in mind, then all you would need to do is let me know here to accept PR in OSSEM project, and you just need to update your document in this PR for us to accept your contribution to this repo.

Thank you @lostInSpaceSomewhere !

shainw:
@Cyb3rWard0g and @ashwin-patil to review and join together existing PR improvements for Sysmon Parser.

oshezaf:
@Cyb3rWard0g, @ashwin-patil, @shainw : you might want to look into moving to use the [ASIM Sysmon parsers](https://aka.ms/ASimSysmonForWindows). They use parse instead of parse_xml which is way more performant, and also try to support multiple Sysmon schema versions, including the new one introduced last month. It does require maintenance as well, but we would love to collaborate on those more "standard" aspiring ones.

shainw:
"Sysmon will be supported by ASIM in the near future to align on the ease of parsing by a singular method via normalization"

PR #1757, opened 2021-02-12
Add 'credentials added after admin consent' query

jannieli:
Fixes #

## Proposed Changes

  - Add 'credentials added after admin consent' query to AuditLog Detection queries

### DIFF
--- a/Detections/AuditLogs/CredentialAddedAfterAdminConsent.yaml
+++ b/Detections/AuditLogs/CredentialAddedAfterAdminConsent.yaml
@@ -1,0 +1,67 @@
+id: 707494a5-8e44-486b-90f8-155d1797a8eb
+name: Credential added after admin consented to Application
+description: |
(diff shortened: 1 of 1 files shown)


Detections/AuditLogs/CredentialAddedAfterAdminConsent.yaml
```diff
@@ -0,0 +1,67 @@
+id: 707494a5-8e44-486b-90f8-155d1797a8eb
+name: Credential added after admin consented to Application
+description: |
+  'This query will identify instances where Service Principal credentials were added to an application by a UserA after the application was granted admin consent rights by some UserB.
```
shainw, line 4: We should change the part that says "UserA" and "UserB" to the following: "..were added to an application by one user after the application was granted admin consent rights by another user".

shainw (changes requested):
Minor change in Description, otherwise looks good.

shainw (approved):

PR #1759, opened 2021-02-12
Sample data validation

v-admahe:
Fixes #

## Proposed Changes
1. Sample data  file should start with an array.
  -
  -
  -


v-rucdu:
Entry to the azure-pipeline file is missing... how will this be executed as part of validation?

v-rucdu (changes requested):
Entry to the azure-pipeline file is missing... how will this be executed as part of validation without it?

PR #1765, opened 2021-02-14
Screenshots and unsigned files

YaronFruchtmann:
Fixes #

## Proposed Changes

  -
  -
  -


shainw (approved):
Any reason that we are not putting the Insight and Activity together?  They use the same base query.

PR #1773, opened 2021-02-16
Identifies connections to Solorigate-related DNS records

yoweiz:
Identifies connections to Solorigate-related DNS records based on VM insights data.
Based on: https://github.com/Azure/Azure-Sentinel/blob/master/Detections/MultipleDataSources/Solorigate-Network-Beacon.yaml

### DIFF
--- a/Detections/MultipleDataSources/Solorigate-VM-Network.yaml
+++ b/Detections/MultipleDataSources/Solorigate-VM-Network.yaml
@@ -26,7 +26,7 @@ tags:
 query:  |  
 
   let domains = dynamic(["incomeupdate.com","zupertech.com","databasegalore.com","panhardware.com","avsvmcloud.com","digitalcollege.org","freescanonline.com","deftsecurity.com","thedoccloud.com","virtualdataserver.com","lcomputers.com","webcodez.com","globalnetworkissues.com","kubecloud.com","seobundlekit.com","solartrackingsystem.net","virtualwebdata.com"]);
-  let timeframe = 1d;
+  let timeframe = 1h;
   let connections = VMConnection 
       | where TimeGenerated >= ago(timeframe)
       | extend DNSName = set_union(todynamic(RemoteDnsCanonicalNames),todynamic(RemoteDnsQuestions))
--- a/Detections/MultipleDataSources/Solorigate-VM-Network.yaml
+++ b/Detections/MultipleDataSources/Solorigate-VM-Network.yaml
@@ -1,0 +1,77 @@
+id: ab4b6944-a20d-42ab-8b63-238426525801
+name: Solorigate Network data
+description: | 
+  'Identifies connections to Solorigate-related DNS records based on VM insights data'
+severity: High 
+requiredDataConnectors: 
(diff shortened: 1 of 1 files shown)


Detections/MultipleDataSources/Solorigate-VM-Network.yaml
```diff
@@ -0,0 +1,77 @@
...
+tactics: 
+  - CommandAndControl 
+relevantTechniques:
+  - T1102
+tags:
+  - Solorigate
+query:  |  
+
+  let domains = dynamic(["incomeupdate.com","zupertech.com","databasegalore.com","panhardware.com","avsvmcloud.com","digitalcollege.org","freescanonline.com","deftsecurity.com","thedoccloud.com","virtualdataserver.com","lcomputers.com","webcodez.com","globalnetworkissues.com","kubecloud.com","seobundlekit.com","solartrackingsystem.net","virtualwebdata.com"]);
+  let timeframe = 1d;
```
YaronFruchtmann, line 29: Should be 1h

YaronFruchtmann (approved):
This is fine. connection table is expected to be small, so it is, as it should be, the first in join operator. I am approving but @shainw knows Solarigate better, and I prefer  he has a look before merge

laithhisham (commented):
Looks good :)
I also recommend waiting for Shain's approval as Yaron said

shainw (changes requested):
I don't see why we would need a separate query on this, I think we should just update this - https://github.com/Azure/Azure-Sentinel/blob/master/Detections/MultipleDataSources/Solorigate-Network-Beacon.yaml

yoweiz:
> I don't see why we would need a separate query on this, I think we should just update this - https://github.com/Azure/Azure-Sentinel/blob/master/Detections/MultipleDataSources/Solorigate-Network-Beacon.yaml

I understood from @tamersalman that you both agreed to write a separated detection. Please correct me if I'm wrong. We currently have a limitation on the number of entities, so we won't be able to add file and process entities to the previous detection.

shainw:
> > I don't see why we would need a separate query on this, I think we should just update this - https://github.com/Azure/Azure-Sentinel/blob/master/Detections/MultipleDataSources/Solorigate-Network-Beacon.yaml
> 
> I understood from @tamersalman that you both agreed to write a separated detection. Please correct me if I'm wrong. We currently have a limitation on the number of entities, so we won't be able to add file and process entities to the previous detection.

Got it, lets make sure the name is clear then.  Commented PR again for name.

Detections/MultipleDataSources/Solorigate-VM-Network.yaml
```diff
@@ -0,0 +1,77 @@
+id: ab4b6944-a20d-42ab-8b63-238426525801
+name: Solorigate Network data
```
shainw, line 2: Since we will have 2 of these very similar but against different data sets we should be clear in the name.  I suggest something like - "Solorigate Domains Found in VM Insights"

shainw (changes requested):

shainw (approved):

PR #1775, opened 2021-02-16
Window Defender AV activities and insights

YaronFruchtmann:
Fixes #

## Proposed Changes

  -
  -
  -


shainw (approved):
Any reason to not have the Activity and Insight together since they share the same base query?

PR #1799, opened 2021-02-19
McAfeeePO Parser and Rules

vu-socprime:
## Proposed Changes
  - Add McAfeeePO Parser
  - Add McAfeeePO Detection Rules and Hunting Queries


preetikr:
@vu-socprime - Please fix the two errors so that the review can happen - need the tests to be 100% pass. Thanks.

vu-socprime:
> @vu-socprime - Please fix the two errors so that the review can happen - need the tests to be 100% pass. Thanks.

@preetikr - There only one error type left :
Template Id:'155243f4-d962-4717-8a7b-b15b6d112660' doesn't have valid connectorIds:'McAfeeePO'. If a new connector is used and already configured in the Portal, please add it's Id to the list in 'ValidConnectorIds.json' file.

Can I just update file .script/tests/detectionTemplateSchemaValidation/ValidConnectorIds.json to pass tests? Thanks.

shainw (changes requested):
**Detections, Hunting and Parsers feedback**
Detections:
For all Detections, you do not need to specify a TimeGenerated filter if you are using the same period as the queryPeriod value.  Please remove this from the query body.

In general, we don't recommend queries shorter than 1h so that customers don't have a negative impact by default from a perf and alert fatigue perspective.  You can leave as is, but we recommend to go with 1h.

Hunting:
Be sure that you include all the required YAML fields - mainly you are missing the entity mappings, but you should confirm all required - https://github.com/Azure/Azure-Sentinel/wiki/Query-Style-Guide

Parsers:
Be sure to include all the fields in the YAML, mainly you are missing the detailed comments - https://github.com/Azure/Azure-Sentinel/wiki/Contribute-to-Sentinel-GitHub-Community-of-Queries#example-parser

@preetikr can answer more fully on the data connector, but if there is no built in data connector, then you will want to use an empty field for requiredDataConnectors - 

```
requiredDataConnectors: []
```

vu-socprime:
Rules and queries were updated according to recommendations in previous comment.
@preetikr - what is better option to pass tests - use empty "requiredDataConnectors" field or update "ValidConnectorIds.json" file?

preetikr:
@vu-socprime and @shainw - McAfee and Oracle DB etc. have data connectors and these analytic rules apply to those data connectors. The only thing is we are delivering the data connectors, active analytic rules etc. as a packaged Solution. In this scenario please provide requiredDataConnectors as the data connector ID value. For e.g. for McAfee ePo it's McAfeeePO and data type as McAfeeEPOEvent - that way when we start packaging analytic rule templates this will already be set fine. Thanks.

vu-socprime:
> requiredDataConnectors as the data connector ID value

@preetikr Field requiredDataConnectors has correct data connector ID value, so validation can be considered as successful.

preetikr:
Thanks, @vu-socprime for making the changes. Yes, the data connector ID is correct so the test validation error is a false positive. @shainw, please go through the incorporated feedback and please review/approve if this looks good to you. Thanks.

Hunting Queries/McAfeeePO/McAfeeEPOMultipleThreats.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - InitialAccess
+query: |
+  let lbtime = 24h;
+  let threshold = 1;
+  McAfeeEPOEvent
+  | where TimeGenerated > ago(lbtime)
+  | where isnotempty(ThreatName)
+  | where ThreatName != '_'
+  | summarize th_cnt = makeset(ThreatName) by DvcIpAddr
+  | where array_len(ThreatList) > threshold
```
shainw, line 19: You have no `ThreatList` field coming thru after the summarize here.  I assume you meant to use `th_cnt`

shainw (changes requested):
- All detections and hunting queries require the relevantTechniques field - https://github.com/Azure/Azure-Sentinel/wiki/Query-Style-Guide#relevanttechniques

- Is there a plan to include additional event ids?  If not any query that has EventId in ('singleEvent#') should be changed to EventId == singleEvent#

- other comments inline

Hunting Queries/McAfeeePO/McAfeeEPOMultipleThreats.yaml
vu-socprime: fixed

vu-socprime:
> * All detections and hunting queries require the relevantTechniques field - https://github.com/Azure/Azure-Sentinel/wiki/Query-Style-Guide#relevanttechniques
> * Is there a plan to include additional event ids?  If not any query that has EventId in ('singleEvent#') should be changed to EventId == singleEvent#
> * other comments inline

@shainw fixed

shainw (approved):

PR #1803, opened 2021-02-19
update missing connectors and missing techniques for CiscoUmbrella

ashwin-patil:
Fixes #

## Proposed Changes

  - Updated missing requiredconnectors as empty list
  - Updated missing MITRE techniques
  -


shainw (changes requested):
Since you are in here, can you add in the entity mappings field also

shainw (approved):

PR #1813, opened 2021-02-22
CEF get info script

NoamLandress:
A new script that runs on all CEF supported OS.
It gathers both system and connector installation data, like free disk space and log files, for ticket investigation.
The script outputs it's data to the file - '/tmp/cef_get_info' which CSS will later attach to the ticket.
Every run of the script deletes the output file and the previous version of tis script in order to avoid taking up disk space


DataConnectors/CEF/cef_gather_info.py
```diff
@@ -0,0 +1,166 @@
...
+
+class SystemInfo:
+    def __init__(self, command, command_output="No output received"):
+        self.command = command
+        self.command_output = command_output
+
+    def __repr__(self):
+        delimiter = "\n-----------------------------------\n"
+        return str(
+            delimiter + "command: " + self.command + '\n' + "output: " + self.command_output + delimiter)
```
liemilyg, line 36: move the escping to here

DataConnectors/CEF/cef_gather_info.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+script_version = '1.0'
+output_file_path = '/tmp/cef_get_info'
+
+basic_command_dict = {
```
liemilyg, line 42: why not creating a dictionary here with one value and avoid using 2 dictionaries/running function

DataConnectors/CEF/cef_gather_info.py
```diff
@@ -0,0 +1,166 @@
...
+def append_content_to_file(command_object, file_path=output_file_path):
+    """
+
+    :param command_object: consists of the name and the output
+    :param file_path: a file to share the commands outputs
+    """
+    output = repr(command_object).replace('%', '%%')
+    command_tokens = ["sudo", "bash", "-c", "printf '" + "\n" + output + "' >> " + file_path]
+    try:
+        write_new_content = subprocess.Popen(command_tokens, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
liemilyg, line 94: use run_command?

DataConnectors/CEF/cef_gather_info.py
```diff
@@ -0,0 +1,166 @@
...
+        write_new_content.communicate()
+    except Exception:
+        print(str(command_object.command) + "was not documented successfully")
+
+
+def run_command(command):
```
liemilyg, line 100: i think u can combine run_command and run_special_command by using array and run the second command if the array contains more than 1 value

DataConnectors/CEF/cef_gather_info.py
```diff
@@ -0,0 +1,166 @@
...
+    :param command: the key value pair from the command dict
+    :return: a command object consisting of a name and an output
+    """
+    command_to_run = subprocess.Popen(basic_command_dict[command].split(' '), stdout=subprocess.PIPE,
+                                      stderr=subprocess.STDOUT)
+    try:
+        o, e = command_to_run.communicate()
+    except Exception:
+        print(basic_command_dict[command] + "failed to run")
+    o = o.decode(encoding='UTF-8')
```
liemilyg, line 112: can be part of the reper?

DataConnectors/CEF/cef_gather_info.py
```diff
@@ -0,0 +1,166 @@
...
+    except Exception:
+        print("Clean up command failed to run")
+
+
+def main():
+    print_notice("Note this script should be run in elevated privileges")
+    print("Beginning to collect server data- This may take a few minutes")
+    clean_up()
+    for command in basic_command_dict.keys():
+        command_object = run_command(command)
```
liemilyg, line 153: create a function that wrap these 2 :)

DataConnectors/CEF/cef_gather_info.py
```diff
@@ -0,0 +1,166 @@
...
+        o, e = clean_up_command.communicate()
+    except Exception:
+        print("Clean up command failed to run")
+
+
+def main():
+    print_notice("Note this script should be run in elevated privileges")
+    print("Beginning to collect server data- This may take a few minutes")
+    clean_up()
+    for command in basic_command_dict.keys():
```
orgadhadas, line 152: I'd suggest to use tqdm here, but this will require library installation, so maybe consider writing at the beginning how many commands there are?

DataConnectors/CEF/cef_gather_info.py
orgadhadas, line 94: subprocess is asynchronous and then you have to use wait, consider using regular python file autput (with open(filename, 'a') as f: f.write(output)) - this is a synchronous operation, you will not need the wait and it's safer (in the theoretic case that writing will take even more than 1 second for some reason).

DataConnectors/CEF/cef_gather_info.py
```diff
@@ -0,0 +1,166 @@
...
+        command_object = run_command(command)
+        append_content_to_file(command_object)
+        time.sleep(1)
+
+    for command in advanced_command_dict.keys():
+        command_object = run_special_command(command)
+        append_content_to_file(command_object)
+        time.sleep(1)
+    print_notice(
+        "Data collection complete. Please provide CSS with the content of the file {}".format(output_file_path))
```
orgadhadas, line 162: complete**d**?

DataConnectors/CEF/cef_gather_info.py
NoamLandress: I changed the way I write to a file, leveraging pythons built in file handling instead of Linux commands (according to Hadas's suggestion).

liemilyg (approved):

DataConnectors/CEF/cef_gather_info.py
NoamLandress: Will add in Future versions.
Currently runs instantly so no need for that.

DataConnectors/CEF/cef_gather_info.py
```diff
@@ -0,0 +1,152 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# ----------------------------------------------------------------------------
+# This script is used to install CEF agent on a linux machine an configure the
+# syslog daemon on the linux machine.
```
ofshomro, line 5: Update

DataConnectors/CEF/cef_gather_info.py
```diff
@@ -0,0 +1,152 @@
...
+        cef_get_info_file.write(output)
+    except Exception:
+        print(str(command_object.command) + "was not documented successfully")
+    cef_get_info_file.close()
+
+
+def run_command(command):
+    """
+
+    :param command: the key value pair from the command dict
```
ofshomro, line 99: Isn't it only the key?

DataConnectors/CEF/cef_gather_info.py
```diff
@@ -0,0 +1,152 @@
...
+    # for commands that consist of a single pipe
+    else:
+        first_command = subprocess.Popen(command_dict[command][0].split(' '), stdout=subprocess.PIPE,
+                                         stderr=subprocess.STDOUT)
+        command_to_run = subprocess.Popen(command_dict[command][1].split(' '), stdin=first_command.stdout,
+                                          stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    try:
+        o, e = command_to_run.communicate()
+    except Exception:
+        pass
```
ofshomro, line 114: We don't want exceptions output?

DataConnectors/CEF/cef_gather_info.py
NoamLandress: Correct.

DataConnectors/CEF/cef_gather_info.py
NoamLandress: Nope. I don't want customers to deal with exception outputting from with this script. If a command output comes with a fault or empty output that's OK and we can investigate it on our side. The script should just finish running successfully and we will get whatever output is possible to the file.

liemilyg (approved):

PR #1814, opened 2021-02-22
fix the time generated issue

Amitbergman:
Fixes #

## Proposed Changes

  -
  -
  -


shainw (approved):
Caught some misses and fixed those, all good now.

PR #1815, opened 2021-02-22
Ignite 2021 Queries

thmcelro:
Covers queries that are part of the Ignite 2021 demo of Azure Sentinel.

### Detection
**Created AdditionalFilesUploadedByActor.yaml**: When an Azure Defender alert for a known-bad file hash triggers, the detection will collate other files uploaded by that same IP address to detect other risky files.

### Hunting
**Created AzureAdministrationFromVPS.yaml:**: Finds instances where Azure Admin Activity takes place from a known VPS provider.

**Created AzureStorageFileCreateAccessDelete.yaml:**: Finds instances where a file is created in Azure Storage, accessed by a single user, and then deleted.

**Created AzureStorageFileCreatedQuicklyDeleted.yaml:** Finds instances where a file is created and then deleted within a given threshold (by default 5min).

**Created AzureStorageFileOnEndpoint.yaml:**: Finds instances where a file in Azure Storage is present on an Endpoint system based on data from Microsoft 365 Defender Advanced Hunting data.

**Created AzureStorageMassDeletion.yaml:**  Finds instances where a mass-deletion event occurs from a single IP address. Configurable thresholds.

**Created AzureStorageUploadFromVPS.yaml:**  Finds instances where a VPS provider IP is used to upload files to Azure Storage.

**Created AzureStorageUploadLinkAccount.yaml:**  Links user account to Azure storage upload action if the upload actions takes place through Azure Portal.

### DIFF
--- a/Detections/MultipleDataSources/AdditionalFilesUploadedByActor.yaml
+++ b/Detections/MultipleDataSources/AdditionalFilesUploadedByActor.yaml
@@ -7,8 +7,8 @@ requiredDataConnectors:
   - connectorId: MicrosoftCloudAppSecurity
     dataTypes:
       - SecurityAlert
-queryFrequency: 15m
-queryPeriod: 30m
+queryFrequency: 1d
+queryPeriod: 1d
 triggerOperator: gt
 triggerThreshold: 0
 tactics:
@@ -23,7 +23,6 @@ query: |
 
   //Collect the alert events
   let alertData = SecurityAlert 
-  | where TimeGenerated > ago(30m) 
   | where DisplayName has "Potential malware uploaded to" 
   | extend Entities = parse_json(Entities) 
   | mv-expand Entities;
@@ -45,7 +44,6 @@ query: |
     union
     StorageFileLogs, 
     StorageBlobLogs 
-    | where TimeGenerated > ago(30m)
     //File upload operations 
     | where OperationName =~ "PutBlob" or OperationName =~ "PutRange"
     //Parse out the uploader IP 
@@ -67,8 +65,7 @@ query: |
         | join kind=leftouter (
           union
           StorageFileLogs,
-          StorageBlobLogs         
-          | where TimeGenerated > ago(30m)         
+          StorageBlobLogs               
           | where OperationName =~ "DeleteFile" or OperationName =~ "DeleteBlob"         
           | extend ClientIP = tostring(split(CallerIpAddress, ":", 0)[0])         
           | extend FileName = extract(@"\/([\w\-. ]+)\?", 1, Uri)         
--- a/Detections/MultipleDataSources/AdditionalFilesUploadedByActor.yaml
+++ b/Detections/MultipleDataSources/AdditionalFilesUploadedByActor.yaml
@@ -7,8 +7,8 @@ requiredDataConnectors:
   - connectorId: MicrosoftCloudAppSecurity
     dataTypes:
       - SecurityAlert
-queryFrequency: 15min
-queryPeriod: 30min
+queryFrequency: 15m
+queryPeriod: 30m
 triggerOperator: gt
 triggerThreshold: 0
 tactics:
@@ -23,7 +23,7 @@ query: |
 
   //Collect the alert events
   let alertData = SecurityAlert 
-  | where TimeGenerated > ago(30min) 
+  | where TimeGenerated > ago(30m) 
   | where DisplayName has "Potential malware uploaded to" 
   | extend Entities = parse_json(Entities) 
   | mv-expand Entities;
@@ -45,7 +45,7 @@ query: |
     union
     StorageFileLogs, 
     StorageBlobLogs 
-    | where TimeGenerated > ago(30min)
+    | where TimeGenerated > ago(30m)
     //File upload operations 
     | where OperationName =~ "PutBlob" or OperationName =~ "PutRange"
     //Parse out the uploader IP 
@@ -68,7 +68,7 @@ query: |
           union
           StorageFileLogs,
           StorageBlobLogs         
-          | where TimeGenerated > ago(30min)         
+          | where TimeGenerated > ago(30m)         
           | where OperationName =~ "DeleteFile" or OperationName =~ "DeleteBlob"         
           | extend ClientIP = tostring(split(CallerIpAddress, ":", 0)[0])         
           | extend FileName = extract(@"\/([\w\-. ]+)\?", 1, Uri)         
--- a/Detections/MultipleDataSources/AdditionalFilesUploadedByActor.yaml
+++ b/Detections/MultipleDataSources/AdditionalFilesUploadedByActor.yaml
@@ -1,0 +1,95 @@
+id: b9e3b9f8-a406-4151-9891-e5ff1ddd8c1d
+name: Linked Malicious Storage Artifacts
+description: |
+  'An IP address which uploaded malicious content to an Azure Blob or File Storage container (triggering a malware alert) also uploaded additional files.'
+severity: Medium
+requiredDataConnectors:
+  - connectorId: MicrosoftCloudAppSecurity
+    dataTypes:
+      - SecurityAlert
+queryFrequency: 15min
+queryPeriod: 30min
+triggerOperator: gt
+triggerThreshold: 0
+tactics:
+  - CommandAndControl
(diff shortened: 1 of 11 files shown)


Hunting Queries/AzureStorage/AzureStorageFileCreateAccessDelete.yaml
```diff
@@ -0,0 +1,57 @@
...
+     //Again, parse the URI to remove the parameters as they change per request 
+     | extend Uri = tostring(split(Uri, "?", 0)[0])
+     //Parse the caller IP as it contains the port 
+     | extend CallerIpAddress = tostring(split(CallerIpAddress, ":", 0)[0])
+     //Summarise the download events by the URI, we are only looking for instances where a single caller IP downloaded the file,
+     //so we can safely use any() on the IP. 
+     | summarize Downloads=count(), DownloadTimeStart=max(TimeGenerated), DownloadTimeEnd=min(TimeGenerated), DownloadIP=any(CallerIpAddress), DownloadUserAgents=make_set(UserAgentHeader), dcount(CallerIpAddress) by Uri 
+     | where dcount_CallerIpAddress == 1
+     ) on Uri 
+ | project UploadedTime, DeletedTime, OperationName, OperationName1, Uri, UploaderAccountName, UploaderIP, UploaderUA, DownloadTimeStart, DownloadTimeEnd, DownloadIP, DownloadUserAgents, DeletionIP, DeletionUA, ResponseMd5
```
shainw, line 57: We need to start adding in the entity mappings for all hunting queries.  I think in this case you only have the IP entity available.  Can you include that?

Hunting Queries/AzureStorage/AzureStorageFileCreatedQuicklyDeleted.yaml
```diff
@@ -0,0 +1,38 @@
...
+      StorageData
+      | where TimeGenerated > ago(timerange)
+      | where StatusText =~ "Success"
+      | where OperationName =~ "DeleteBlob" or OperationName =~ "DeleteFile"
+      | extend Uri = tostring(split(Uri, "?", 0)[0])
+      | project OperationName, DeletedTime=TimeGenerated, Uri
+  ) on Uri
+  | project TimeGenerated, DeletedTime, OperationName, OperationName1, Uri, CallerIpAddress, UserAgentHeader, ResponseMd5, AccountName
+  | extend windowEnd = TimeGenerated+5m 
+  | where DeletedTime between (TimeGenerated .. windowEnd)
```
shainw, line 38: Same here on Entity mapping, looks like IP address and Account

Hunting Queries/AzureStorage/AzureStorageFileOnEndpoint.yaml
```diff
@@ -0,0 +1,51 @@
...
+  | join (
+    DeviceFileEvents
+    | where TimeGenerated > ago(7d)
+    | where ActionType =~ "FileCreated"
+    | where isnotempty(MD5)
+    | extend p = pack("FileCreateTime", TimeGenerated, "Device", DeviceName, "DeviceId", DeviceId, "FileName", FileName, "InititatingProcess", InitiatingProcessFileName)
+    | summarize make_bag(p), dcount(DeviceName) by MD5
+  ) on $left.Md5Hash == $right.MD5
+  | project TimeGenerated, FileName, FileHashCustomEntity=Md5Hash, AccountName, SourceTable, DevicesImpacted=dcount_DeviceName, Entitites=bag_p
+entityMappings:
```
shainw, line 45: Please include Account also in the entity mapping

Hunting Queries/AzureStorage/AzureStorageMassDeletion.yaml
```diff
@@ -0,0 +1,33 @@
...
+  StorageFileLogs,
+  StorageBlobLogs
+  | where TimeGenerated > ago(3d)
+  | where StatusText =~ "Success"
+  | where OperationName =~ "DeleteBlob" or OperationName =~ "DeleteFile"
+  | extend CallerIpAddress = tostring(split(CallerIpAddress, ":", 0)[0])
+  | summarize dcount(Uri) by bin(TimeGenerated, deleteWindow), CallerIpAddress, UserAgentHeader, AccountName
+  | where dcount_Uri >= deleteThreshold
+  | project TimeGenerated, IPCustomEntity=CallerIpAddress, UserAgentHeader, FilesDeleted=dcount_Uri, AccountName
+entityMappings:
```
shainw, line 29: Also an entity mapping for AccountName

Hunting Queries/AzureStorage/AzureStorageUploadLinkAccount.yaml
```diff
@@ -0,0 +1,49 @@
...
+  ) on $left.CallerIpAddress == $right.IPAddress
+  //Look back in the signinlogs for the most recent login
+  | where AzureLoginTime between (WindowStart .. FileUploadTime)
+  | project AccountUsed=UserPrincipalName, AzureLoginTime, OperationName, FileUploadPath=Uri, CallerIpAddress, LoginUserAgent, UploadUserAgent=UserAgentHeader
+  //Optional user agent check
+  | where LoginUserAgent =~ UploadUserAgent
+  //Pack and summarise the matching login events by the upload event
+  | extend p = pack("AccountUsed", AccountUsed, "AzureLoginTime", AzureLoginTime, "UserAgent", LoginUserAgent)
+  | summarize LoginEvents=make_bag(p) by FileUploadPath, OperationName, UploadUserAgent
+
```
shainw, line 49: Thoughts on bringing CallerIPAddress thru?  Looks like you drop it in the pack and summarize.

Detections/MultipleDataSources/AdditionalFilesUploadedByActor.yaml
```diff
@@ -0,0 +1,95 @@
+id: b9e3b9f8-a406-4151-9891-e5ff1ddd8c1d
+name: Linked Malicious Storage Artifacts
+description: |
+  'An IP address which uploaded malicious content to an Azure Blob or File Storage container (triggering a malware alert) also uploaded additional files.'
+severity: Medium
+requiredDataConnectors:
+  - connectorId: MicrosoftCloudAppSecurity
+    dataTypes:
+      - SecurityAlert
+queryFrequency: 15m
```
shainw, line 10: Are we sure we want to run this every 15 minutes?  Also, seems like you would not want to run every 15 mins and look back at 30 mins, otherwise I think this would cause duplicates.  Does it make sense to move this to daily and then bin on 30 mins?

Detections/MultipleDataSources/AdditionalFilesUploadedByActor.yaml
```diff
@@ -0,0 +1,95 @@
...
+  - ResourceDevelopment
+relevantTechniques:
+  - T1584
+  - T1071
+  - T1567
+query: |
+
+  //Collect the alert events
+  let alertData = SecurityAlert 
+  | where TimeGenerated > ago(30m)
```
shainw, line 26: Queries are moving away from hard coding the lookback unless it is required based on different lookback timeframes.  In this case, we use the same 30mins.  This should be removed and the query will be controlled by the queryPeriod value in the YAML.

shainw (changes requested):

Detections/MultipleDataSources/AdditionalFilesUploadedByActor.yaml
thmcelro: We do not need to do that at all, this was a remnant of the test environment that I didn't change. I think we can run daily, and I don't think we need to bin, as it's triggered based on an alert, any upload activity from the actor IP in a day seems OK to me?

Hunting Queries/AzureStorage/AzureStorageFileCreatedQuicklyDeleted.yaml
thmcelro: Just IP I think, the account is the StorageAccount "name".

Hunting Queries/AzureStorage/AzureStorageFileCreateAccessDelete.yaml
thmcelro: Done

shainw (approved):

PR #1827, opened 2021-02-25
Gain Code Execution to ADFS via WMI query

jannieli:
Fixes #

## Proposed Changes

  - Add query for 'Gain Code Execution on ADFS Server via Remote WMI Execution'

### DIFF
--- a/Detections/MultipleDataSources/GainCodeExecutionADFSviaWMI.yaml
+++ b/Detections/MultipleDataSources/GainCodeExecutionADFSviaWMI.yaml
@@ -40,45 +40,49 @@ query: |
   | evaluate pivot(Key, any(Value), TimeGenerated, Source, EventLog, Computer, EventLevel, EventLevelName, EventID, UserName, RenderedDescription, MG, ManagementGroupName, Type, _ResourceId)
   | extend process = split(Image, '\\', -1)[-1]
   | where process =~ "Microsoft.IdentityServer.ServiceHost.exe"
-  | summarize by Computer);
+  | distinct Computer
+  | union (
+  SecurityEvent
+  | where TimeGenerated > ago(timeframe+lookback)
+  | where EventID == 4688 and SubjectLogonId != "0x3e4"
+  | where ProcessName has "Microsoft.IdentityServer.ServiceHost.exe"
+  | distinct Computer
+  )
+  | distinct Computer);
   (union isfuzzy=true
   (
   SecurityEvent
   | where TimeGenerated > ago(timeframe)
-  | where EventID == 4624 and LogonType == 3 and Account !endswith "$" 
-  | join kind=inner 
-  ( SecurityEvent
-  | where TimeGenerated > ago(timeframe)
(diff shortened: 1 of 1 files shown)


Detections/MultipleDataSources/GainCodeExecutionADFSviaWMI.yaml
```diff
@@ -0,0 +1,96 @@
...
+  - Solorigate
+query: |
+  let timeframe = 1d;
+  // Adjust for a longer timeframe for identifying ADFS Servers
+  let lookback = 6d;
+  // Identify ADFS Servers
+  let ADFS_Servers = (
+  Event
+  | where TimeGenerated > ago(timeframe+lookback)
+  | where Source == "Microsoft-Windows-Sysmon"
```
shainw, line 35: Can we filter to a specific Event ID for Sysmon, such as EventID 1 Process Creation?  This will allow the query to be more efficient.

Detections/MultipleDataSources/GainCodeExecutionADFSviaWMI.yaml
```diff
@@ -0,0 +1,96 @@
...
+  ),
+  (
+  Event
+  | where TimeGenerated > ago(timeframe)
+  | where Source == "Microsoft-Windows-Sysmon"
+  | extend EventData = parse_xml(EventData).DataItem.EventData.Data
+  | mv-expand bagexpansion=array EventData
+  | evaluate bag_unpack(EventData)
+  | extend Key=tostring(['@Name']), Value=['#text']
+  | evaluate pivot(Key, any(Value), TimeGenerated, Source, EventLog, Computer, EventLevel, EventLevelName, EventID, UserName, RenderedDescription, MG, ManagementGroupName, Type, _ResourceId)
```
shainw, line 80: I don't think you want EventID in there, it is already available as a top level field.  See - https://github.com/Azure/Azure-Sentinel/blob/master/Parsers/Sysmon/Sysmon-AllVersions_Parser.txt

Detections/MultipleDataSources/GainCodeExecutionADFSviaWMI.yaml
```diff
@@ -0,0 +1,96 @@
...
+  | where Computer in~ (ADFS_Servers)
```
shainw, line 81: Computer is a top level field and so we should move to just under the Source match to improve query performance.

Detections/MultipleDataSources/GainCodeExecutionADFSviaWMI.yaml
```diff
@@ -0,0 +1,96 @@
...
+  // Check for WMI Events
+  | where EventID in (19, 20, 21)
```
shainw, line 83: EventID is a top level field, to improve performance, you can AND it with the Source match.

shainw (changes requested):
Minor changes, otherwise looks good.

shainw (approved):

PR #1830, opened 2021-02-26
Update ExplicitMFADeny.yaml

anb-coder:
The other rules are using this mapping (AccountCustomEntity = UserPrincipalName).
AlternateSignInName is not always present in the logs.

Fixes #
entities not showing up

## Proposed Changes

  - AccountCustomEntity = UserPrincipalName
  -
  -


shainw (approved):
@anb-coder - Thanks was meaning to fix this a couple weeks back. :)

Issue #1771, opened 2021-02-15
Bad entity mapping in "Multiple Password Reset by User"


javiersoriano:
**Describe the bug**
Analytic Rule "multiple password reset by user" may have issues with Host entity when a match is found in SigninLogs.

**To Reproduce**
Make this rule trigger for SigninLogs and you will find this in the entities:
![image](https://user-images.githubusercontent.com/13329585/107922368-2f913d80-6f70-11eb-8304-5882ea9bd33e.png)
 
![image](https://user-images.githubusercontent.com/13329585/107922413-446dd100-6f70-11eb-81e1-5a0bf9f25ee4.png)

This is because of the wrong entity mapping is done in line 68 of the template [here](https://github.com/Azure/Azure-Sentinel/blob/fbe8b107ba5516f2a0b3c29218cb8233d40f0500/Detections/MultipleDataSources/MultiplePasswordresetsbyUser.yaml#L68). 

and then this field (Comptuter) is converted to string and mapped to Host entity.

javiersoriano:
tagging @shainw and @petebryan

Lodewyk-Git:
Yea, this has been a problem for quite some time now tbh.

javiersoriano:
Thanks @ashwin-patil !

NVolcz:
Are there any plans for fixing this alert rule? I don't think a rule this broken should be published as a template rule :-(

Ronmarsiano:
@ashwin-patil  do we have any update on this ?

ashwin-patil:
Sorry for the delay on this fix. We have added fix for entity parsing for now in PR #1834 . 
we are also working on improving the logic of this detection to be less noisy , will raise separate PR for that change.

tzhwuen1-el:
> Sorry for the delay on this fix. We have added fix for entity parsing for now in PR #1834 . we are also working on improving the logic of this detection to be less noisy , will raise separate PR for that change.

@ashwin-patil 
How far is this fix deployed in the Azure Sentinel already since Jun 21, 2021?

I am still observing the same issue "Multiple password reset by User not mapping correctly" in my Azure Sentinel. 

The host entry is neither IP nor FQDN, but ["f.s@company.com", ""], which is still not correct. 

Thank you for your feedback

PR #1834, opened 2021-03-01
addresing issue 1771 bad entity mapping

ashwin-patil:
Fixes #

## Proposed Changes

Addressing issue reported as Bad Entity mapping -  #1771
  - removed SigninLogs as results were not corresponding directly with password reset actions by user.
  - set computer entity as blank for AuditLogs
  - set regex for Linux syslog to exclude unrelated events and to avoid blank entities


Lodewyk-Git:
Hi @ashwin-patil, is there a solution where we keep SignInLogs - this is one of the most triggered rules for this specific data type.

ashwin-patil:
> Hi @ashwin-patil, is there a solution where we keep SignInLogs - this is one of the most triggered rules for this specific data type.

Hi @Lodewyk-Git sorry for the delay, this fix has somehow got delayed. I have updated the PR with fix, we will  get this fix merged by today.

shainw (approved):

PR #1835, opened 2021-03-01
Add ReversingLabs playbook

ivukovicRL:


ivukovicRL:
@dicolanl @sarah-yo Hi, can you point me in right direction on how to solve DocumentsLinkValidation check fail?

dicolanl:
@ivukovicRL you have links like this https://docs.microsoft.com/en-us/connectors/reversinglabsintelligence/. since we want to support folks who speak all languages we check for en-us or other specific languages.

Just remove the language specifier (in this case en-us) and the link will still work and default to the the users language.  correct link https://docs.microsoft.com/connectors/reversinglabsintelligence/

ivukovicRL:
@dicolanl thx, fixed

ivukovicRL:
@dicolanl Hi, when can we expect this to be merged? Is there anything else that we need to do on our side?

dicolanl (approved):

PR #1838, opened 2021-03-02
Move New Solution Parsers

elforb:
## Proposed Changes

  - Move Parsers for new solutions to their respective Solution sub-directory


nazang (changes requested):
Eli, what is this change. The Proof point connector is already deployed. Moving the .zip file will break current customers.  Who asked for this change and why are doing this?

elforb:
> Eli, what is this change. The Proof point connector is already deployed. Moving the .zip file will break current customers. Who asked for this change and why are doing this?

Jaya and I had discussed with Preeti this morning on including Data Connectors in the Solutions, and with the upcoming Solution automation I've used Proofpoint TAP as example to compare against given that it's the only solution with a playbook, and when I'd asked Preeti what all can/should be moved for the DataConnector data she said the entire contents could be moved.

nazang:
> > Eli, what is this change. The Proof point connector is already deployed. Moving the .zip file will break current customers. Who asked for this change and why are doing this?
> 
> Jaya and I had discussed with Preeti this morning on including Data Connectors in the Solutions, and with the upcoming Solution automation I've used Proofpoint TAP as example to compare against given that it's the only solution with a playbook, and when I'd asked Preeti what all can/should be moved for the DataConnector data she said the entire contents could be moved.

Let's sync internally:

1. This connector is already deployed as a connector template. We don't want to package into the solution any more. 
2. there is existing customer using the connector, and load the .zip file from master branch. We cannot change the location of the connectors. 

For the already deployed connectors, we don't want to  move them or package them into the solution.

preetikr:
@elforb / @v-jayakal - I agree with @nazang - we should not move anything for shipped (in public preview / GA) data connectors. I see this PR is moving the parsers for new data connectors which is OK as long as we also update the aka.ms link of those referenced in data connectors to point to the right location too. Looking at this is just moving parsers I'm approving this change - please merge and also handle the aka.ms update. thanks.

PR #1850, opened 2021-03-03
Host ip insights

shainw:
Fixes #

## Proposed Changes

  -
  -
  -


Insights/Host/HostIPStats.yaml
```diff
@@ -0,0 +1,349 @@
...
+  - DataType: AzureNetworkAnalytics_CL
+  - DataType: DnsEvents
+  - DataType: CommonSecurityLog
+  - DataType: Event
+  - DataType: SecurityEvent
+  - DataType: Syslog
+EntitiesFilter: 
+ Host_OsFamily:
+  - Windows
+  - Linux
```
YaronFruchtmann, line 22: I think it is better to leave this field blank. If left blank the Insight will run for any Host. If populated - the insight will not run if the OsFamily of a Host entity  is null (was not assigned any value).

Insights/Host/HostIPStats.yaml
```diff
@@ -0,0 +1,349 @@
...
+  IpStats | join kind=leftouter Host_Account on HostIPAddress | project-away HostIPAddress1
+  };
+  GetIPResolutionWithHost('{{Host_HostName}}', '{{Host_NTDomain}}', '{{Host_DnsDomain}}', '{{Host_AzureID}}', '{{Host_OMSAgentID}}')
+
+Insights:
+ Id: 2ed19a9a-7ebe-4941-a6a0-fa0b020c48da
+ DisplayName: "Host IP address remote connections"
+ Description: |
+  'Provides the remote IP connection count information for outbound and inbound connections to an IP address associated with a given host. Note due to potential performance impact, data is limited to a 7 day max window.'
+ DefaultTimeRange:
```
YaronFruchtmann, line 276: The disclaimer states that data is "7 days" max, but time range is 12h. Since currently Insights are not configurable - is the disclaimer needed (or do we want to change the time range?)

Insights/Host/HostIPStats.yaml
```diff
@@ -0,0 +1,349 @@
...
+  QueriesDefinitions:
+  # Top inbound counts
+  - Filter:    "where InboundCount > 0"
+    Summarize: "summarize Total = sum(InboundCount) by HostIPAddress, RemoteIPAddress | top 1 by Total"
+    Project:   "project Direction = 'Top In', HostIP = HostIPAddress, RemoteIP = RemoteIPAddress, Total"
+    LinkColumnsDefinitions:
+    - ProjectedName: Direction
+      Query: "{{BaseQuery}} | {{RowFilter}}"
+    - ProjectedName: HostIP
+      Query: "{{BaseQuery}} | {{RowFilter}}"
```
YaronFruchtmann, line 302: These all link to the same query. Is this intentional? Wouldn't linking only the "Direction" column be enough?

Insights/Host/HostIPStats.yaml
```diff
@@ -0,0 +1,349 @@
...
+    - ProjectedName: RemoteIP
+      Query: "{{BaseQuery}} | {{RowFilter}}"
+    - ProjectedName: Total
+      Query: "{{BaseQuery}} | {{RowFilter}}"
+
+ AdditionalQuery: 
+  Text: "See All connections"
+  Query: "summarize min(StartTime), max(EndTime), sum(InboundCount), sum(OutboundCount), sum(ReceivedBytes), sum(SentBytes), sum(UnknownDirection) by HostIPAddress, RemoteIPAddress, Type, Computer"
```
YaronFruchtmann, line 349: Consider naming the columns

Insights/Host/HostIPStats.yaml
```diff
@@ -0,0 +1,349 @@
...
+  - Linux
+RequiredInputFieldsSets: 
+ - - Host_HostName
+   - Host_NTDomain
+ - - Host_HostName
+   - Host_DnsDomain
+ - - Host_AzureID
+ - - Host_OMSAgentID
+BaseQuery: |  
+  let GetIPResolutionWithHost = (Host_Name:string, Host_NTDomain:string, Host_DnsDomain:string, Host_AzureID:string, Host_OMSAgentID:string){
```
YaronFruchtmann, line 31: This is a great function. We should make this a solution function, Maybe even a materialized function.

Insights/Host/HostIPStats.yaml
shainw: The slider on the entity page will change the query timespan, so while for Investigation Graph it will only use 12h, for Entity Page we need to control the max range.

Insights/Host/HostIPStats.yaml
shainw: This is the default for each of the deeplinks, it is not required, but I include them so it is clear what the deeplink is using.

Insights/Host/HostIPStats.yaml
shainw: I am good with the solution function, materialize would make me a bit nervous on potential failure if we hit the 5gb limit.

Insights/Host/HostIPStats.yaml
shainw: okay, will do.

Insights/Host/HostIPStats.yaml
shainw: yep, just missed it.

PR #1860, opened 2021-03-03
Add Gain Code Execution via SMB query

jannieli:
Fixes #

## Proposed Changes

  - Add 'Gain Code Execution on ADFS Server via SMB + Remote Service/Scheduled Task' query for SecurityEvent

### DIFF
--- a/Detections/SecurityEvent/GainCodeExecutionADFSViaSMB.yaml
+++ b/Detections/SecurityEvent/GainCodeExecutionADFSViaSMB.yaml
@@ -1,0 +1,70 @@
+id: 12dcea64-bec2-41c9-9df2-9f28461b1295
+name: Gain Code Execution on ADFS Server via SMB + Remote Service/Scheduled Task
+description: |
+   'This query detects instances where an attacker has gained the ability to execute code on an ADFS Server through SMB and Remote Service/Scheduled Task.'
+severity: Medium
+requiredDataConnectors:
+  - connectorId: SecurityEvents
+    dataTypes:
+      - SecurityEvent
+queryFrequency: 1d
+queryPeriod: 1d
+triggerOperator: gt
+triggerThreshold: 0
+tactics:
+  - LateralMovement
+relevantTechniques:
+  - T1210
+tags:
+  - Solorigate
+query: |
+  let timeframe = 1d;
+  let lookback = 6d;
+  let ADFS_Servers = (
+  SecurityEvent
+  | where TimeGenerated > ago(timeframe+lookback)
+  | where EventID == 4688 and SubjectLogonId != "0x3e4"
+  | where ProcessName has "Microsoft.IdentityServer.ServiceHost.exe"
+  | distinct Computer
+  );
+  SecurityEvent
+  | where TimeGenerated > ago(timeframe)
+  | where Computer in~ (ADFS_Servers)
+  | where Account !endswith "$"
+  // Check for SMB + remote service/scheduled task, event IDs for scheduled tasks
+  // Check for scheduled task events
+  | where EventID in (4697, 4698, 4699, 4700, 4701, 4702)
+  | extend d = parse_xml(EventData)
+  | extend SubjectLogonId = tostring(d.EventData.Data[3]["#text"])
+  // Check specifically for access to IPC$ share and PIPE\svcctl and PIPE\atsvc for Service Control Services and Schedule Control Services
+  | union ( 
+      SecurityEvent
+      | where TimeGenerated > ago(timeframe)
+      | where Computer in~ (ADFS_Servers)
+      | where Account !endswith "$"
+      | where EventID == 5145
+      | where RelativeTargetName == "svcctl" or RelativeTargetName  == "atsvc"
+  )
+  // Check for lateral movement
+  | join kind=inner
+  (SecurityEvent
+  | where TimeGenerated > ago(timeframe)
+  | where Account !endswith "$"
+      | where EventID == 4624 and LogonType == 3
+      | project SubjectLogonId = TargetLogonId 
+  ) on SubjectLogonId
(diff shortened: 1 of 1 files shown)


Detections/SecurityEvent/GainCodeExecutionADFSViaSMB.yaml
```diff
@@ -0,0 +1,72 @@
...
+  | distinct Computer
+  );
+  SecurityEvent
+  | where TimeGenerated > ago(timeframe)
+  | where Computer in~ (ADFS_Servers)
+  | where Account !endswith "$"
+  // Check for SMB + remote service/scheduled task, event IDs for scheduled tasks
+  // Check for scheduled task events
+  | where EventID in (4697, 4698, 4699, 4700, 4701, 4702)
+  | extend d = parse_xml(EventData)
```
shainw, line 39: I would recommend using something more descriptive than 'd', EvData or something like that will allow this to be more readable.

Detections/SecurityEvent/GainCodeExecutionADFSViaSMB.yaml
```diff
@@ -0,0 +1,72 @@
...
+      | where EventID == 5145
+      | where RelativeTargetName == "svcctl" or RelativeTargetName  == "atsvc"
+  )
+  // Check for lateral movement
+  | join kind=inner
+  (SecurityEvent
+  | where TimeGenerated > ago(timeframe)
+  | where Account !endswith "$"
+      | where EventID == 4624 and LogonType == 3
+      | project SubjectLogonId = TargetLogonId
```
shainw, line 56: I don't recommend you overwrite another field in the data as this may cause confusion when the user reviews the query or the results.  Instead, just use the fully qualified name for the join.  For example - on $left.SubjectLogonId == $right.TargetLogonId.

Detections/SecurityEvent/GainCodeExecutionADFSViaSMB.yaml
```diff
@@ -0,0 +1,72 @@
...
+  | where EventID == 4688 and SubjectLogonId != "0x3e4"
+  | where ProcessName has "Microsoft.IdentityServer.ServiceHost.exe"
+  | distinct Computer
+  );
+  SecurityEvent
+  | where TimeGenerated > ago(timeframe)
+  | where Computer in~ (ADFS_Servers)
+  | where Account !endswith "$"
+  // Check for SMB + remote service/scheduled task, event IDs for scheduled tasks
+  // Check for scheduled task events
```
shainw, line 37: I think you only meant to make one comment here?

Detections/SecurityEvent/GainCodeExecutionADFSViaSMB.yaml
```diff
@@ -0,0 +1,72 @@
...
+  )
+  // Check for lateral movement
+  | join kind=inner
+  (SecurityEvent
+  | where TimeGenerated > ago(timeframe)
+  | where Account !endswith "$"
+      | where EventID == 4624 and LogonType == 3
+      | project SubjectLogonId = TargetLogonId 
+  ) on SubjectLogonId
+  | project TimeGenerated, Account, Computer, EventID, RelativeTargetName, IpAddress, IpPort
```
shainw, line 58: I don't think you will get an IPAddress from the left side of this unless it shows up from the 5145, the others don't have an IP address field, for example - https://docs.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4698.  Did you mean to bring thru the IPAddress from the 4624?

Detections/SecurityEvent/GainCodeExecutionADFSViaSMB.yaml
```diff
@@ -0,0 +1,72 @@
...
+name: Gain Code Execution on ADFS Server via SMB + Remote Service/Scheduled Task
+description: |
+   'This query detects instances where an attacker has gained the ability to execute code on an ADFS Server through SMB and Remote Service/Scheduled Task.'
+severity: Medium
+requiredDataConnectors:
+  - connectorId: SecurityEvents
+    dataTypes:
+      - SecurityEvent
+queryFrequency: 1d
+queryPeriod: 1d
```
shainw, line 11: This should be 7d as you are looking back 6d + 1d for the ADFS servers

Detections/SecurityEvent/GainCodeExecutionADFSViaSMB.yaml
```diff
@@ -0,0 +1,72 @@
+id: 12dcea64-bec2-41c9-9df2-9f28461b1295
+name: Gain Code Execution on ADFS Server via SMB + Remote Service/Scheduled Task
```
shainw, line 2: I don't know if the UI will like the slash here, maybe just write out the work 'or'. 'Remote Service or Scheduled Task'

Detections/SecurityEvent/GainCodeExecutionADFSViaSMB.yaml
```diff
@@ -0,0 +1,72 @@
...
+description: |
+   'This query detects instances where an attacker has gained the ability to execute code on an ADFS Server through SMB and Remote Service/Scheduled Task.'
```
shainw, line 4: Same here, maybe 'Remote Service or Scheduled Task'.

shainw (changes requested):

Detections/SecurityEvent/GainCodeExecutionADFSViaSMB.yaml
jannieli: Yes, I think I thought that the ones that had them could carry them over, but since it's not consistent that I'll take IP addresses off in the edit.

shainw (approved):

shainw (approved):

PR #1867, opened 2021-03-04
Create KVPairExtraction

CyberSekure:
Extracts key value pairs such as the below and then stores them as separate fields. Tweak the regex in the extend for your log source if required. Tested on Cisco ISE logs
DestinationIPAddress=10.0.0.1, UserName=Test, AcsSessionID=SomeSession

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Parsers/KVPairExtraction
+++ b/Parsers/KVPairExtraction
@@ -1,0 +1,5 @@
+| extend kvpairs=parse_json(extract_all("([^\\d\\s]+)=(\\S+),\\s", dynamic([1,2]), SyslogMessage))
+| mv-apply kvpairs on (
+    summarize make_bag(pack(tostring(kvpairs[0]), kvpairs[1]))
+)
+| evaluate bag_unpack(bag_)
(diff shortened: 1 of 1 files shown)


CyberSekure:
Just noticed that some fields were being extracted with a hyphen and KQL doesn't seem to like these. I've added a replace command to remove hyphens

Parsers/KVPairExtraction
```diff
@@ -0,0 +1,5 @@
+| extend kvpairs=parse_json(extract_all("([^\\d\\s]+)=(\\S+),\\s", dynamic([1,2]), SyslogMessage))
```
shainw, line 1: Please follow the parser format identified here - https://github.com/Azure/Azure-Sentinel/wiki/Contribute-to-Sentinel-GitHub-Community-of-Queries#example-parser

shainw (changes requested):

shainw (approved):

PR #1885, opened 2021-03-08
Created Rule Migration folder in root

iwafula025:
Created folder to upload artefacts to support rule migration from ArcSight, QRadar and Splunk
  -
  -


iwafula025:
Cancelling to fix branch

PR #1944, opened 2021-03-15
Submit HealthService start/stop query

jannieli:
Fixes #

## Proposed Changes

  - Submit HealthService start/stop query

### DIFF
--- a/Detections/SecurityEvent/StartStopHealthService.yaml
+++ b/Detections/SecurityEvent/StartStopHealthService.yaml
@@ -1,0 +1,55 @@
+id: 2bc7b4ae-eeaa-4538-ba15-ef298ec1ffae
+name: Starting or Stopping HealthService to Avoid Detection
+description: |
+   'This query detects events where an actor is stopping or starting HealthService to disable telemetry collection/detection from the agent.
+    The query requires a SACL to audit for access request to the service.'
+severity: Medium
+requiredDataConnectors:
+  - connectorId: SecurityEvents
+    dataTypes:
+      - SecurityEvent
+queryFrequency: 1d
+queryPeriod: 1d
+triggerOperator: gt
+triggerThreshold: 0
+tactics:
+  - DefenseEvasion
+relevantTechniques:
+  - T1562.001
+tags:
+  - Solorigate
+  - NOBELIUM
+query: |
+  let timeframe = 1d;
+  SecurityEvent
+  | where TimeGenerated > ago(timeframe)
+  | where EventID == 4656
+  | extend data = parse_xml(EventData).EventData.Data
+  | extend ObjectServer = tostring(data[4]["#text"])
+  | extend ObjectType = tostring(data[5]["#text"])
+  | extend ObjectName = tostring(data[6]["#text"])
+  | where ObjectServer =~ "SC Manager"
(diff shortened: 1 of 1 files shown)


Detections/SecurityEvent/StartStopHealthService.yaml
```diff
@@ -0,0 +1,55 @@
...
+  | extend data = parse_xml(EventData).EventData.Data
+  | extend ObjectServer = tostring(data[4]["#text"])
+  | extend ObjectType = tostring(data[5]["#text"])
+  | extend ObjectName = tostring(data[6]["#text"])
+  | where ObjectServer =~ "SC Manager"
+  | where ObjectType =~ "SERVICE OBJECT"
+  | where ObjectName =~ "HealthService"
+  | project-away data
+  // Comment out the join below if the SACL only audits users that are part of the Network logon users, i.e. with user/group target pointing to "NU."
+  | join (
```
shainw, line 36: Let's look at making this a leftouter join kind so that we don't lose the hit if there is no match in the right side of the join.

Detections/SecurityEvent/StartStopHealthService.yaml
```diff
@@ -0,0 +1,55 @@
...
+  - T1562.001
+tags:
+  - Solorigate
+  - NOBELIUM
+query: |
+  let timeframe = 1d;
+  SecurityEvent
+  | where TimeGenerated > ago(timeframe)
+  | where EventID == 4656
+  | extend data = parse_xml(EventData).EventData.Data
```
shainw, line 27: We may want to consider expanding more of the EventData fields so we always have the right index location for this and the later parsing.  You could use the method from the [Sysmon All Events parser](https://github.com/Azure/Azure-Sentinel/blob/master/Parsers/Sysmon/Sysmon-AllVersions_Parser.txt) -
```
| extend EventData = parse_xml(EventData).DataItem.EventData.Data
| mv-expand bagexpansion=array EventData
| evaluate bag_unpack(EventData)
| extend Key=tostring(['@Name']), Value=['#text']
```

Detections/SecurityEvent/StartStopHealthService.yaml
```diff
@@ -0,0 +1,55 @@
...
+  | extend ObjectServer = tostring(data[4]["#text"])
+  | extend ObjectType = tostring(data[5]["#text"])
+  | extend ObjectName = tostring(data[6]["#text"])
+  | where ObjectServer =~ "SC Manager"
```
shainw, line 31: These three where statements could be one statement with ```and``` used instead of separate lines

Detections/SecurityEvent/StartStopHealthService.yaml
```diff
@@ -0,0 +1,55 @@
...
+triggerThreshold: 0
+tactics:
+  - DefenseEvasion
+relevantTechniques:
+  - T1562.001
+tags:
+  - Solorigate
+  - NOBELIUM
+query: |
+  let timeframe = 1d;
```
shainw, line 23: no timeframe needed as you are not doing a lookback comparison.  the queryPeriod will control the timeframe.  You can remove this and the ```| where TimeGenerated...``` parts.

shainw (changes requested):

shainw (approved):

Issue #1561, opened 2021-01-07
False Positives on PasswordNotSet Detection


gcap12:
**Describe the bug**
False positives are generated from the "PasswordNotSet" detection:
https://github.com/Azure/Azure-Sentinel/blob/master/Detections/SecurityEvent/password_not_set.yaml
Accounts are showing that have had the account password reset within the same day and sometimes even seconds after the initial 4722 event. Also this query is looking for 4723 which is a failed reset instead of looking for 4724 which is a successful reset. If this is what the query is intending to do then maybe some wording can be changed in the description to reflect this. 

**To Reproduce**
1. Run query from above yaml file in Kusto.
2. User accounts appear that have had their password reset successfully well within the 24-48 hour threshold. 

**Expected behavior**
It should show when an account created with a default password hasn't changed it within 24 hours.

**Screenshots**
Unable to add due to PII/security concerns.

shainw:
@gcap12 - Taking a look at this and I am not sure it is incorrect in relation to the events we are using.  Documentation indicates the following:

- [4724](https://docs.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4724) occurs when someone other than the user changes the password, specifically - "This event generates every time an account attempted to reset the password for another account."  
  - Effectively, this occurs when an account is enabled by an admin and a default password is set.
- [4723](https://docs.microsoft.com/en-us/windows/security/threat-protection/auditing/event-4723) occurs when the user associated to the account attempts to change the account password, specifically - "This event generates every time a user attempts to change his or her (their own) password."  
  - Additionally, the event 4723 occurs for Success or Fail as indicated by the Keywords in the event, however, this is not currently available in the 4723 event in the SecurityEvent table.  This is being looked into.

The initial goal of this template is to identify if a user attempts to set their password after the account is enabled within a timely manner not if the attempt was successful or not.  I think that updating the language would help clarify this and if we had the Keywords value in the event we could indicate success or failure.  I do agree this is something we should include.

Also, reviewing the time ranges of the 4722 and 4723 there is indication of a bug.  I have added a PR to fix this, among other improvements - #1722 - please feel free to review.

shainw:
@gcap12 - any further questions?  #1722  has been merged into master and should be available in the Azure Sentinel UX in the next week.

PR #1957, opened 2021-03-17
Update password_not_set.yaml with new Event ID

Lodewyk-Git:
## Proposed Changes
  -Added event ID 4724 which indicates password reset attempt. 
  -Also changed the description and reason to match the change. 
  -Also changed the variable name for event ID 4723 to be generalized for event ID's 4723 and 4724

### DIFF
--- a/Detections/SecurityEvent/password_not_set.yaml
+++ b/Detections/SecurityEvent/password_not_set.yaml
@@ -2,8 +2,8 @@ id: 62085097-d113-459f-9ea7-30216f2ee6af
 name: AD user enabled and password not set within 48 hours
 description: |
   'Identifies when an account is enabled with a default password and the password is not set by the user within 48 hours.
-  Effectively, there is an event 4722 indicating an account was enabled and within 48 hours, no event 4723 occurs which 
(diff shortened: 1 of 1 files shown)


shainw (commented):
@Lodewyk-Git - This was discussed in the issue #1561.  If you notice the description, you can see this was intended to indicate when "the user" doesn't set their password.  This was a use case that several customers requested.  By adding in the 4724, this is going to include when any user (likely an admin) "resets".  My suggestion is that we create another detection to cover the scenario where a user account is enabled and the password is not reset by an admin.  I don't want to conflate the 2 scenarios as it is very specific to the user in this case.  Thoughts?

Lodewyk-Git:
Hi @shainw , Thanks for the link, I agree that the description for the alert needed some updating as it was not following the rule logic. Looks like the update has been pushed through to Sentinel. 

I agree that there should be a new rule created specific to accounts with escalated privileges, "Privileged AD User Account and password not changed within 24-48 hours"

shainw:
@Lodewyk-Git - Great, I will close this PR.  Thanks for all the contributions lately, very helpful and keep them coming. :)

PR #1966, opened 2021-03-17
CVE-2021-27065 Detection

petebryan:
Detection of CVE-2021-27065 being exploited to drop webshells. Detection occurs on AD DS rather than Exchange sever.

### DIFF
--- a/Detections/SecurityEvent/ExchangeOABVirtualDirectoryAttributeContainingPotentialWebshell.yaml
+++ b/Detections/SecurityEvent/ExchangeOABVirtualDirectoryAttributeContainingPotentialWebshell.yaml
@@ -1,7 +1,9 @@
 id: faf1a6ff-53b5-4f92-8c55-4b20e9957594
 name: Exchange OAB Virtual Directory Attribute Containing Potential Webshell
 description: |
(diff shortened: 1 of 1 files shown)


petebryan:
Thanks to @piaudonn and Karine Sanguinet for identifying this detection opportunity.

Detections/SecurityEvent/ExchangeOABVirtualDirectoryAttributeContainingPotentialWebshell.yaml
```diff
@@ -0,0 +1,43 @@
...
+tactics:
+  - InitialAccess
+relevantTechniques:
+  - T1190
+query: |
+  SecurityEvent
+  // Look for specific Directory Service Changes
+  | where EventID == 5136
+  | extend Data = parse_xml(EventData)
+  | extend ObjectClass = tostring( Data.EventData.Data[10].["#text"])
```
shainw, line 23: We may want to consider  ```EventData has 'msExchOABVirtualDirectory'``` and then expanding more of the EventData fields so we always have the right index location for this and the later parsing.  You could use the method from the [Sysmon All Events parser](https://github.com/Azure/Azure-Sentinel/blob/master/Parsers/Sysmon/Sysmon-AllVersions_Parser.txt) -
```
| extend EventData = parse_xml(EventData).DataItem.EventData.Data
| mv-expand bagexpansion=array EventData
| evaluate bag_unpack(EventData)
| extend Key=tostring(['@Name']), Value=['#text']
```

shainw (dismissed):
Suggested changes, but not required.

shainw (approved):

PR #1975, opened 2021-03-18
Create Certutil-LOLBins.yaml

chihebchebbi:
A new hunting query to hunt a Living Off The Land Binary (LOLBIN) certutil

### DIFF
--- a/Hunting Queries/SecurityEvent/Certutil-LOLBins.yaml
+++ b/Hunting Queries/SecurityEvent/Certutil-LOLBins.yaml
@@ -1,0 +1,41 @@
+id: 0e429446-2798-49e4-924d-c37338f24e23
+name: Certutil (LOLBins and LOLScripts)
+description: |
+  'This detection uses Sysmon telemetry to hunt Certutil activities'
+
+requiredDataConnectors:
+  - connectorId: SecurityEvents
+    dataTypes:
+      - SecurityEvent
+tactics:
+  - Command And Control
+relevantTechniques:
+  - T1105
+
+query: |
+  let timeframe = 1d;
+  Event
+  //This query uses sysmon data depending on table name used this may need updataing
+  | where TimeGenerated >= ago(timeframe)
+  | where Source == "Microsoft-Windows-Sysmon"
+  | where EventID == 1
+  | extend RenderedDescription = tostring(split(RenderedDescription, ":")[0])
+  | extend EventData = parse_xml(EventData).DataItem.EventData.Data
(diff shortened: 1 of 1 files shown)


Hunting Queries/SecurityEvent/Certutil-LOLBins.yaml
```diff
@@ -0,0 +1,41 @@
...
+  | where EventID == 1
+  | extend RenderedDescription = tostring(split(RenderedDescription, ":")[0])
+  | extend EventData = parse_xml(EventData).DataItem.EventData.Data
+  | mv-expand bagexpansion=array EventData
+  | evaluate bag_unpack(EventData)
+  | extend Key=tostring(['@Name']), Value=['#text']
+  | evaluate pivot(Key, any(Value), TimeGenerated, Source, EventLog, Computer, EventLevel, EventLevelName, EventID, UserName, RenderedDescription, MG, ManagementGroupName, Type, _ResourceId)
+  | extend RuleName = column_ifexists("RuleName", ""), TechniqueId = column_ifexists("TechniqueId", ""),  TechniqueName = column_ifexists("TechniqueName", "")
+  | parse RuleName with * 'technique_id=' TechniqueId ',' * 'technique_name=' TechniqueName
+  | where Image has "certutil.exe"
```
shainw, line 30: I think this will be noisy on its own based on common commandlines, even for a hunting query. :) I suggest to add in an ignore list related to common commandlines, it can be commented out by default, but would be easier for users if it was already available and they could add to the list.

shainw (changes requested):

Hunting Queries/SecurityEvent/Certutil-LOLBins.yaml
shainw: @chihebchebbi - want to make sure you saw this feedback. :)

chihebchebbi: @shainw Thank you so much for the feedback. Yes, i saw it and i am planning to fix it asap

chihebchebbi:
Maybe we can use another option for the ignored commandlines as follows:
`
// An example of an ignored commandline 
let CommandLine_Ignore_list = dynamic(["certutil.exe  -urlcache -split -f http://7-zip.org/a/7z1604-x64.exe 7zip.exe"]);
| where CommandLine !in~ (CommandLine_Ignore_list)
`

shainw (approved):

PR #1984, opened 2021-03-21
updating entity mappings and descriptions to fix some characters that…

shainw:
… may cause issue on import to other tooling.

Fixes #

## Proposed Changes

  -
  -
  -


laithhisham (commented):
Hunting queries still do not support the new entity mappings
This probably won't work :(

shainw:
> Hunting queries still do not support the new entity mappings
> This probably won't work :(

Yeah, this is just prepping for future and I need them for another process we are using. :)

PR #2090, opened 2021-04-06
Advanced Multistage Attack Detection (aka Fusion) Analytic rule should be enabled by default

viskumar-microsoft:
Fixes #

## Proposed Changes

  - Advanced Multistage Attack Detection (aka Fusion) Analytic rule should be enabled by default
  -
  -


sreedharande:
Thanks @viskumar-microsoft. I am closing this PR. In ARM Template, the default values for all the Analytical rules is false. This setting will be updated based on user's action while deploying, attached screenshot for reference 
image

PR #2093, opened 2021-04-06
ACN_CD_SEP_Parser04

acnccd:
## Proposed Changes

  - Modifications to Agent Security and Agent Behavior logs regex, working in live environment did not correctly parse these. Loosened up the regex on these log types.
  - Other minor modifications to resolve data types and typos.


preetikr:
@shainw for review and @v-jayakal for solution update

shainw (approved):
Tested against temp dataset, looks good.

PR #2110, opened 2021-04-08
Fixing a Typo for SysmonEvent13_RegistrySetValue function

sukster:
Hello, there was a typo in let SysmonEvent13_RegistrySetValue function. In the original there was an extra underscore like this: SysmonEvent13__RegistrySetValue=() while the projected value was with only one underscore like this: SysmonEvent13_RegistrySetValue.

Fixes #

## Proposed Changes

  -
  -
  -


sukster:
I forgot to add that because of this type the Event ID 13 was not parsed and projected in the query results.

shainw (approved):

PR #2125, opened 2021-04-09
fixed casing on SrcIpAddr extract - lower case left field empty

FrodeHus:
Fixes #

SrcIpAddr field is empty due to casing of regex - fixed by using uppercase

## Proposed Changes

  -
  -
  -


FrodeHus:
Just want to clarify -- the log message looks somewhat like this:
`<4>Apr  9 16:36:53 ubnt kernel: [WAN_LOCAL-1234-R]IN=eth0 OUT= MAC=e0:63:xx:29:da:ØØ:b0:a8:6e:f1:60:a6:ko:00 SRC=111.123.23.44 DST=84.64.119.165 LEN=60 TOS=0x00 PREC=0x00 TTL=47 ID=5177 DF PROTO=TCP SPT=18752 DPT=3306 WINDOW=64240 RES=0x00 SYN URGP=0 
`
so the `extract()` function runs the regex as specified - `src` does not match and the `SrcIpAddr` is empty

v-jayakal:
@FrodeHus , closing this PR as this has been taken care of by the PR: https://github.com/Azure/Azure-Sentinel/pull/2211

PR #2134, opened 2021-04-12
Create SignedBinaryProxyExecutionRundll32.yaml

chihebchebbi:
A new  rule to hunt Rundll32 activities

### DIFF
--- a/Hunting Queries/SecurityEvent/SignedBinaryProxyExecutionRundll32.yaml
+++ b/Hunting Queries/SecurityEvent/SignedBinaryProxyExecutionRundll32.yaml
@@ -21,7 +21,7 @@ query: |
   | extend EventData = parse_xml(EventData).DataItem.EventData.Data
   | mv-expand bagexpansion=array EventData
   | evaluate bag_unpack(EventData)
-  | extend Key=tostring(['@Name']), Value=['#text']
+  | extend Key = tostring(column_ifexists('@Name', "")), Value = column_ifexists('#text', "")
   | evaluate pivot(Key, any(Value), TimeGenerated, Source, EventLog, Computer, EventLevel, EventLevelName, EventID, UserName, RenderedDescription, MG, ManagementGroupName, Type, _ResourceId)
   | extend RuleName = column_ifexists("RuleName", ""), TechniqueId = column_ifexists("TechniqueId", ""),  TechniqueName = column_ifexists("TechniqueName", "")
   | parse RuleName with * 'technique_id=' TechniqueId ',' * 'technique_name=' TechniqueName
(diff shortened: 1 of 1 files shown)


Hunting Queries/SecurityEvent/SignedBinaryProxyExecutionRundll32.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - connectorId: SecurityEvents
+    dataTypes:
+      - SecurityEvent
+tactics:
+  - Defense Evasion
+relevantTechniques:
+  - T1218.011
+
+query: |
+  let timeframe = 1d;
```
shainw, line 16: Remove, see explanation in overall comments.

Hunting Queries/SecurityEvent/SignedBinaryProxyExecutionRundll32.yaml
```diff
@@ -0,0 +1,43 @@
...
+  Event
+  //This query uses sysmon data depending on table name used this may need updataing
+  | where TimeGenerated >= ago(timeframe)
```
shainw, line 19: Remove, see explanation in overall comments.

shainw (changes requested):
We are not including timeframes any longer when there is no comparison lookback.  This is to support the Hunting UX and how hard coded timeframes can cause issues with running queries automatically.  Please remove ```let timeframe = 1d;``` and ```| where TimeGenerated >= ago(timeframe)```

Hunting Queries/SecurityEvent/SignedBinaryProxyExecutionRundll32.yaml
```diff
@@ -0,0 +1,41 @@
...
+query: |
+  Event
+  //This query uses sysmon data depending on table name used this may need updataing
+  | where Source == "Microsoft-Windows-Sysmon"
+  | where EventID == 1
+  | extend RenderedDescription = tostring(split(RenderedDescription, ":")[0])
+  | extend EventData = parse_xml(EventData).DataItem.EventData.Data
+  | mv-expand bagexpansion=array EventData
+  | evaluate bag_unpack(EventData)
+  | extend Key=tostring(['@Name']), Value=['#text']
```
shainw, line 24: This query will fail execution and rule creation if we don't have a recent event or an event within the timeframe.  Changing this line to - `| extend Key = tostring(column_ifexists('@Name', "")), Value = column_ifexists('#text', "")` - will allow it to not fail.

shainw (changes requested):
One more item I missed previously.  Quick fix and then good to go.

Hunting Queries/SecurityEvent/SignedBinaryProxyExecutionRundll32.yaml
chihebchebbi: Done. Thank you @shainw

shainw (approved):

PR #2180, opened 2021-04-21
Add Palo Alto High Risk Port hunting query

mmaitre314:
Add a hunting query listing communications on dangerous ports which were not blocked by Palo Alto firewalls and need to be reviewed to ensure sufficient network isolation.


v-jayakal (dismissed):

v-jayakal:
@shainw , metadata looks good, approved.

shainw (changes requested):
I would suggest swapping the left and right sides of the join here.  The HighRiskPorts is a much smaller data set and when comparing with very large CommonSecurityLog the performance is not great in its current format.  Based on testing in large data sets, swapping the sides of the join cuts the query time by 50% and the total CPU time used by 33% for a 24 hour period, even more when expanding to 3 and 7 days.

shainw (approved):

PR #2192, opened 2021-04-22
Create AdFind_Usage.yaml

TestLeper:
Fixes #

## Proposed Changes

  -
  -
  -


shainw (changes requested):
Can you test using the same regex pattern match in a related hunting query - https://github.com/Azure/Azure-Sentinel/blob/master/Hunting%20Queries/SecurityEvent/Suspicious_enumeration_using_adfind.yaml.  It would be good to use the same method if possible and the regex might be more resilient to things like extra spaces and such.

shainw (approved):

PR #2227, opened 2021-04-29
AD FS  Configuration Settings Export (Local and Remote) Rules

Cyb3rWard0g:
Three rules to use AD FS auditing, Sysmon and Application events to capture users connecting to the AD FS server and monitor incoming network traffic over port 80 on the AD FS server to detect the potential use of replication to get the AD FS database configuration settings from the server.

References: https://twitter.com/OTR_Community/status/1387038995016732672

Fixes #
+ Pivot error on one rule where the field name `type` already existed in the pivot table. Events 20 and 25 have a field with that name. Check schema: https://github.com/OTRF/OSSEM/blob/master/resources/schemas/sysmonv13.10_4.60.xml
## Proposed Changes
+ Update on rule name to clarify detection goal. Rule was built to detect local connections to the AD FS database over a specific named pipe. The name was reflecting a different action.
+ Moved Filter to look for events 18 to the top of rule.
+ Removed `type` from the Pivot function since event 20 and 25 have a field with that name. Before change, it was throwing a `Pivot` error saying that the field already existed.


shainw (changes requested):
We found out that these types of queries using the bag_unpack function can have a problem in the UX with the Key creation part.  To resolve this, we have added in a [column_ifexists()](https://docs.microsoft.com/azure/data-explorer/kusto/query/columnifexists) check.  The query runs fine, but the rule wizard is using a limited data set that may not have the event.   Please update all of the Detections.  It looks like you will need to use the column if exists check on several locations for each query - Example changes here - https://github.com/Azure/Azure-Sentinel/pull/2191/files.

Cyb3rWard0g:
Hello @shainw ! Everything looks good now I believe. Thanks to @ashwin-patil !

shainw (approved):

PR #2229, opened 2021-04-29
Checking for column in case evaluate doesn't contain the new column.

shainw:
Fixes #

## Proposed Changes

  - Adding in handling if column is not present after bag unpack
  - ```| extend displayName = column_ifexists("displayName", "NotAvailable"), newValue = column_ifexists("newValue", "NotAvailable")```
  -


juliango2100 (approved):
Reviewed and approved.

PR #2269, opened 2021-05-10
Adding initial versioning to detection templates

Amitbergman:
Fixes #

## Proposed Changes

  -
  -
  -


.script/tests/detectionTemplateSchemaValidation/Models/ScheduledTemplateInternalModel.cs
```diff
@@ -22,6 +22,9 @@ public class ScheduledTemplateInternalModel : AnalyticsTemplateInternalModelBase
         [JsonProperty("query", Required = Required.Always)]
         [StringLength(10000, MinimumLength = 1)]
         public string Query { get; set; }
+        
+        [JsonProperty("version", Required = Required.Always)]
```
Ronmarsiano, line 26: isn't there a way to just say that default version is 1.0.0?

PR #2308, opened 2021-05-19
Feature/noamlandress/cef get info new version

NoamLandress:
Updating the script version with some minor changes-
1. Adding 2 additional tests to collect some logs from the messages/syslog files.
2. Adding a condition before inserting the command output to the file that checks it's in a string format (needed for Ubuntu 20 support)
3. Adding a shebang in order to be able to run the script as an executable.


liemilyg (approved):

shschwar (commented):
Approved

PR #2310, opened 2021-05-19
Get samAccountName from output of previous action

### LINKED ISSUE #2309: Playbook Get-AlertEntitiesEnrichment fails trying to add a comment >3000 characters
**Describe the bug**
Playbook fails trying to add a comment >3000 characters because the results (from Advanced Hunting action contains all devices not just the users devices) - error is in the query returning too many results

Playbook Get-AlertEntitiesEnrichment | UserEnrichment.template.json | onPremisesSamAccountName

**To Reproduce**
Trigger the Get-AlertEntitiesEnrichment playbook on an incident with a user entity that does not have an on premises account

**Expected behavior**
Query should only return user's MDATP devices

**Additional context**
We do not have any on-prem Active Directory - all our users are AAD only

rohancragg:
Fixes #2309

## Proposed Changes
Use output of previous action (samAccountName) in Advanced Hunting query - i.e. where onPremisesSamAccountName is an empty string and causes too many results form query


rohancragg:
This only fixes an edge case with the Devices scope of the UserEnrichment logic app. It does not resolve the limitation that the Add comment in Incident action is limited to posting comments up to 3000 characters

PR #2316, opened 2021-05-20
Include values for Apps

### LINKED ISSUE #2314: AAD Audit Logs workbook not parsing the Service Principal ID
**Describe the bug**
https://github.com/Azure/Azure-Sentinel/blob/master/Workbooks/AzureActiveDirectoryAuditLogs.json

AAD workbook is parsing only the User Principal name, however there are many cases where changes are being done by a Service Principal ID. 

The query is checking if the User Principal is not null, otherwise puts "unknown" for everything else. This is causing the workbook to show "unknown" for most of the cases (example below 126 Update User changes, 118 marked as Unknown)

**Expected behavior**
Instead of parsing the SP as unknown, the ID should be displayed.

**Screenshots**
image
image
image

Lodewyk-Git:
Fixes #2314

## Proposed Changes
Changed method of getting data, also added values for Application
Changed User data from user DisplayName to userPrincipalName - displayname is sometimes null when UPN is populated.
Displaying servicePrincipalID as SPName is always null and dispalyname only shows application - could be changed.

### DIFF
--- a/Workbooks/AzureActiveDirectoryAuditLogs.json
+++ b/Workbooks/AzureActiveDirectoryAuditLogs.json
@@ -84,7 +84,7 @@
             "multiSelect": true,
             "quote": "'",
             "delimiter": ",",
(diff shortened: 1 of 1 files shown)


Lodewyk-Git:
Hi @liatlishams any progress on this?

shainw:
@liatlishams and @Amitbergman - please review or the plan is to close this PR by end of November due to age.

PR #2322, opened 2021-05-20
GroupName was commented out

### LINKED ISSUE #2319: Analytics rule "User added to Azure Active Directory Privileged Groups" not working as intended
Analytics rule "User added to Azure Active Directory Privileged Groups": the rule is supposed to alert from a specific list of groups set with the "let PrivilegedGroups" statement in the query. But the actual Where operator that would match the list is commented out, so the rule does not work as intended.

https://github.com/Azure/Azure-Sentinel/blob/master/Detections/AuditLogs/UseraddedtoPrivilgedGroups.yaml

Lodewyk-Git:
Fixes #2319


ashwin-patil (changes requested):
@Lodewyk-Git Could you add version property back and also update its version ? It is required field in the template hence build validation is failing. rest looks good to me. Thanks for submitting the change.

ashwin-patil (approved):

PR #2327, opened 2021-05-21
FlareSystems: add workbook

aviau:
Adds a workbook to the Flare Systems solution.


aviau:
Nvm, we are not ready yet.

PR #2326, opened 2021-05-21
FlareSystems: Adjust logo and add Workbook

aviau:
- Fix the color of the logo which was unintentionally black
- Add a workbook
- Rename Flare Systems' solution folder to remove spaces which messed with `createSolution.ps1`.

Thanks!

PR #2328, opened 2021-05-21
FlareSystems: add workbook

aviau:
This adds a workbook to our solution!

Cheers,


aviau:
Included as part of https://github.com/Azure/Azure-Sentinel/pull/2326

PR #2332, opened 2021-05-23
Add basic playbooks validations

ityankel:
Add very initial validation for playbook resources - to validate that json file contains a valid ARM template schema, and simple validations on the content (if template contains a playbook resource, its name and location match the expected convention).
After completing this PR I'll add more complex validations on other parts of the template - API connections, variables, metadata, etc.


.script/playbooksValidator.ts
```diff
@@ -0,0 +1,52 @@
...
+let fileKinds = ["Modified"];
+let CheckOptions = {
+  onCheckFile: (filePath: string) => {
+    return IsValidTemplate(filePath);
+  },
+  onExecError: async (e: any, filePath: string) => {
+    console.log(`Playbooks validation failed. File path: ${filePath}. Error message: ${e.message}`);
+  },
+  onFinalFailed: async () => {
+    logger.logError("An error occurred, please open an issue");
```
Amitbergman, line 48: where will we see these logs?

Amitbergman: In the results of the test?

Amitbergman (approved):

.azure-pipelines/playbooksValidations.yaml
```diff
@@ -0,0 +1,12 @@
+jobs:
```
Amitbergman, line 1: How did you do it not required?
image

ityankel: I have no idea :(

ityankel: I'm fine with this not being required check, at least at the start, so we'll be able to verify these new validations work fine and don't block contribution. But I have no idea what controls it so I don't know how to change it to required later on

.script/playbooksValidator.ts
ityankel: my understanding is that they appear in the logs when you open the build pipeline for the specific check

PR #2370, opened 2021-05-28
fix a transcription error in one of the hashes

qc-gordon:
a sha-256 should have 64 characters, this only had 63.
I found the full one here

https://www.microsoft.com/security/blog/2021/05/27/new-sophisticated-email-based-attack-from-nobelium/


shainw:
@qc-gordon -  Thanks!! Fix in #2468 due to this PR not having latest and not wanting to resolved branch conflicts.

PR #2377, opened 2021-05-28
Update CloudflareUnexpectedPost.yaml

dpeeke:
This query fails on enabling the analytic, need to convert HttpStatusCode to string before "startswith" comparison.


preetikr:
@shainw to review and @v-jayakal to update the solution package.

v-jayakal (approved):

PR #2381, opened 2021-05-30
Add more validations for Playbooks

ityankel:
Add more validations for Playbooks:
1. All parameters in the ARM template should have a user-friendly description to explain the parameter value's usage and format
2. All API connection resources deployed in the template should have their resource name and display name as values taken from variables to allow them to be computed from the playbook name + those variables should exist in the template. Validation on the value of the API connection variables (that they reference the playbook name parameter + the resource name variable references the API connection type) will be added in a separate PR to keep this one short and focused


preetikr:
@lior-tamir - please help review this PR / playbook validations. Thanks.

lior-tamir (approved):

PR #2393, opened 2021-06-03
Update CorrelateIPC_Unfamiliar-Atypical.yaml

briandelmsft:
The KQL logic was not consistent with the description of the alert in line 4.  The description indicates a correlation within 20 minutes but the KQL logic looked for correlation within 10 minutes

Fixes #

## Proposed Changes

  - Update KQL to look for +/- 20 minutes instead of +/- 10 minutes
  -
  -

### DIFF
--- a/Detections/SecurityAlert/CorrelateIPC_Unfamiliar-Atypical.yaml
+++ b/Detections/SecurityAlert/CorrelateIPC_Unfamiliar-Atypical.yaml
@@ -1,7 +1,7 @@
 id: a3df4a32-4805-4c6d-8699-f3c888af2f67
 name: Correlate Unfamiliar sign-in properties and atypical travel alerts
 description: |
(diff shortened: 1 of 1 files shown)


shainw (changes requested):
The timeframe is +-10m so 20 min window around the Alert time so the 10m value is correct. I would recommend an update to the description to make this more clear versus changing the time window.

briandelmsft:
Updated description and reverted KQL change based on feedback

shainw (approved):
@briandelmsft - I did some further tweaks, look good?

briandelmsft:
I'm good with that option, but here's a slight rewording for your consideration.  Either way is fine with me.

"The combination of an Unfamiliar sign-in properties alert and an Atypical travel alert about the same user within a +10m or -10m window is considered a high severity incident."

shainw:
@briandelmsft - Better wording, I agree.  I implemented it and merged the PR.  Thanks for having a look at this and improving the explanation.

PR #2434, opened 2021-06-08
Adding Watchlist template content.

malowe101:
Fixes #

## Proposed Changes
Adding Nobelium TI template. Modifying existing templates and readmes.
  -
  -
  -


malowe101:
Need to update templates again with small changes

Issue #2445, opened 2021-06-10
False Positives from "Corelight - Forced External Outbound SMB" rule


mburrough:
**Describe the bug**
The rule "Corelight - Forced External Outbound SMB" fired for many hosts immediately upon running. On closer investigation, it was triggering for all kinds of traffic, including HTTPS web browsing.

**To Reproduce**
1. Set up Corelight sensor and configure to feed events to Sentinel
2. Enable "Corelight - Forced External Outbound SMB" rule
3. Create normal HTTPS traffic on network
4. Observe alert

**Expected behavior**
The alert only fires on SMB traffic, not other TLS traffic

**Additional context**
The "`NetworkConnectionHistory hasprefix 'Sh' or`" part of the rule on line 23 of [this file](https://github.com/Azure/Azure-Sentinel/blob/a6f6275b4ba79f8bfcefe7ae72afcb055a5bbf80/Solutions/Corelight/Analytic%20Rules/CorelightForcedExternalOutboundSMB.yaml) seems to make this fire on all kinds of traffic, not just SMB. Removing this part of the rule eliminated the false positives. I have not tested to confirm this doesn't break the rule for true positives, however.
Reviewing the documentation, Sh here indicates any Syn/Syn+Ack message:
> history (string) Records the state history of connections as a string of letters:
> • s a SYN w/o the ACK bit set
> • h a SYN+ACK (“handshake”)

Perhaps then this should be an AND not an OR.

PR #2447, opened 2021-06-10
Update CorelightForcedExternalOutboundSMB.yaml

mburrough:
Eliminate false positives in Corelight - Forced External Outbound SMB rule

Fixes Azure/Azure-Sentinel#2445

## Proposed Changes
 - Change the 'or' to an 'and' in the rule logic


shainw:
@v-ampami and @v-maudan - This looks like it is the correct change, but since you both worked on the original Solution would like your review.

v-jayakal (approved):

PR #2467, opened 2021-06-12
Add ApacheHTTP TI Rule

mburrough:
This change adds a new "(Preview) TI map IP entity to Apache HTTP" rule, modeled after the existing TI rule for IIS logs. 

## Proposed Changes
  - New TI Rule for ApacheHTTPServer
  - Updated ApacheHTTPServer function to include the Computer value in its output


oshezaf:
This would be prohibitive performance wise, and would not work for many customers. The reason is that the Apache parser will work on 14 days of data each time, rather than just an hour. To make it work, the query cant use the parser, but rather parse out the relevant fields after filtering for time.

PR #2516, opened 2021-06-17
UEBA Workbook patch - Query fails when multiple param values selected

briandelmsft:
In the Anomalies Breakdown section of the workbook if you select 2+ Anomaly Name, Tactic and/or Anomaly Scores from the drop down parameters the dependent queries fail.  The where statements in the queries are not properly setup to handle the array of values that selecting multiple items causes.  This PR fixes that issue, allowing 1, multiple or all options to be selected without issue.

Also, if you select 0 values from any of the above parameters, the query also fails so this change includes making those parameters required.

In addition, an option for '0' was added to the Anomaly Score Parameter as it is possible to have an anomaly with score 0, this enables selecting it.


v-admahe:
@briandelmsft Workbook query is not working, could you please check the query and do the required changes.

Please find the screenshot below:
image
image

Thanks,
Aditi Maheshwari

briandelmsft:
@v-admahe it looks like you don't have UEBA enabled in the workspace as your error is loading the BehaviorAnalytics table.  You would see the same error if you tried the currently published version of the workbook in your config...

PR #2544, opened 2021-06-22
Process Events Normalization Package 

oshezaf:
- New normalizing parsers for process events
  - Migrated detection and hunting queries that use normalized data


oshezaf:
@shainw: we need your review of the detections and hunting queries. All just converted. We review internally the parsers.

Issue #2408, opened 2021-06-04
Update required for Azure Sentinel Infoblox NIOS parser?


oniled:
**Describe the bug**
I can confirm I am receiving Infoblox logs in Azure Log Analytics and Azure Sentinel.  However the parser is not correctly parsing a lot of the logs.  Guessing the parser may need an update?

Others have reported similar issues and the problem is not the format of the data or receiving the data, as both of those have been verified.  I am getting partial/no parsing by the parser is the problem.

**To Reproduce**
1. Setup Syslog server to forward logs to Azure Log Analytics
2. Setup Kusto Function named InfobloxNIOS based on this documentation: https://aka.ms/sentinelgithubparsersinfoblox

**Expected behavior**
Full parsing of the syslogs

**Screenshots**

![image](https://user-images.githubusercontent.com/8421682/120858380-90e91680-c537-11eb-9992-99328f217684.png)

![image](https://user-images.githubusercontent.com/8421682/120858362-89c20880-c537-11eb-92ff-a95afb19c902.png)

**Additional context**

sample syslog that is not being parsed: 

Jun  4 12:20:09 130.191.200.2 dhcpd[7625]: DHCPACK on 10.130.17.107 to b6:57:72:77:93:cc via eth2 relay 10.130.255.254 lease-duration 28727 (RENEW) uid 01:b6:57:72:77:93:cc

oniled:
Records of other having same problem:

https://www.gitmemory.com/issue/Azure/Azure-Sentinel/2296/849183343

elforb:
Hi @oniled,

I have a potential fix for this created at #2547 , could you review and confirm it solves your issue?

oniled:
This looks pretty good for the DHCPAck but maybe need work to correct "r-l-e" and "client"?  Not seeing those parse correctly.  
![image](https://user-images.githubusercontent.com/8421682/124041564-f0fa9d80-d9bb-11eb-8cf2-733857833e5f.png)
![image](https://user-images.githubusercontent.com/8421682/124041575-f821ab80-d9bb-11eb-943e-176e4ea31517.png)

v-maudan:
> This looks pretty good for the DHCPAck but maybe need work to correct "r-l-e" and "client"? Not seeing those parse correctly.
> ![image](https://user-images.githubusercontent.com/8421682/124041564-f0fa9d80-d9bb-11eb-8cf2-733857833e5f.png)
[...]

PR #2547, opened 2021-06-22
InfobloxNIOS Parser - DHCPAck Fix

elforb:
This change aims to fix the DHCPAck log parsing in the InfobloxNIOS Parser.

Fixes #
  - Modified the regex which parses DHCPAck logs
  - Modified the capture groups used for `Relay` and `LeaseDuration`
  - Update regex in DNSClient Parsing to make view portion optional, to ensure current data can be matched.
- "r-l-e" type of logs raw string field is not starting with ' ':' hence making it as Optional

### DIFF
--- a/Parsers/InfobloxNIOS/InfobloxNIOS.txt
+++ b/Parsers/InfobloxNIOS/InfobloxNIOS.txt
@@ -185,7 +185,7 @@ let dnszone = RawData
     | project-away RawData_subString, dnszone_substring, dnszone;
 let dnsclient = RawData 
     | where ProcessName == "named" and Log_Type == "client"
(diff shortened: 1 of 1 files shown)


shainw:
@acnccd - Can you have a look at this to confirm this change as you originally submitted the parser?

acnccd:
@shainw judging from log sample I found in a screenshot from a link http://www.callevanetworks.com/understanding-infoblox-isc-dhcp-and-abandoned-leases/, the suggested changes would work. If there was sample data with DHCPACK events we could further validate.

shainw:
> @shainw judging from log sample I found in a screenshot from a link http://www.callevanetworks.com/understanding-infoblox-isc-dhcp-and-abandoned-leases/, the suggested changes would work. If there was sample data with DHCPACK events we could further validate.

Agreed, @elforb - can you supply the sample log please?

elforb:
The sample log is as follows:
Jun 4 12:20:09 130.191.200.2 dhcpd[7625]: DHCPACK on 10.130.17.107 to b6:57:72:77:93:cc via eth2 relay 10.130.255.254 lease-duration 28727 (RENEW) uid 01:b6:57:72:77:93:cc

It's also available in the linked issue, #2408

v-jayakal (approved):

Issue #2497, opened 2021-06-16
No IPAddress output in RareApplicationConsent.yaml


aletcetera:
Hello!

I'm using the RareApplicationConsent KQL query to hunt, and the IPAddress field comes out empty all the time... is this expected ?

The code tries to extract the IPAddress from the InitiatedBy field, that only seems to contain the ServicePrincipalName (no json data).

Thanks!

![image](https://user-images.githubusercontent.com/47009952/122226716-3c715f80-ce84-11eb-97fb-dfe22219b22a.png)

shainw:
@aletcetera - I am taking a further look at this and its possible that the IPAddress is not always available.  Will get back to you after some further validation.

aletcetera:
Thanks, that did the trick! I think the same fix would apply to https://github.com/Azure/Azure-Sentinel/blob/master/Hunting%20Queries/AuditLogs/ConsentToApplicationDiscovery.yaml as well.

PR #2607, opened 2021-07-01
Update RareApplicationConsent.yaml

shainw:
Adding in proper case statement for IPAddress from #2497 feedback.  Also added in the UserAgent as an output and entity mapping.

Fixes #

## Proposed Changes

  -
  -
  -


aprakash13 (approved):
Looks great to me. Thanks.

PR #2612, opened 2021-07-01
Fix ActorUserSession, removed unneeded fields from empty

oshezaf:
- ActorUserSessionId -> ActorSessionId
- Removed non documented fields from Empty parser.


oshezaf (commented):
Should have generated the empty template from the YAML to remove the extra fields, but will go ahead for now without it.

PR #2614, opened 2021-07-01
Fix CommandLine Issue

jannieli:
Fixes #

## Proposed Changes

  - CommandLine is missing in Event 4657, so it was removed from the query.
  - Checking the ObjectName for IFEO was added.

### DIFF
--- a/Hunting Queries/MultipleDataSources/PersistViaIFEORegistryKey.yaml
+++ b/Hunting Queries/MultipleDataSources/PersistViaIFEORegistryKey.yaml
@@ -23,7 +23,7 @@ query: |
   (
   SecurityEvent
   | where EventID == 4657
-  | where ObjectName has "Image File Execution Options"
+  | where ObjectName has_all ("\REGISTRY\MACHINE, "Image File Execution Options")
   | summarize Count=count() by Computer, Account, ObjectName
   | top 10 by Count desc 
   | extend AccountCustomEntity = Account, HostCustomEntity = Computer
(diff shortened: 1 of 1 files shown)


oshezaf (changes requested):
- This branch is not merged with recent changes to this query in master.
- The SecurityEvent and Sysmon variants are closer, but the SecurityEvent variant does not require "HKLM" which the Sysmon variant (and normalized variant do). Use has_all ("\\REGISTRY\\MACHINE, "Image File Execution Options")
- The parse_xml/bag_unpack/pivot method of parsing Sysmon is invalid (it would not return the original event count, but less) and will not work on more than a low volume of events performance wise. Refer to the new ASimRegistry Sysmon parser for how to efficiently parse in this case.

Hunting Queries/MultipleDataSources/PersistViaIFEORegistryKey.yaml
```diff
@@ -23,7 +23,7 @@ query: |
   (
   SecurityEvent
   | where EventID == 4657
-  | where ObjectName has "Image File Execution Options"
+  | where ObjectName has_all ("\REGISTRY\MACHINE, "Image File Execution Options")
```
oshezaf, line 26: Yyou need double back slashes: "\\REGISTRY\\MACHINE". Sorry if I mislead you. Anyways, just this fix and we are fine

oshezaf (changes requested):

Hunting Queries/MultipleDataSources/PersistViaIFEORegistryKey.yaml
jannieli: Added, thank you!

oshezaf (approved):

PR #2645, opened 2021-07-09
moved ADFSDBLocalSqlStatements detection rule to hunting query

Cyb3rWard0g:
I tested the detection rule ADFSDBLocalSqlStatements and it is best to use it as a hunting query.


shainw (changes requested):
Severity and requiredDataConnector and version need to be included.  Also, the entity mappings should stay in as we want these for future usage.

shainw (approved):

PR #2654, opened 2021-07-10
Issue 2652: Fixing issues with comparing LA tables to supported tables

anigan:
Fixes #
$nugetPackageLocation was never defined and is not required to be defined
$decision is used but is not passed as a parameter. It is required.


sreedharande (changes requested):
@anigan - Please add $nugetPackageLocation = "$($env:USERPROFILE)\.nuget\packages" under Static Variables Region

anigan:
@sreedharande This is done

PR #2666, opened 2021-07-13
Update Common_Deployed_Resources.yaml

shainw:
This was rolled back, need to change back to something closer to May 3rd PR.  This is updated to use new Value fields.

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Hunting Queries/AzureActivity/Common_Deployed_Resources.yaml
+++ b/Hunting Queries/AzureActivity/Common_Deployed_Resources.yaml
@@ -16,7 +16,7 @@ relevantTechniques:
 query: |
 
   AzureActivity
-  | where OperationNameValue has_any (@"deployments/write", @"virtualMachines/write")// or OperationNameVaue has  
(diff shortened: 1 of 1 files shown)


Hunting Queries/AzureActivity/Common_Deployed_Resources.yaml
```diff
@@ -16,7 +16,12 @@ relevantTechniques:
 query: |
 
   AzureActivity
+  | where OperationNameValue has_any (@"deployments/write", @"virtualMachines/write")// or OperationNameVaue has
```
ashwin-patil, line 19: may want to remove this comment as it looks incomplete -` // or OperationNameVaue has`

ashwin-patil (changes requested):

Hunting Queries/AzureActivity/Common_Deployed_Resources.yaml
shainw: Thanks, missed removing that

PR #2672, opened 2021-07-14
Cisco ISE Playbooks

vu-socprime:
## Proposed Changes

  - add Cisco ISE custom connector
  - add CiscoISE-FalsePositivesClearPolicies playbook
  - add CiscoISE-SuspendGuestUser playbook
  - add CiscoISE-TakeEndpointActionFromTeams playbook


lior-tamir:
Very nice playbooks! 

@vu-socprime, 
For the FalsePositivesClearPolicies, logo is missing from the comment
The Teams adaptive card seems with a long complicated experience (all choices are expended) can each choice use dropdowns which have default choices?

vu-socprime:
> For the FalsePositivesClearPolicies, logo is missing from the comment

@lior-tamir fixed

vu-socprime:
> The Teams adaptive card seems with a long complicated experience (all choices are expended) can each choice use dropdowns which have default choices?

@lior-tamir changed to dropdowns

v-jayakal:
@lior-tamir - Can you please review the fixes? Thanks.

lior-tamir:
Thanks for the changes. In the adaptive card, after the change, the actual host that the user is taking an action on is not shown?
Meaning, before the "take action", there should be given information on the host that user is going to take action on, so user will be able to take a decision.

PR #2705, opened 2021-07-19
Added pwsh.exe for PowerShell 7

anwather:
Fixes #

## Proposed Changes

  - Added pwsh.exe for PowerShell 7 to query


oshezaf (changes requested):
This hunting query should really also use the ASIM process parsers to broaden its applicability. Also, since FileName is a computed value, this may be very slow. Add 

where NewProcessName has_any ("powershell.exe","pwsh.exe")

to prequalify. It would boost performance.

jannieli (approved):
Consider Ofer's comment, but the addition of pwsh.exe is approved

shainw:
There have been changes since this, new PR with current master and these changes - #3500

PR #2712, opened 2021-07-19
Advanced Snow Integration Logic App updated to support new Incident Trigger

jingsta:
Fixes #

## Proposed Changes

- Migrated playbook to support incidents instead of alerts
- fixed some JSON parsing problems with adaptive cards by adding an compose action to render the card.


lior-tamir:
@jingsta,
The main playbook should be created with the exact playbookName parameter (line 175)
"name": "[concat(parameters('PlaybookName'),'-workflow-incident')]" should be changed to 
"name": "[parameters('PlaybookName')]",

In the description, would it be right to say "incident" instead of "alert"?

Regarding tags, no need to say names of connectors. We do should have a unique tag for "response from Teams" scenarios. I would suggest to change tags to:
["Sync", "Notification", "Teams response"]

sreedharande (approved):

PR #2717, opened 2021-07-21
AWS CloudTrail - Azure Native

sreedharande:
Fixes #

## Proposed Changes

  - Azure native Sentinel Data connector to ingest AWS CloudTrail Logs from centralized AWS S3 bucket - Receiving CloudTrail Log Files from Multiple Accounts
  -
  -


Kalakaarboyz1:
How about that.

PR #2733, opened 2021-07-25
Adding Parser and ParserVersion tags to detections

YaronFruchtmann:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/ASimDNS/imDNS_Miners.yaml
+++ b/Detections/ASimDNS/imDNS_Miners.yaml
@@ -19,7 +19,7 @@ tags:
   - Id: 0d76e9cf-788d-4a69-ac7d-f234826b5bed
   - version: 1.0.0
   - Schema: ASIMDns
-  - SchemaVersion: 1.0.0
+  - SchemaVersion: 0.1.1
 query: |
   union isfuzzy=true
   (datatable(TimeGenerated:datetime)[]),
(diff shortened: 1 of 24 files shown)


oshezaf (changes requested):
As discussed, the contract is schema and not parser, and those should be Schema and SchemaVersion. Note that Schema != Parser as we may use multiple source agnostic parsers per schema (ProcessCreate, ProcessTerminate). Use ASIMProcessEvent and the like.

Also I don't know if you don't need to update version numbers.

Detections/ASimDNS/imDNS_Miners.yaml
```diff
@@ -18,6 +18,8 @@ relevantTechniques:
 tags:
   - Id: 0d76e9cf-788d-4a69-ac7d-f234826b5bed
   - version: 1.0.0
+  - Schema: ASIMDns
```
oshezaf, line 21: Current schema version for ASIMDns is 0.1.1

oshezaf (changes requested):

Detections/ASimDNS/imDNS_Miners.yaml
YaronFruchtmann: Thanks. also fixed on the MS parser

oshezaf (approved):

PR #2742, opened 2021-07-27
Add validations on playbook template metadata

ityankel:
If playbook template contains metadata in the new schema (title, description, etc.) validate that the metadata is valid:
* Mandatory fields are provided
* Update time is a valid timestamp
* Entities are valid and from the list of supported entity types


preetikr:
@mpo-sec - Can we please review this from metadata arm api fields perspective? Thanks.

ityankel:
> @mpo-sec - Can we please review this from metadata arm api fields perspective? Thanks.

Note that this PR isn't related to metadata ARM API in any way. This only affects the metadata object in playbook templates which use a different schema. 
This PR will need to be reviewed by @lior-tamir who owns that area

mpo-sec (approved):

lior-tamir (approved):

PR #2767, opened 2021-07-30
Azure Sentinel Training Lab

javiersoriano:
Adding Azure Sentinel Training Lab content

## Proposed Changes

  -
  -
  -


preetikr:
Thanks, @javiersoriano for the contribution. 
@v-jayakal - can you please review the ARM template deploys for this? Thanks.

preetikr:
@javiersoriano - I had shared this on email but also sharing here so that tracked as part of overall feedback:

I looked at the content aspects and have a few observations:
1.	The data connectors focusses more on 1P data connectors only. Wondering if we could cover 3P data connectors as well at least a type of the following. 
a.	Syslog
b.	CEF
c.	Azure Functions 
d.	REST API Push
e.	File (using syslog agent to custom tables)

2.	I also observed that workbooks are interspersed in other modules. It’d be great to cover a module on workbooks and how can customer view and edit queries. 

Thanks!

PR #2789, opened 2021-08-04
Case sensitivity in JOIN operation leading to unexpected results

MahanteshHiremath:
When I run the query in my lab, here's what I found.

Output from the temp table AC_Add has two rows.
--------------------------------------------------------
AccountAdded_GroupAddedTo_AddingAccount
cn=Joe Pesci||Domain Admins||contoso\labadmin
CN=Joe Pesci||Administrators||contoso\labadmin

Notice how the string 'cn' is in lower case in the case of user addition to 'Domain Admins'

Output from the temp table AC_Remove has two rows.
------------------------------------------------------------
AccountRemoved_GroupRemovedFrom_RemovingAccount
CN=Joe Pesci||Domain Admins||contoso\labadmin
CN=Joe Pesci||Administrators||contoso\labadmin

Notice how the string 'CN' is in upper case in the case of user removal from 'Domain Admins'

I expected the query (when run as a whole) to show 2 results. Rather, output contained only 1 row.
------------------------------------------------------------------------------------------------------------
AccountAdded_GroupAddedTo_AddingAccount
CN=Joe Pesci||Administrators||contoso\labadmin

This is because of the case sensitivity which must be accounted for before we do the join operation.

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/SecurityEvent/UserAccountAdd-Removed.yaml
+++ b/Detections/SecurityEvent/UserAccountAdd-Removed.yaml
@@ -32,7 +32,7 @@ query: |
   | parse EventData with * '"MemberName">' AccountAdded ",OU" * 
   | where isnotempty(AccountAdded)
   | extend GroupAddedTo = TargetUserName, AddingAccount = Account 
-  | extend  AccountAdded_GroupAddedTo_AddingAccount = strcat(AccountAdded, "||", GroupAddedTo, "||", AddingAccount )
+  | extend  AccountAdded_GroupAddedTo_AddingAccount = tolower(strcat(AccountAdded, "||", GroupAddedTo, "||", AddingAccount ))
   | project AccountAdded_GroupAddedTo_AddingAccount, AccountAddedTime = TimeGenerated;
   let AC_Remove = 
   SecurityEvent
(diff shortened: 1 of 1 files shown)


aprakash13:
Thanks Mhantesh for highlighting the issue.

Adding tolower is great however it would probably also change the case of the commonName(CN) for those names which are upper case. Was hoping to address the issue without it. It appears that the  issue seems to happen because of the lower/upper case 'CN' when parsing for AccountAdded/AccountRemoved field. How about we just modify the query like below to address this issue i.e.

from:
| parse EventData with * '"MemberName">' AccountAdded ",OU" * 

to:
 | parse EventData with * '"MemberName">' * '=' AccountAdded ",OU" * 

We can do something similar to AccountRemoved field as well. This should probably solve the case sensitivity issue. Can you please check in your environment if that works and I will update the query. Thanks again @MahanteshHiremath

aprakash13:
Created the below PR to address the change:  https://github.com/Azure/Azure-Sentinel/pull/2854

aprakash13:
Changes merged. Closing this out.

PR #2790, opened 2021-08-04
SimpleMemberName should account for case sensitivity

MahanteshHiremath:
The definition of SimpleMemberName in its present form fails to account for distinguished names containing either the member name or the string 'CN' in lower case.

extend SimpleMemberName = tostring(split(tostring(split(MemberName, ",")[0]),"CN=")[1])

For instance - event ID 4728 was logged in my lab for adding a user to Domain Admins group. However, the string 'CN' was in lower case in the distinguished name for the member name. As a result, the columns like SimpleMemberName, AccountCustomEntity were empty in the output corresponding to this event.

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/SecurityEvent/UserAccountAddedToPrivlegeGroup_1h.yaml
+++ b/Detections/SecurityEvent/UserAccountAddedToPrivlegeGroup_1h.yaml
@@ -36,7 +36,7 @@ query: |
   | where TargetSid matches regex WellKnownLocalSID or TargetSid matches regex WellKnownGroupSID
   // Exclude Remote Desktop Users group: S-1-5-32-555
   | where TargetSid !in ("S-1-5-32-555")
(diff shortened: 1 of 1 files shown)


aprakash13:
Thanks Mhantesh for bringing this up. 

Adding tolower is great however it would also change the case of the commonName(CN) for those names which are upper case. How about we just remove the CN from the query to address this issue i.e.

from:
| extend SimpleMemberName = tostring(split(tostring(split(MemberName, ",")[0]),"**CN=**")[1])

to:
| extend SimpleMemberName = tostring(split(tostring(split(MemberName, ",")[0]),"**=**")[1])

This should ideally solve the case sensitivity issue. Can you please check in your environment if that works and I will update the query. Thanks again Mahantesh.

aprakash13:
Created the below PR to address the change:  
https://github.com/Azure/Azure-Sentinel/pull/2854

aprakash13:
Changes merged. Closing this out.

PR #2801, opened 2021-08-09
Create ProxyShellPwn2Own.yml

thmcelro:
Creation of a new detection to identify likely ProxyShell Pwn2Own exploit attempts targeting "autodiscover/autodiscover.json".

Reference: https://peterjson.medium.com/reproducing-the-proxyshell-pwn2own-exploit-49743a4ea9a1

This detection looks for 4 data points observed in use alongside the exploit:

1. Protocol parameter missing from AutoDiscover requests is unusual and in testing, wholly malicious
2. "/mapi/" or the "powershell" endpoint being called via AutoDiscover in testing is wholly malicious
3. Pwn2Own is often formatted to have a forward slash "/" following an email address "/" 
4. Some Pwn2Own attempts have included a port number followed by a "/"


Detections/W3CIISLog/ProxyShellPwn2Own.yml
```diff
@@ -0,0 +1,42 @@
+id: 968358d6-6af8-49bb-aaa4-187b3067fb95
+name: Exchange ProxyShell Pwn2Own
```
shainw: We are looking to standardize on naming convention a bit going forward.  We are not finalized on the naming convention yet, but it may be something close to this pattern.  Can you see about adjusting a bit, if it makes sense, to map to this on the name value please - {Activity|Operation} {Filter|Pattern} {Context|Entity} {-|:|[]} {Hunting|Detection}.  Just an example, so tweak as you see firt, but maybe use something like - "Exchange SSRF Autodiscover ProxyShell - Detection".

Detections/W3CIISLog/ProxyShellPwn2Own.yml
```diff
@@ -0,0 +1,42 @@
...
+description: |
+  'This query looks for suspicious request patterns to Exchange servers that fit patterns recently
```
shainw: Any chance you can put a bit more in the description here.  For example, maybe the chain of high level events indicated in the blog?

Detections/W3CIISLog/ProxyShellPwn2Own.yml
```diff
@@ -0,0 +1,42 @@
...
+  blogged about by PeterJson.
+  Reference: https://peterjson.medium.com/reproducing-the-proxyshell-pwn2own-exploit-49743a4ea9a1'
+severity: Medium
```
shainw: Seems like this is High based on your comments in the PR submission?

Detections/W3CIISLog/ProxyShellPwn2Own.yml
```diff
@@ -0,0 +1,42 @@
...
+queryPeriod: 14d
+triggerOperator: gt
+triggerThreshold: 0
+tactics:
+  - InitialAccess
+relevantTechniques:
+  - T1190
+query: |
+  let successCodes = dynamic([200, 302]);
+  let timeRange = 14d;
```
shainw: You can remove the query period will control this.

Detections/W3CIISLog/ProxyShellPwn2Own.yml
```diff
@@ -0,0 +1,42 @@
...
+  'This query looks for suspicious request patterns to Exchange servers that fit patterns recently
+  blogged about by PeterJson.
+  Reference: https://peterjson.medium.com/reproducing-the-proxyshell-pwn2own-exploit-49743a4ea9a1'
+severity: Medium
+requiredDataConnectors:
+  - connectorId: AzureMonitor(IIS)
+    dataTypes:
+      - W3CIISLog
+queryFrequency: 1d
+queryPeriod: 14d
```
shainw: This is a detection, I don't recommend we look back 14 days on a daily run.  Can you change this to 1d?  Or maybe even more often if we think this is High, could be 6 hours or less.

Detections/W3CIISLog/ProxyShellPwn2Own.yml
```diff
@@ -0,0 +1,42 @@
...
+  | where TimeGenerated > ago(timeRange)
+  | where csUriStem startswith "/autodiscover/autodiscover.json"
+  | where scStatus has_any (successCodes)
+  | project TimeGenerated, cIP, sIP, sSiteName, csUriStem, csUriQuery
+  | where (csUriQuery !has "Protocol" and isnotempty(csUriQuery))
+  or (csUriQuery has_any("/mapi/", "powershell"))
+  or (csUriQuery contains "@" and csUriQuery matches regex @"\.[a-zA-Z]{2,4}?(?:[a-zA-Z]{2,4}\/)")
+  or (csUriQuery contains ":" and csUriQuery matches regex @"\:[0-9]{2,4}\/")
+  | extend timestamp = TimeGenerated, HostCustomEntity = Computer, IPCustomEntity = cIP
+entityMappings:
```
shainw: I would suggest you bring thru User from csUserName, Azure Resource from ResourceId_ and maybe File from FileUri:  when available - [Entities](https://docs.microsoft.com/en-us/azure/sentinel/entities-reference).  We are limited to 5 entitityTypes right now.

shainw (changes requested):

Detections/W3CIISLog/ProxyShellPwn2Own.yaml
```diff
@@ -0,0 +1,42 @@
...
+queryFrequency: 1d
+queryPeriod: 14d
+triggerOperator: gt
+triggerThreshold: 0
+tactics:
+  - InitialAccess
+relevantTechniques:
+  - T1190
+query: |
+  let successCodes = dynamic([200, 302]);
```
ashwin-patil, line 21: I also saw 401 status for SSRF into “/powershell endpoint stage in the blog. 
Should we include that to this status list ?

ashwin-patil (changes requested):

Detections/W3CIISLog/ProxyShellPwn2Own.yml
thmcelro: Updated in line with suggested convention

Detections/W3CIISLog/ProxyShellPwn2Own.yml
thmcelro: Improved description to include more detail

Detections/W3CIISLog/ProxyShellPwn2Own.yml
thmcelro: Increased to high severity, agreed.

Detections/W3CIISLog/ProxyShellPwn2Own.yml
thmcelro: Removed

Detections/W3CIISLog/ProxyShellPwn2Own.yml
thmcelro: Suggested fields brought through

Detections/W3CIISLog/ProxyShellPwn2Own.yaml
thmcelro: I've added that status code in. In all honesty the detection is tight enough we could remove the status codes, however we'd be unable to filter out attempts vs. success. I'll use 200, 302 and 401 for now but if we start seeing lots of status code diversity we can consider removing this check.

Detections/W3CIISLog/ProxyShellPwn2Own.yml
thmcelro: Updated to run every 12 hours

shainw (approved):

PR #2806, opened 2021-08-10
ADX Health Playbook 

mariavaladas:
Fixes #

## Proposed Changes

  -
  -
  -


mariavaladas:
I have created a playbook that is complementary to this workbook by Naomi https://github.com/Azure/Azure-Sentinel/blob/master/Workbooks/ADXvsLA.json - we will create a blog post to explain how to use both. The playbook looks back and compares the number of logs in ADX vs LA tables. It sends an email if a delay is detected.

mariavaladas:
I have added the Deployment section that was missing.

PR #2813, opened 2021-08-11
Yaml2ARM_lite

YaronFruchtmann:
Fixes #

## Proposed Changes
Introducing lite converter version: given yaml file produces ARM template file


oshezaf (changes requested):
- Let's call it ASimYaml2Arm. Forget the light, we will find a a name for the big one if we ever release it.
- Add README.md, point out to the python modules that one needs to install.

oshezaf (approved):

PR #2836, opened 2021-08-13
Updating Azure Purview workbook based on results from bug bash

libarson:
Fixes #

## Proposed Changes

  -
  -
  -


v-rucdu (changes requested):
Please update the version of AzurePurview workbook in workbookmetadata.json so the updates are deployed.

e.g. Update version from 1.0 to something like 1.1 or 2.0

libarson:
The version has been updated in the workbook metadata.

v-rucdu (approved):

PR #2844, opened 2021-08-16
Create checkThatTemplatesVersionWasChanged.sh

Amitbergman:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/.script/checkThatTemplatesVersionWasChanged.sh
+++ b/.script/checkThatTemplatesVersionWasChanged.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-var=$(echo $(git diff origin/main -U0))
+var=$(echo $(git diff origin/master -U0))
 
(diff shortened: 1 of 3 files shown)


.script/checkThatTemplatesVersionWasChanged.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+var=$(echo $(git diff origin/master -U0))
+
+if [[ "$var" == *"version"* ]];
+then
+    echo "Changed the version of the template, nice!"
+    exit 0
+else
+    echo "Hey there, you did not change the version of the template, please update the version of the template you changed."
```
laithhisham, line 10: just to make the comment a bit more clear, maybe use something like
"""
There are one or more templates that were modified and their version was not updated. Please update the version of the template(s) you changed.
"""

laithhisham (approved):

PR #2864, opened 2021-08-17
Delete CiscoASA playbook from playbook folder

v-maudan:
Fixes #

## Proposed Changes

  -
  -
  -


v-jayakal:
As per earlier comments from Lior, closing this PR.
thoughts are:
•	Until we have the “partial deployment” feature in solution (to deploy only playbooks for example, or only a specific playbook), I would recommend to keep the content available from GitHub.
•	For the “deploy the whole package” (playbooks+connectors):
When a solution is published, I would add to the readme file “deploy the whole package from {productname} solution” with a link to the product instead of the linked template which you have taken off.
Until the solution is available to customers, I would keep the full deployment option in Github.

PR #2883, opened 2021-08-20
ZeroTrust(TIC3.0) (Preview)

thbanasi:
Please publish as (Preview) in Azure Marketplace during our Private Preview. Feel free to reach out if you need anything, thank you

Fixes #

## Proposed Changes

  -
  -
  -


NikTripathi:
This file already exists in master branch with same name and same codes. If you want some changes I think you should move the file so that it gets deleted from master and then we will push it to solution. Otherwise it will create a copy.

thbanasi:
Yes, these are being migrated from Workbooks to Solutions. Upon GA, I'll depreciate the Workbooks upon GA

v-rucdu (approved):

PR #2885, opened 2021-08-20
MITREATT&CK

thbanasi:
MITRE ATT&CK for Cloud Matrix

Fixes #

## Proposed Changes

  -
  -
  -


NikTripathi:
This file already exists in master branch with same name. If you want some changes I think you should move the file so that it gets deleted from master and then we will push it to solution. Otherwise it will create a copy.

thbanasi:
Yes, these are being migrated from Workbooks to Solutions. Upon GA, I'll depreciate the Workbooks upon GA

PR #2897, opened 2021-08-25
detecting access to ADHealth Service Agent registry keys

Cyb3rWard0g:
This detection uses Windows security events to detect suspicious access attempts to the registry key values and sub-keys of Azure AD Health service agents (e.g AD FS).  Information from AD Health service agents can be used to potentially abuse some of the features provided by those services in the cloud (e.g. Federation).

Example: Spoofing AD FS signing logs
https://o365blog.com/post/hybridhealthagent/

This detection requires an access control entry (ACE) on the system access control list (SACL) of the following securable object: HKLM:\SOFTWARE\Microsoft\ADHealthAgent. Make sure you set the SACL to propagate to its sub-keys. You can find more information in here https://github.com/OTRF/Set-AuditRule/blob/master/rules/registry/aad_connect_health_service_agent.yml

### DIFF
--- a/Detections/SecurityEvent/AADHealthSvcAgentRegKeyAccess.yaml
+++ b/Detections/SecurityEvent/AADHealthSvcAgentRegKeyAccess.yaml
@@ -1,0 +1,80 @@
+id: 06bbf969-fcbe-43fa-bac2-b2fa131d113a
+name: Azure AD Health Service Agents Registry Keys Access
+description: |
+  'This detection uses Windows security events to detect suspicious access attempts to the registry key values and sub-keys of Azure AD Health service agents (e.g AD FS).
+  Information from AD Health service agents can be used to potentially abuse some of the features provided by those services in the cloud (e.g. Federation).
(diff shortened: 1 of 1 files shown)


Detections/SecurityEvent/AADHealthSvcAgentRegKeyAccess.yaml
```diff
@@ -0,0 +1,80 @@
...
+  // Filter out known processes
+  let aadConnectHealthProcs = dynamic ([
+      'Microsoft.Identity.Health.Adfs.DiagnosticsAgent.exe',
+      'Microsoft.Identity.Health.Adfs.InsightsService.exe',
+      'Microsoft.Identity.Health.Adfs.MonitoringAgent.Startup.exe',
+      'Microsoft.Identity.Health.Adfs.PshSurrogate.exe',
+      'Microsoft.Identity.Health.Common.Clients.ResourceMonitor.exe'
+  ]);
+  // Adjust this to adjust the key export detection  timeframe
+  //let timeframe = 1d;
```
petebryan, line 36: You might want to leave this out so that if doesn't accidently overwrite timeframes set during analytics creation. If someone sets that to > 1d but doesn't change this then it will still only be 1d.

petebryan (changes requested):

PR #2898, opened 2021-08-25
detecting access to the ADhealth monitoring agent registry key

Cyb3rWard0g:
This detection uses Windows security events to detect suspicious access attempts to the registry key of Azure AD Health monitoring agent.

Information such as the machine identity is retrieved from this key.

Example: Spoofing AD FS signing logs
https://o365blog.com/post/hybridhealthagent/

This detection requires an access control entry (ACE) on the system access control list (SACL) of the following securable object HKLM\SOFTWARE\Microsoft\Microsoft Online\Reporting\MonitoringAgent. You can find more information in here https://github.com/OTRF/Set-AuditRule/blob/master/rules/registry/aad_connect_health_monitoring_agent.yml

### DIFF
--- a/Detections/SecurityEvent/AADHealthMonAgentRegKeyAccess.yaml
+++ b/Detections/SecurityEvent/AADHealthMonAgentRegKeyAccess.yaml
@@ -1,0 +1,80 @@
+id: f819c592-c5f9-4d5c-a79f-1e6819863533
+name: Azure AD Health Monitoring Agent Registry Keys Access
+description: |
+  'This detection uses Windows security events to detect suspicious access attempts to the registry key of Azure AD Health monitoring agent.
+  This detection requires an access control entry (ACE) on the system access control list (SACL) of the following securable object HKLM\SOFTWARE\Microsoft\Microsoft Online\Reporting\MonitoringAgent.
+  You can find more information in here https://github.com/OTRF/Set-AuditRule/blob/master/rules/registry/aad_connect_health_monitoring_agent.yml
+  '
+severity: Medium
+requiredDataConnectors:
+  - connectorId: SecurityEvents
+    dataTypes:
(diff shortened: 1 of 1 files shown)


Detections/SecurityEvent/AADHealthMonAgentRegKeyAccess.yaml
```diff
@@ -0,0 +1,80 @@
...
+  // Filter out known processes
+  let aadConnectHealthProcs = dynamic ([
+      'Microsoft.Identity.Health.Adfs.DiagnosticsAgent.exe',
+      'Microsoft.Identity.Health.Adfs.InsightsService.exe',
+      'Microsoft.Identity.Health.Adfs.MonitoringAgent.Startup.exe',
+      'Microsoft.Identity.Health.Adfs.PshSurrogate.exe',
+      'Microsoft.Identity.Health.Common.Clients.ResourceMonitor.exe'
+  ]);
+  // Adjust this to adjust the key export detection  timeframe
+  //let timeframe = 1d;
```
petebryan, line 35: Same with this one, you probably don't want this here.

Detections/SecurityEvent/AADHealthMonAgentRegKeyAccess.yaml
```diff
@@ -0,0 +1,80 @@
...
+entityMappings:
+  - entityType: Account
+    fieldMappings:
+      - identifier: FullName
+        columnName: AccountCustomEntity
+  - entityType: Host
+    fieldMappings:
+      - identifier: FullName
+        columnName: HostCustomEntity
+version: 2.0.0
```
petebryan, line 80: Its a new query - why is it 2.0.0?

petebryan (changes requested):

PR #2902, opened 2021-08-25
Updated legacy auth hunting query

petebryan:
Updated legacy auth hunting query with customer feedback regarding mis-match in legacy auth vs modern auth details.


aprakash13 (approved):
Will let Shainw merge this.

PR #2927, opened 2021-08-30
Update MultiplePasswordresetsbyUser.yaml

Amnetka:
In PerUserReset pivot Account field shows Actor account.
In TotalUserReset it just takes random one using arg_max which is very confusing, makes you think that this Account was responsible for all these aggregated resets. I propose just leaving this field blank for TotalUserReset, not to introduce confusion.

Fixes #

## Proposed Changes

  -
  -
  -


shainw (approved):
In testing, I don't see that this makes a big impact on result set, so approving.

PR #2938, opened 2021-08-31
Added UnifiSG Image back

v-rucdu:
Fixes #

## Proposed Changes

  -
  -
  -


v-rucdu:
The image got deleted, which caused WorkbookMetadata tests to fail. Added the image back from History

PR #2945, opened 2021-08-31
Adding the Az.EventHub module

piaudonn:
This is a dependincy for the New-AzEventHubNamespace cmdLet used in the New-EventHubNamespace function.
This fonction will fail if that module isn't installed.

## Proposed Changes

  - Adding Get-RequiredModules("Az.EventHub")


sreedharande:
@piaudonn - Please update package (.zip) with the fix

sreedharande (approved):

PR #2956, opened 2021-09-01
Change the location used to create the data connection

piaudonn:
The current version is using the location of the log analytic workspace. That works only if the ADX cluster is also in the same location.
For this to work in cases where the ADX cluster is in a different location as the log analytic workspace we need to provide the ADX's location.
I suggest to extract the location from $ADXClusterURL and provide it in the location parameter when swending the PUT to create the data connection.

Fixes #
Current version is returning a 400 when attemping to create a data connection on an ADX cluster is the ADX cluster location is different than the log analytic workspace location

## Proposed Changes

  - Extract the location of the ADX cluster this way: $ADXClusterURL.Split(".")[1] and save it as $ADXClusterName
  - Use $ADXClusterName in the location parameter sent in the PUT request


piaudonn:
@sreedharande doesn't seem to be the case. I have my ALA in Canada Central and my ADX in East US and that works fine.

PR #2957, opened 2021-09-01
Updating conenector values and fixing Tactics

ashwin-patil:
Fixes #

## Proposed Changes

  - Updating connector values to match with other ASIM files in same folder
  -  updating Tactics to standard values to align with others


shainw (commented):
Approved from perspective of Tactic updates, lets get review from ASIM owners on changes to connectors.

Hunting Queries/ASimProcess/imProcess_ExchangePowerShellSnapin.yaml
```diff
@@ -5,17 +5,20 @@ description: |
   Whilst this is a legitimate administrative tool it is abused by attackers to performs actions on a compromised
   Exchange server. Hunt for unusual activity related to this Snapin including it being added on new hosts or by
   new accounts.'
-requiredDataConnectors: []
+requiredDataConnectors:
```
oshezaf, line 8: When only ASIM sources are involved, at this time, the data connectors list should be empty. 

To explain: 
- "required" is actually always only a recommendation. No actual enforcement. It is mostly for the UX.
- In this case, imProcessCreate supports MDE, Sysmon and potentially others (WindowsEvents coming soon), so SecurityEvent is not the right value to show. We (ab)used the detection name to specify the require source, until a better solution is available.

Hunting Queries/ASimProcess/imProcess_HostExportingMailboxAndRemovingExport.yaml
```diff
@@ -16,19 +19,19 @@ tags:
   - NOBELIUM
 query: |
 
-     // Adjust the timeframe to change the window events need to occur within to alert
```
oshezaf, line 19: @YaronFruchtmann : note that the team prefers to maintain uniformity around indents. I am with them on that :-). So let's try.

oshezaf (changes requested):

shainw:
@ashwin-patil - let's get an update on this.

Hunting Queries/ASimProcess/imProcess_ExchangePowerShellSnapin.yaml
ashwin-patil: reverted connector changes to original

ashwin-patil:
> @ashwin-patil - let's get an update on this.

Updated. resolved conflicts, reverting ASIM connector changes

yaronMSFT (approved):

shainw (approved):

PR #2980, opened 2021-09-08
Dns/normalization/dns params content

YaronFruchtmann:
Fixes #

## Proposed Changes

  -
  -
  -


Detections/ASimDNS/imDns_DomainEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,66 @@
...
+        imDns(starttime=ago(dt_lookBack))
+        //Extract domain patterns from syslog message
+        | where isnotempty(DnsQuery)
+        //Split out the TLD
+        | extend parts = split(DnsQuery, '.')
+        //Validate parsed domain by checking if the TLD is in the list of TLDs in our threat feed
+        | extend tld = split(DnsQuery, '.')[-1]
+        | extend DNS_TimeGenerated = TimeGenerated
+  ) on $left.DomainName==$right.DnsQuery
+  | where DNS_TimeGenerated >= TimeGenerated and DNS_TimeGenerated < ExpirationDateTime
```
oshezaf, line 49: Why repeat the "TimeGenerated" test part (i.e. the first test)?

Detections/ASimDNS/imDns_DomainEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,66 @@
...
+  | summarize LatestIndicatorTime = arg_max(TimeGenerated, *) by IndicatorId
```
oshezaf, line 50: This has to be before the join. Doesn't make sense otherwise.

Detections/ASimDNS/imDns_DomainEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,66 @@
...
+tags:
+  - Id: 85aca4d1-5d15-4001-abd9-acb86ca1786a
+    version: 1.0.0
+  - Schema: ASIMDns
+    SchemaVersion: 0.1.1
+query: |
+  let dt_lookBack = 1h;
+  let ioc_lookBack = 14d;
+  //Create a list of TLDs in our threat feed for later validation
+  let list_tlds = ThreatIntelligenceIndicator
```
oshezaf, line 29: Why did you keep those. You don't use them as far as I can tell.

Detections/ASimDNS/imDns_DomainEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,66 @@
...
+        | extend parts = split(DnsQuery, '.')
+        //Validate parsed domain by checking if the TLD is in the list of TLDs in our threat feed
+        | extend tld = split(DnsQuery, '.')[-1]
+        | extend DNS_TimeGenerated = TimeGenerated
+  ) on $left.DomainName==$right.DnsQuery
+  | where DNS_TimeGenerated >= TimeGenerated and DNS_TimeGenerated < ExpirationDateTime
+  | summarize LatestIndicatorTime = arg_max(TimeGenerated, *) by IndicatorId
+  | project LatestIndicatorTime, Description, ActivityGroupNames, IndicatorId, ThreatType, ExpirationDateTime, ConfidenceScore, Url, DNS_TimeGenerated, Dvc, SrcIpAddr, DnsQuery, QueryType
+  | extend timestamp = DNS_TimeGenerated, HostCustomEntity = Dvc, IPCustomEntity = SrcIpAddr, URLCustomEntity = Url
+entityMappings:
```
oshezaf, line 53: Let's add entities, custom fields and dynamic values while at it. The first two are easy. The latest we should discuss.

Detections/ASimDNS/imDns_IPEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,67 @@
...
+  - Schema: ASIMDns
+    SchemaVersion: 0.1.1
+query: |
+  let dt_lookBack = 1h;
+  let ioc_lookBack = 14d;
+  let IP_TI = (ThreatIntelligenceIndicator
+  | where TimeGenerated >= ago(ioc_lookBack) and ExpirationDateTime > now()
+  | where Active == true
+  // Picking up only IOC's that contain the entities we want
+  | where isnotempty(NetworkIP) or isnotempty(EmailSourceIpAddress) or isnotempty(NetworkDestinationIP) or isnotempty(NetworkSourceIP)
```
oshezaf, line 35: Do you need this check at all?

Detections/ASimDNS/imDns_IPEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,67 @@
...
+  // As there is potentially more than 1 indicator type for matching IP, taking NetworkIP first, then others if that is empty.
+  // Taking the first non-empty value based on potential IOC match availability
+  | extend TI_ipEntity = coalesce(NetworkIP, NetworkDestinationIP, NetworkSourceIP,EmailSourceIpAddress,"")
+  | summarize LatestIndicatorTime = arg_max(TimeGenerated, *) by IndicatorId, TI_ipEntity);
```
oshezaf, line 39: No need to summarize by TI_ipEntity (which also slow things a lot).

Detections/ASimDNS/imDns_IPEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,67 @@
...
+  let TI_IP_List=IP_TI | summarize make_set( TI_ipEntity);
+  imDns(starttime=ago(dt_lookBack), response_has_any=todynamic(toscalar(TI_IP_List)))
```
oshezaf, line 41: We changed to response_has_any_prefix? didn't we? Also, using things, I feel we need to change to response_has_any_ipv4_prefix and response_has_ipv4 respectively.

Detections/ASimDNS/imDns_IPEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,67 @@
...
+      | extend tilist = toscalar(TI_IP_List)
```
oshezaf, line 42: I am not sure I follow why do you need this post-filtering exercise with mv-expand.

Detections/ASimDNS/imDns_IPEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,67 @@
...
+      | mv-expand tilist
+      | extend SingleIP=tostring(tilist)
+      | project-away tilist
+      | where has_ipv4(DnsResponseName, SingleIP)
+      | extend DNS_TimeGenerated = TimeGenerated
+  | join IP_TI
+        on $left.SingleIP == $right.TI_ipEntity
+  | where DNS_TimeGenerated >= TimeGenerated and DNS_TimeGenerated < ExpirationDateTime
```
oshezaf, line 50: The first test, as in the previous parser, may not be needed.

Detections/ASimDNS/imDns_IPEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,67 @@
...
+  | project LatestIndicatorTime, Description, ActivityGroupNames, IndicatorId, ThreatType, Url, ExpirationDateTime, ConfidenceScore, DNS_TimeGenerated,
+  TI_ipEntity, Dvc, EventId, SubType, SrcIpAddr, DnsQuery, DnsResponseName, NetworkIP, NetworkDestinationIP, NetworkSourceIP, EmailSourceIpAddress
+  | extend timestamp = DNS_TimeGenerated, IPCustomEntity = SrcIpAddr, HostCustomEntity = Dvc, URLCustomEntity = Url
+entityMappings:
```
oshezaf, line 54: See note above about additional entities, custom fields and dynamic alerts

oshezaf (changes requested):

Detections/ASimDNS/imDns_DomainEntity_DnsEvents.yaml
YaronFruchtmann: It is not the same TimeGenerated. This one comes from the TI table, and line 49 verifies that the event timestamp  (DNS_TimeGenerated) is within the validity  timeframe of the TI: after it was marked as suspicious and before it expired

oshezaf: Actually, it is a bad bug in the original query. Since the refresh the table every 14 days, TimeGenerated is not a correct value to compare to for the "start of life" of the IoC. They have a bug, that in the refresh they don't maintain the original TimeGenerated in any field, however, in practice. If there is a new IoC and an old event, you still want to alert on it, so I would remove this check.

Detections/ASimDNS/imDns_DomainEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,69 @@
+id: 999e9f5d-db4a-4b07-a206-29c4e667b7e8
+name: (Preview) TI map Domain entity to Dns Event (Normalized DNS)
+description: |
+  'Identifies a match in DNS events from any Domain IOC from TI
```
oshezaf, line 4: Do we need the single quotes? does it work in multi-line? Also, need to end sentences with a full stop (".")

Detections/ASimDNS/imDns_DomainEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,69 @@
...
+    dataTypes:
+      - ThreatIntelligenceIndicator
+queryFrequency: 1h
+queryPeriod: 14d
+triggerOperator: gt
+triggerThreshold: 0
+tactics:
+  - Impact
+tags:
+  - Id: 85aca4d1-5d15-4001-abd9-acb86ca1786a
```
oshezaf, line 21: Didn't we decide to name it differently (ParentId or so)?

Detections/ASimDNS/imDns_DomainEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,69 @@
...
+    | where TimeGenerated >= ago(ioc_lookBack) and ExpirationDateTime > now()
+    // Picking up only IOC's that contain the entities we want
+    | where isnotempty(DomainName)
+    | where Active == true
+    | summarize LatestIndicatorTime = arg_max(TimeGenerated, *) by IndicatorId;
+  let Domains= toscalar(DomainTIs |summarize make_set(DomainName));
+  DomainTIs
+    | join (
+        imDns(starttime=ago(dt_lookBack), domain_has_any=(Domains))
+        | where isnotempty(DnsQuery)
```
oshezaf, line 38: Is this possible? Doesn't domain_has_any need to take care of this?

Detections/ASimDNS/imDns_DomainEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,69 @@
...
+  Description: Description
+  ActivityGroupNames: ActivityGroupNames
+  IndicatorId: IndicatorId
+  ThreatType: ThreatType
+  ExpirationDateTime: ExpirationDateTime
+  ConfidenceScore: ConfidenceScore
+  DNSRequestTime: DNS_TimeGenerated
+  SourceIPAddress: SrcIpAddr
+  DnsQuery: DnsQuery
+  QuertYpe: QueryType
```
oshezaf, line 68: Capitalization

Detections/ASimDNS/imDns_IPEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,77 @@
...
+  | summarize LatestIndicatorTime = arg_max(TimeGenerated, *) by IndicatorId);
+  let TI_IP_List=IP_TI | summarize make_set( TI_ipEntity);
+  imDns(starttime=ago(dt_lookBack), response_has_any_prefix=todynamic(toscalar(TI_IP_List)))
+      | extend tilist = toscalar(TI_IP_List)
+      | mv-expand tilist
+      | extend SingleIP=tostring(tilist)
+      | project-away tilist
+      | where has_ipv4(DnsResponseName, SingleIP)
+      | extend DNS_TimeGenerated = TimeGenerated
+  | join IP_TI
```
oshezaf, line 45: Don't change, but I wonder if it make sense to reverse the join to the same order used in the domain query. The thing is that the IoC list - while it can be shorter or longer then the filtered event list, is more predictable.

Detections/ASimDNS/imDns_IPEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,77 @@
...
+        columnName: HostCustomEntity
+  - entityType: IP
+    fieldMappings:
+      - identifier: Address
+        columnName: IPCustomEntity
+  - entityType: URL
+    fieldMappings:
+      - identifier: Url
+        columnName: URLCustomEntity
+customDetails:
```
oshezaf, line 64: I am adding a task to add dynamic alert details to our alert rules. Not now.

Detections/ASimDNS/imDns_IPEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,77 @@
+id: 67775878-7f8b-4380-ac54-115e1e828901
```
oshezaf, line 1: Some of the comments from the domain query apply.

oshezaf (changes requested):

Detections/ASimDNS/imDns_DomainEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,69 @@
+id: 999e9f5d-db4a-4b07-a206-29c4e667b7e8
+name: (Preview) TI map Domain entity to Dns Event (Normalized DNS)
```
oshezaf, line 2: "Dns Event" -> "DNS Events"

Detections/ASimDNS/imDns_IPEntity_DnsEvents.yaml
```diff
@@ -0,0 +1,77 @@
...
+name: (Preview) TI map IP entity to DnsEvents (Normalized DNS)
```
oshezaf, line 2: "DnsEvents" -> "DNS Events"

oshezaf (changes requested):

oshezaf (approved):

PR #2867, opened 2021-08-17
Multiple Query fixes

ashwin-patil:
Fixes #

## Proposed Changes

  -  Fixing logic `OperationNameValue` changes in multiple queries and workbooks where incorrectly replaced in last PR.
  - Replacing deprecated `parsejson` with `parse_json` and `makset` with `make_set`

PR #2982, opened 2021-09-09
Reedit RareOperations.yaml to make it functional

### LINKED ISSUE #2976: Template regarding "Rare subscription-level operations in Azure" may not useful
**Describe the bug**
The script is here: https://github.com/Azure/Azure-Sentinel/blob/2cad1a602c99d6e3f8be2548e31e4ca63ed75c6f/Detections/AzureActivity/RareOperations.yaml

I believe this script will never return any result.

For the value of the ‘OperationNameValue’, it looks like this:
image

Thus, when we define SensitiveOperationList to "List keys", "List Storage Account Keys", "Register Subscription", "Create or Update Snapshot", "Create or Update Network Security Group", they will never meet the value of ‘OperationNameValue’:
image

However, when I switch one of those parameters inside the list:
image

Then, there will be some results that can be met.

**Expected behavior**
Reedit the template to make it functional.

Lodewyk-Git:
Changed OperationNameValue to OperationName

Fixes #2976

### DIFF
--- a/Detections/AzureActivity/RareOperations.yaml
+++ b/Detections/AzureActivity/RareOperations.yaml
@@ -28,7 +28,7 @@ query: |
   let SensitiveOperationList = dynamic(
   ["List keys", "List Storage Account Keys", "Register Subscription", "Create or Update Snapshot", "Create or Update Network Security Group"]);
   let SensitiveActivity = AzureActivity
-  | where OperationNameValue in~ (SensitiveOperationList)
+  | where OperationName in~ (SensitiveOperationList)
   | where ActivityStatusValue =~ "Succeeded";
   SensitiveActivity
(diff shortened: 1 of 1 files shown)


Lodewyk-Git:
Has the same values as #2281

shainw:
@Lodewyk-Git - These were missed on update of the field name, the value will be changed shortly which will make this PR not needed.  AzureActivity uses a new connector and the change to OperationNameValue is needed.  @ashwin-patil is implementing the fixes and we will post back when fixed.

ashwin-patil:
PR #2867  has all the fixes where logic is not updated with the field `OperationNameValue` in queries, workbooks.

Lodewyk-Git:
Ok Thanks @ashwin-patil, I wasn't aware of that request.

PR #2993, opened 2021-09-10
Updating queries with common timestamp param to support future features.

shainw:
Fixes #

## Proposed Changes

  - Fixing standard time reference by removing UTC from end of labels
  - Including timestamp value in output to support future features
  -


aprakash13 (approved):
Looks Great. Thanks.

ashwin-patil (approved):

PR #3076, opened 2021-09-19
attribute folder, add queryBased + nrt data models

ShaniFelig:
Fixes #

## Proposed Changes

  - created subfolder for model validation attributes
  - implemented the query based data model
  - added nrt data model


aprakash13:
Working offline with Offir on this.

.script/tests/detectionTemplateSchemaValidation/Models/AnalyticsTemplateInternalModelBase.cs
```diff
@@ -15,6 +15,9 @@ public abstract class AnalyticsTemplateInternalModelBase
         [JsonProperty("id", Required = Required.Always)]
         public Guid Id { get; set; }
 
+        [JsonProperty("kind", Required = Required.Always)]
+        public AlertRuleKind Kind { get; set; }
```
oshvartz, line 19: this will break all existing templates - we want to to support backward something like 
  public AlertRuleKind Kind { get; set; } => Schedule (default)

oshvartz (changes requested):
:clock1:

.script/tests/detectionTemplateSchemaValidation/Models/AnalyticsTemplateInternalModelBase.cs
ShaniFelig: I have another PR that adds 'kind' to all existing templates which I plan to complete before this PR so nothing will break.

oshvartz: I think this is not good apporch since the date modified of all the templates will change to your PR = same of event groping - if not stated the default is used but it's up to you

ShaniFelig: It's true that the date modified will be changed but it has no significance to the customer because it is not exposed anymore (we have version property for this). btw, this has happened already a few times previously.

Regarding setting scheduled as default, I think this not accurate and we should avoid it. 
A) Scheduled is not the default, it is only one of the two kinds available now. So the template kind must be written explicitly by the author of the template to avoid mistakes
B) Removing the required validation of the kind property opens a window to many issues in serialization, tests and in the UX

@Amitbergman

PR #3096, opened 2021-09-22
Adding Azure Firewall Support to the NetworkConnectiontoOMIPorts Hunting Query

mohitkusecurity:
Fixes # Added Azure Firewall query to find attacks targeting OMI vulnerability (CVE-2021-38647) in DNAT traffic.

## Proposed Changes

  - Adding Azure Diagnostics query to detect connection attempts from the external IP addresses to the management ports(5985,5986,1270) related to Open Management Infrastructure(OMI) through DNAT rules in Azure Firewall.

### DIFF
--- a/Hunting Queries/MultipleDataSources/NetworkConnectiontoOMIPorts.yaml
+++ b/Hunting Queries/MultipleDataSources/NetworkConnectiontoOMIPorts.yaml
@@ -14,6 +14,9 @@ requiredDataConnectors:
   - connectorId: AzureMonitor(VMInsights)
     dataTypes:
       - VMConnection
+  - connectorId: AzureFirewall
+    dataTypes: 
+      - AzureDiagnostics
 tactics:
   - Reconnaissance
   - Initial Access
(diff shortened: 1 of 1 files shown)


Hunting Queries/MultipleDataSources/NetworkConnectiontoOMIPorts.yaml
```diff
@@ -51,5 +54,12 @@ query: |
   | parse VM_s with * '/' VM 
   | project TimeGenerated, SourceIp = SrcIP_s, DestinationIp = DestIP_s, DestinationPort = DestPort_d, Protocol = L7Protocol_s, NSGRule_s, VM, Type
   | extend Timestamp = TimeGenerated, HostCustomEntity = VM, IPCustomEntity = SourceIp
+  ),
+  AzureDiagnostics
+  | where Category == "AzureFirewallNetworkRule" and OperationName == "AzureFirewallNatRuleLog"
+  | parse msg_s with Protocol ' request from ' SourceIp ':' SourcePort ' to ' DestinationIp ':' DestinationPort " was " Action " to " InternalIP ':' InternalPort
+  | where DestinationPort in (Port)
+  | project TimeGenerated, SourceIP, DestinationIP, DestinationPort, Protocol, Action, Resource
```
aprakash13, line 62: We probably want to correct the case of the Source and Destination IP to be the same as what we have in the parse statement.
I guess we probably need SourceIp and DestinationIp in the project statement. Thanks.

Hunting Queries/MultipleDataSources/NetworkConnectiontoOMIPorts.yaml
```diff
@@ -51,5 +54,12 @@ query: |
...
+  | extend Timestamp = TimeGenerated, HostCustomEntity = Resource, IPCustomEntity = SourceIp
```
aprakash13, line 63: Seems like when we are mapping HostCustomEntity = Resource, we are mapping the host entity to the firewall that received the connection.  Wondering, if we really need the host entity mapping here or we need to remove it as it does not show the impacted host machine. Thanks.

aprakash13 (changes requested):

Hunting Queries/MultipleDataSources/NetworkConnectiontoOMIPorts.yaml
mohitkusecurity: @aprakash13 

Thanks for reviewing and catching the issue with case, it has been updated now.  I have also removed the HostCustomEntity = Resource mapping.  The idea behind this was to show Firewall resource name in the query result but understand how this can be confusing for customers when they are expecting to see the actual host machine in this mapping.

Let me know if there is anything else we need to update else, please approve and merge.

aprakash13 (approved):

PR #3099, opened 2021-09-23
OMI detection 

Ronmarsiano:
Adding a new detection that will validate that all agents connected to the workspace have close the OMI vulnerability

Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Detections/OMI_vulnerability/OMI_vulnerability_detection.yaml
+++ b/Detections/OMI_vulnerability/OMI_vulnerability_detection.yaml
@@ -1,0 +1,54 @@
+id: 3cc5ccd8-b416-4141-bb2d-4eba370e37a5
+name: OMI Vulnerability Exploitation
+description: |
+  Following the September 14th, 2021 release of three Elevation of Privilege
+  (EoP) vulnerabilities (CVE-2021-38645, CVE-2021-38649, CVE-2021-38648) and one
+  unauthenticated Remote Code Execution (RCE) vulnerability (CVE-2021-38647) in
+  the Open Management Infrastructure (OMI) Framework.
+
+  This detection validate that any OMS-agent that is reporting to the Azure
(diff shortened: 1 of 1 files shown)


Detections/OMI_vulnerability/OMI_vulnerability_detection.yaml
```diff
@@ -0,0 +1,54 @@
...
+  | where Category == "Direct Agent"
+  | summarize arg_max(TimeGenerated,*) by Computer
+  | parse strcat("Version:" , Version) with * "Version:" Major:long "."
+  Minor:long "." Patch:long "-" *
+  | parse OMIVulnerabilityPatchVersion with * "OMIVulnerabilityPatchVersion:"
+  OMIVersionMajor:long "." OMIVersionMinor:long "." OMIVersionPatch:long "-" *
+  | where Major <OMIVersionMajor or (Major==OMIVersionMajor and Minor
+  <OMIVersionMinor) or (Major==OMIVersionMajor and Minor==OMIVersionMinor and
+  Patch<OMIVersionPatch) 
+  | project Version, Major,Minor,Patch,
```
aprakash13, line 39: I tested the  query in one of our environments and seem to get high volume of results. I was wondering if we could aggregate the result so that we don't have multiple rows for the same machine just those with the latest status. We could probably project TimeGenerated and then do a summarize using an arg_max on it to get just the latest values. Please correct me if I am missing something here. Thanks for all your work on this.

aprakash13 (changes requested):

Detections/OMI_vulnerability/OMI_vulnerability_detection.yaml
Ronmarsiano: I am taking arg max so the load was decreased dramatically

aprakash13 (approved):

PR #3142, opened 2021-09-30
Dev/normalization/sysmon new version

YuvalNaor:
Fixes #

## Proposed Changes

  - fixing type mismatch in several parsers
  -
  -


oshezaf:
@YuvalNaor : 

- Process Id is documented as string.
- You changed the deployment folder to "Sessions". I think it is "Session"

oshezaf:
Also, i think - viewing on my phone so might be wrong - that your last commit was for a lot of files not belonging to us.

YuvalNaor:
https://docs.microsoft.com/en-us/azure/sentinel/process-events-normalization-schema

Procesaid is documented as it here

PR #3160, opened 2021-10-04
Created first version of parsers for Azure Firewall

TheCloudScout:
Fixes #parsing Azure Firewall logs

## Proposed Changes

Unfortunately logs from Azure Firewall are by default not parsed into separate columns. All details are logged as one single string in a column named "msg_s". These parsers will split up all of the details to make them better readable and available for further drilling down into the details.


AzureFirewall/AzureFirewallApplicationRule.kql
```diff
@@ -0,0 +1,38 @@
+let applicationRuleAllowsHttp = AzureDiagnostics
+| where ResourceType == "AZUREFIREWALLS"
+| where Category == "AzureFirewallApplicationRule"
+| parse msg_s with * " request from " ip_source ":" port_source " to " ip_destination ":" port_destination ". Url: " url ". Action: " action ". Policy: " policy ". Rule Collection Group: " rulecollectiongroup "Rule Collection: " rulecollection ". Rule: " rule
+| where action == "Allow"
```
oshezaf, line 5: A union parser with each union branch filtering after parsing is prohibitively slow. You need to filter on the original field (as in | where **msg_s has "Action: Allow"** two lines above). This comment is relevant for each branch.

AzureFirewall/AzureFirewallApplicationRule.kql
```diff
@@ -0,0 +1,38 @@
+let applicationRuleAllowsHttp = AzureDiagnostics
+| where ResourceType == "AZUREFIREWALLS"
+| where Category == "AzureFirewallApplicationRule"
+| parse msg_s with * " request from " ip_source ":" port_source " to " ip_destination ":" port_destination ". Url: " url ". Action: " action ". Policy: " policy ". Rule Collection Group: " rulecollectiongroup "Rule Collection: " rulecollection ". Rule: " rule
```
oshezaf, line 4: It would be highly beneficial if the field names correspond to [ASIM](https://docs.microsoft.com/en-us/azure/sentinel/normalization). The Azure Firewall would map to the Network and DNS schemas.

AzureFirewall/AzureFirewallApplicationRule.kql
```diff
@@ -0,0 +1,38 @@
...
+| where ResourceType == "AZUREFIREWALLS"
+| where Category == "AzureFirewallApplicationRule"
+| parse msg_s with * " request from " ip_source ":" port_source " to " ip_destination ":" port_destination ". Action: " action ". Policy: " policy ". Rule Collection Group: " rulecollectiongroup "Rule Collection: " rulecollection ". Rule: " rule ". Web Category: " webcategory
+| where action == "Allow"
+| project-away msg_s;
+let applicationRuleDeniesHttp = AzureDiagnostics
+| where ResourceType == "AZUREFIREWALLS"
+| where Category == "AzureFirewallApplicationRule"
+| parse msg_s with * " request from " ip_source ":" port_source " to " ip_destination ":" port_destination ". Url: " url ". Action: " action ". " rule ". " *
+| where action == "Deny" and port_destination == 80
```
oshezaf, line 23: Would only rules for port 80 log URL? Doesn't the firewall know how to handle other ports as HTTP (or even HTTPS)?

AzureFirewall/AzureFirewallApplicationRule.kql
```diff
@@ -0,0 +1,38 @@
...
+| where ResourceType == "AZUREFIREWALLS"
+| where Category == "AzureFirewallApplicationRule"
+| parse msg_s with * " request from " ip_source ":" port_source " to " ip_destination ":" port_destination ". Url: " url ". Action: " action ". Policy: " policy ". Rule Collection Group: " rulecollectiongroup "Rule Collection: " rulecollection ". Rule: " rule
+| where action == "Allow"
+| project-away msg_s;
+let applicationRuleAllowsHttps = AzureDiagnostics
+| where ResourceType == "AZUREFIREWALLS"
+| where Category == "AzureFirewallApplicationRule"
+| parse msg_s with * " request from " ip_source ":" port_source " to " ip_destination ":" port_destination ". Action: " action ". Policy: " policy ". Rule Collection Group: " rulecollectiongroup "Rule Collection: " rulecollection ". Rule: " rule
+| where action == "Allow" and port_destination !hasprefix "80" and rule !has "Web Category"
```
oshezaf, line 11: The union branch qualifying filters are not mutually exclusive. This will lead to duplication of events, even if the parse statement itself fails for one of the branches.

AzureFirewall/AzureFirewallApplicationRule.kql
```diff
@@ -0,0 +1,38 @@
+let applicationRuleAllowsHttp = AzureDiagnostics
+| where ResourceType == "AZUREFIREWALLS"
+| where Category == "AzureFirewallApplicationRule"
+| parse msg_s with * " request from " ip_source ":" port_source " to " ip_destination ":" port_destination ". Url: " url ". Action: " action ". Policy: " policy ". Rule Collection Group: " rulecollectiongroup "Rule Collection: " rulecollection ". Rule: " rule
+| where action == "Allow"
+| project-away msg_s;
+let applicationRuleAllowsHttps = AzureDiagnostics
+| where ResourceType == "AZUREFIREWALLS"
+| where Category == "AzureFirewallApplicationRule"
+| parse msg_s with * " request from " ip_source ":" port_source " to " ip_destination ":" port_destination ". Action: " action ". Policy: " policy ". Rule Collection Group: " rulecollectiongroup "Rule Collection: " rulecollection ". Rule: " rule
```
oshezaf, line 10: It is important to minimize union branches. When the two variants are similar enough, they can be covered by a single branch, say using two parse statements.

AzureFirewall/AzureFirewallApplicationRule.kql
```diff
@@ -0,0 +1,38 @@
...
+| where action == "Deny" and port_destination !hasprefix "80"
+| project-away msg_s;
+let applicationRuleDenies = AzureDiagnostics
+| where ResourceType == "AZUREFIREWALLS"
+| where Category == "AzureFirewallApplicationRule"
+| parse msg_s with * " request from " ip_source ":" port_source ". Action: " action ". Reason: " rule "." *
+| where isnotempty(ip_source)
+| project-away msg_s;
+applicationRuleAllowsHttp
+| union applicationRuleAllowsHttps, applicationRuleAllowsHttpsWebCategory, applicationRuleDeniesHttp, applicationRuleDeniesHttps, applicationRuleDenies
```
oshezaf, line 38: It would make sense to separate proxy (with URL) and network events (without URL) into different parsers as commonly queries need either but not both (and can union in the query if otherwise).

AzureFirewall/AzureFirewallApplicationRule.kql
```diff
@@ -0,0 +1,38 @@
...
+| parse msg_s with * " request from " ip_source ":" port_source " to " ip_destination ":" port_destination ". Url: " url ". Action: " action ". " rule ". " *
+| where action == "Deny" and port_destination == 80
+| project-away msg_s;
+let applicationRuleDeniesHttps = AzureDiagnostics
+| where ResourceType == "AZUREFIREWALLS"
+| where Category == "AzureFirewallApplicationRule"
+| parse msg_s with * " request from " ip_source ":" port_source " to " ip_destination ":" port_destination ". Action: " action ". " rule ". " *
+| where action == "Deny" and port_destination !hasprefix "80"
+| project-away msg_s;
+let applicationRuleDenies = AzureDiagnostics
```
oshezaf, line 31: This branch is a good example of why the added events fields that ASIM requires such as EventType and EventResult are very important to understand the parsers output. The branch provides minimal information but, potentially, not enough to understand what it is about.

AzureFirewall/AzureFirewallNetworkRule.kql
```diff
@@ -0,0 +1,32 @@
+let AzureFirewallNetworkRulesTcp = AzureDiagnostics
```
oshezaf, line 1: This is a good example of where union (which is not recommended if not needed) can be easily avoided.

oshezaf (changes requested):

oshezaf:
@YuvalNaor : FYI

shainw:
@mohitkusecurity - can you have a look at this parser?

shainw (changes requested):
These files are on the root of the AzureSentinel folder, it needs to be moved to AzureSentinel/Parsers/AzureFirewall.

TheCloudScout:
I was actually working on inprovements on this addition tonight and also noticed the incorrect path.

I’ll make sure this gets fixed in the next commit

Greets,

Koos
________________________________
Van: Shain ***@***.***>
Verzonden: Thursday, December 30, 2021 9:00:40 PM
Aan: Azure/Azure-Sentinel ***@***.***>
CC: TheCloudScout ***@***.***>; Author ***@***.***>
Onderwerp: Re: [Azure/Azure-Sentinel] Created first version of parsers for Azure Firewall (#3160)


@shainw requested changes on this pull request.

These files are on the root of the AzureSentinel folder, it needs to be moved to AzureSentinel/Parsers/AzureFirewall.

AzureFirewall/AzureFirewallApplicationRule.kql
TheCloudScout: I think I've managed to optimize everything by combining both `where` as `parse-where` statements before perming the `union`.

AzureFirewall/AzureFirewallApplicationRule.kql
TheCloudScout: I wasn't aware of the already existing ASIM parsers. I think Microsoft already did a way better job parsing the DNS entries for Azure Firewall in that parser. So I've removed my entry and pointed towards the ASIM one.

AzureFirewall/AzureFirewallApplicationRule.kql
TheCloudScout: You're right. I've taken a better look at how all the entries differ and create a better version with parse-where for each of the event types without looking to much at specific protocol of port details.

AzureFirewall/AzureFirewallApplicationRule.kql
TheCloudScout: I've now used `parse-where` and made sure I didn't miss anything by applying this on larger datasets (2 week of Azure Firewall logs) and comparing it with unparsed data.

AzureFirewall/AzureFirewallApplicationRule.kql
TheCloudScout: Like I've stated above, I think I managed to improve the parsers now by using a combination of `where` and `parse-where` before performing the `union`.

AzureFirewall/AzureFirewallApplicationRule.kql
TheCloudScout: Yes, you're right. That's what I did now with the new version.

AzureFirewall/AzureFirewallApplicationRule.kql
TheCloudScout: I think I used proper column names now matching with ASIM standards....

AzureFirewall/AzureFirewallNetworkRule.kql
TheCloudScout: Yes agreed. Like I've stated above, I think I managed to improve the parsers now by using a combination of `where` and `parse-where` before performing the `union`.

TheCloudScout:
Thanks everybody for looking at this! I think I've hugely improved the quality of the parsers now.
I've learned new things about ASIM and parsing. Thanks for pointing this out.

I hope this new commit now lives up to the proper standards.

oshezaf (changes requested):
Looks good. There is a challenge though with union/parse-where structure.
- Parse-where will not eliminate the unparsed events, just not set the relevant fields. Therefore you will get a lot of duplicates in the result set.
- Counting on parse failures will affect performance considerably.

To solve that you need to start each branch with a `where has` or `where has_all` clause to qaulify the events in the branch, for example `where msg_s has 'Failed to resolve'`.

Also, do you have, and can add to the sample data folder, events that can used to test the parser?

TheCloudScout:
> * Counting on parse failures will affect performance considerably.
> 
> To solve that you need to start each branch with a `where has` or `where has_all` clause to qaulify the events in the branch, for example `where msg_s has 'Failed to resolve'`.
> 
> Also, do you have, and can add to the sample data folder, events that can used to test the parser?

I've constructed two small sample datasets for both `ApplicationRule` and `NetworkRule` events which contains all variations of "msg_s" which led to the creation of each `parse-where` statement in these parsers.

> * Parse-where will not eliminate the unparsed events, just not set the relevant fields. Therefore you will get a lot of duplicates in the result set.

Maybe I understand your remark incorrectly but the whole idea of using `parse-where` is because it **only** outputs parsed rows. See also: _"The result is only the successfully parsed strings."_ on https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/parsewhereoperator.

If certain rows are applicable to multiple parsers, then yes; this indeed might cause issues and I did saw this happen during development of these parsers. But I did many comparisons on the amount of parsed and unparsed rows of larger datasets, that I don't end up with duplicates anymore.

oshezaf:
> > * Counting on parse failures will affect performance considerably.
> > 
> > To solve that you need to start each branch with a `where has` or `where has_all` clause to qaulify the events in the branch, for example `where msg_s has 'Failed to resolve'`.
> > Also, do you have, and can add to the sample data folder, events that can used to test the parser?
> 
> I've constructed two small sample datasets for both `ApplicationRule` and `NetworkRule` events which contains all variations of "msg_s" which led to the creation of each `parse-where` statement in these parsers.
> 
> > * Parse-where will not eliminate the unparsed events, just not set the relevant fields. Therefore you will get a lot of duplicates in the result set.
> 
> Maybe I understand your remark incorrectly but the whole idea of using `parse-where` is because it **only** outputs parsed rows. See also: _"The result is only the successfully parsed strings."_ on https://docs.microsoft.com/en-us/azure/data-explorer/kusto/query/parsewhereoperator.
> 
> If certain rows are applicable to multiple parsers, then yes; this indeed might cause issues and I did saw this happen during development of these parsers. But I did many comparisons on the amount of parsed and unparsed rows of larger datasets, that I don't end up with duplicates anymore.

You are right on inclusion of only the relevant rows. Apologies. My bad. The performance challenge is still there though I believe.

Also, can you share the test messages?

TheCloudScout:
Thanks Ofer for all the effort you're putting into this PR. ;-) I don't want to drag it any further but I'm learning as I go...

> The performance challenge is still there though I believe.

I finally get what you mean! I was afraid that a `where` with `has` (although more efficient than `contains`) was still a huge performance bottleneck, because you're doing some form of text search on all columns. My idea was that you might as well parse the rows while you're processing each and every single one of them.

But you are right. I did some testing with large datasets (up to 300 million events) and it is indeed much faster when I apply these statements within the parsing sections.

I updated the code so that it represents this optimization.

> Also, can you share the test messages?

I've committed two small sample datasets within this PR:

- Parsers/AzureFirewall/AzureFirewallNetworkRuleLogs_sample-dataset.kql
- Parsers/AzureFirewall/AzureFirewallApplicationRuleLogs_sample-dataset.kql

These contain all different variations on "msg_s" where every different `parse-where` statement was written for.
I'm also happy to provide a larger anonymized dataset if you'd like. But I don't think this would be the best place to store it.

Let me know if you have any other remarks!

Hopefully now all issues are resolved. :-)

oshezaf:
Looks fine and I can approve and merge. But questions first:
- Wouldn't you like to offer an ARM deploy rather than manual instructions?
- I am happy to support you in you want to convert also to an ASIM parser. It is not hard, but neither trivial.

~ Ofer

TheCloudScout:
> Wouldn't you like to offer an ARM deploy rather than manual instructions?

I've added a "Deploy to Azure" button which will deploy these parsers as a Query Pack. Since saving queries inside a workspace is now considered 'legacy' by Microsoft. ;-)

Thanks again for all the help!

oshezaf:
> > Wouldn't you like to offer an ARM deploy rather than manual instructions?
> 
> I've added a "Deploy to Azure" button which will deploy these parsers as a Query Pack. Since saving queries inside a workspace is now considered 'legacy' by Microsoft. ;-)
> 
> Thanks again for all the help!

Did you check that it works? As far as I know, Query packs do not work for functions. Happy to find out otherwise as it would be very useful.

TheCloudScout:
Yes I checked the deployment. You still need to save a query from a query pack as a function afterwards as stated in the readme. But the same counts for query saved inside the workspace. Query packs just replace the previous “savedQueries” within a workspace.

I figured that if it went through the effort of creating an ARM template and a “deploy to Azure” button, I might as well used the “new” approach for this.

But the readme still provides manual steps with pictures.

________________________________
Van: Ofer Shezaf ***@***.***>
Verzonden: Tuesday, January 11, 2022 5:27:21 PM
Aan: Azure/Azure-Sentinel ***@***.***>
CC: TheCloudScout ***@***.***>; Author ***@***.***>
Onderwerp: Re: [Azure/Azure-Sentinel] Created first version of parsers for Azure Firewall (#3160)


Wouldn't you like to offer an ARM deploy rather than manual instructions?

I've added a "Deploy to Azure" button which will deploy these parsers as a Query Pack. Since saving queries inside a workspace is now considered 'legacy' by Microsoft. ;-)

Thanks again for all the help!

Did you check that it works? As far as I know, Query packs do not work for functions. Happy to find out otherwise as it would be very useful.

oshezaf:
> Yes I checked the deployment. You still need to save a query from a query pack as a function afterwards as stated in the readme. But the same counts for query saved inside the workspace. Query packs just replace the previous “savedQueries” within a workspace. I figured that if it went through the effort of creating an ARM template and a “deploy to Azure” button, I might as well used the “new” approach for this. But the readme still provides manual steps with pictures.
> […](#)

I see. 

When deploying using a saved search ARM template, the function is automatically available as a function in the workspace eliminating the manual steps. Also, when using parameters with functions, the manual step becomes very cumbersome, and not possible at all for functions that accept table parameters. Long story short: since query packs do not support functions, Saved Searches ARM/API deployment is still the recommended method for functions and sometimes the only way.

TheCloudScout:
> Long story short: since query packs do not support functions, Saved Searches ARM/API deployment is still the recommended method for functions and sometimes the only way.

Thanks for clarifying Ofer! Sometimes we tend to embrace new features too soon I guess. ;-)
I've created a new ARM template based on 'savedQueries' and updated the instructions and "Deploy to Azure" button.

Does this mean this PR can finally be approved and you won't be bothered with it any longer?! ;-)

oshezaf (approved):

oshezaf:
@TheCloudScout : it was rather good as it was. I did want to help improve :-). Anyways. Merged. 

Now, I am going to use your code (with appropriate attribution obviously) for ASIM parsers as well, unless you strongly object, that is.

PR #3174, opened 2021-10-05
moving dependency on user input from query to WL

YaronFruchtmann:
Fixes #

## Proposed Changes

  - Changing InfobloxNIOS ASim and vim parsers
  - Created a template for ASimSourceType Watchlist 
  -


Parsers/ASimDns/ProductParsers/vimDnsInfobloxNIOS.yaml
```diff
@@ -49,22 +50,21 @@ ParserQuery: |
...
           and (srcipaddr=="*" or  has_ipv4( SyslogMessage ,srcipaddr))
           and (array_length(domain_has_any) ==0 or SyslogMessage has_any (domain_has_any))
           and (responsecodename=="*" or SyslogMessage has responsecodename)
           and (array_length(response_has_any_prefix)==0 or has_any_ipv4_prefix(SyslogMessage, response_has_any_prefix))
-          and (response_has_any_prefix=='*' or has_ipv4(SyslogMessage,response_has_any_prefix))
+          and (response_has_ipv4=='*' or has_ipv4(SyslogMessage,response_has_ipv4))
     // *****************************************************************
-    | where ProcessName == "named" and SyslogMessage has "client"
-      | where Computer in ("datasource") and Facility == "local6"
+      | where Computer in (Sources_by_SourceType('InfobloxNIOS'))
```
oshezaf, line 67: I think this should also precede the parameter checks.

Watchlists/ASim/readme.md
```diff
@@ -0,0 +1,13 @@
+# ASIM Source Types Watchlist
```
oshezaf, line 1: I will edit this once live.

oshezaf (changes requested):

Parsers/ASimDns/ProductParsers/ASimDns.yaml
```diff
@@ -15,12 +15,17 @@ References:
...
   union isfuzzy=true
     vimDnsEmpty
-    , ASimDnsCiscoUmbrella
-    , ASimDnsInfobloxNIOS
-    , ASimDnsMicrosoftOMS
-    , ASimDnsGcp
-    , ASimDnsCorelightZeek
-    , ASimDnsMicrosoftSysmon
\ No newline at end of file
+    , ASimDnsCiscoUmbrella     (disabled or (DisabledParser has 'CiscoUmbrella'))
```
oshezaf, line 26: The parser is DnsCiscoUmbrella and so on. CiscoUmbrella can have other parsers.

oshezaf (changes requested):

Parsers/ASimDns/ProductParsers/ASimDns.yaml
YaronFruchtmann: makes sense

Parsers/ASimDns/ProductParsers/ASimDns.yaml
```diff
@@ -15,12 +15,17 @@ References:
 Description: |
   ASIM Source Agnostic DNS Parser
 ParserName: ASimDns
+ParserParams:
+  - Name: disabled
+    Type: bool
+    Default: false
 ParserQuery: |
+  let DisabledParser=toscalar(_GetWatchlist('ASimDisabledParser') | project Parser=column_ifexists('Parser','') | summarize make_set(Parser));
```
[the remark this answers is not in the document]
YaronFruchtmann: Since the lists are very short (the number of disabled parsers, 10s of record worst case) I don't think it is signifcant. Also, running on several lines  "in" vs "has" is not significant. Changed it non the less.
However, the _GetWatchlist(*) function is slow, so I now added a materialize, to reduce the number of access calls.
Not sure what you mean with "lookup".

Parsers/ASimDns/ProductParsers/ASimDns.yaml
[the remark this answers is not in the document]
YaronFruchtmann: again, running on 5 records performance diff is not significant, but changed

PR #3185, opened 2021-10-07
Join Kind leftouter might be preferred

ep3p:
If we are not registering 4624 events, the results of replication could be missing, due to join of kind inner.

Please, could you review if this is possible? I might be mistaken, as with new Azure Monitor Agent you can select which events to register.

Fixes #

Results of replication could be missing, if 4624 events are not found.

## Proposed Changes

Use join kind of type leftouter, just in case right table might be missing.


aprakash13 (approved):
The changes look good. Approving this.

PR #3224, opened 2021-10-14
fix: Awake Security Workbook unset parameter value

adarshb20:
Fixes

- unsets hostname parameter value for Awake Security Workbook


adarshb20:
We didn't modified or has any KQL updates but it is still throwing the KqlValidation failure. Could you please us here?

adarshb20:
thanks @v-jayakal

PR #3231, opened 2021-10-17
Fix typo and case sensitivity to account watchlist insight

maayan-microsoft:
Add ```tolower``` to fix case sensitivity for the insight.


shainw:
@maayan-microsoft - Can you resolve the conflicts with the current Insights_Preview branch.  Also, the order of items in the YAML in Watchlist Insights (Account).yaml, please fix this.  Also @YaronFruchtmann for review.

maayan-microsoft:
I will fix the problem in another way

PR #3243, opened 2021-10-18
Updated the RiskIQ solution.

9b:
This pull requests overhauls a lot of the RiskIQ Solution to ease deployments for customers and take advantage of new functionality. More specifically, the following has been done within this push:

- Introduces a "base" playbook that pushes a single API connection for RiskIQ
- Directs all other playbooks to inherit from the base playbook variables
- Keeps "alert" trigger playbooks and introduces "incident" trigger playbooks for a number of categories
- Overhauled documentation to be more specific on deployment, post-deployment and configuration
- Rename of all playbook names to better describe their function

### DIFF
--- a/Solutions/RiskIQ/Playbooks/RiskIQ-Automated-Triage/alert-trigger/azuredeploy.json
+++ b/Solutions/RiskIQ/Playbooks/RiskIQ-Automated-Triage/alert-trigger/azuredeploy.json
@@ -452,12 +452,7 @@
                         "azuresentinel": {
                             "connectionId": "[resourceId('Microsoft.Web/connections', variables('AzureSentinelConnectionName'))]",
                             "connectionName": "[variables('AzureSentinelConnectionName')]",
-                            "id": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Web/locations/', resourceGroup().location, '/managedApis/azuresentinel')]",
-                            "connectionProperties": {
-                                "authentication": {
-                                    "type": "ManagedServiceIdentity"
-                                }
-                            }
+                            "id": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Web/locations/', resourceGroup().location, '/managedApis/azuresentinel')]"
                         },
                         "riskiqpassivetotal": {
                             "connectionId": "[resourceId('Microsoft.Web/connections', variables('RiskIQConnectionName'))]",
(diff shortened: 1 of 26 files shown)


9b:
Thank you @lior-tamir! I had tested these in the visual interface, but did not account for my template/replacement process to form the parameter arguments. We should be all set now!

lior-tamir:
> Apologies @anki-narravula, I assumed that was the only step I needed. I've configured my playbooks as outlined per the template.
> 
> With this managed identity change, I've noticed that additional steps[[1]](https://techcommunity.microsoft.com/t5/azure-sentinel/what-s-new-managed-identity-for-azure-sentinel-logic-apps/ba-p/2068204) are now required in order to have the playbooks run successfully. I need to add an Azure Sentinel Responder role to my workspace and then for each playbook, ensure that it's attached to that role. These steps are not obvious and a bit cumbersome. Is there a way to have all playbooks execute under this responder identity to avoid this configuration step? Additionally, is there anything I can do within the Solution package to automate this?
> 
> [1] https://techcommunity.microsoft.com/t5/azure-sentinel/what-s-new-managed-identity-for-azure-sentinel-logic-apps/ba-p/2068204

If playbooks are configured with managed identity, assining permissions to that managed identity on the Sentinel workspace requires to be an Owner on the workspace. I don't think the current Solutions deployment can support autom assignment / we want to add this prerequisite for the solution deployer? @anki-narravula 
My reccomendation is to make the playbook  (connector + workflow) configured with managed identity, so the only post deployment task would be grant permissions on the workspace.  If user wants to authenticate with user sign in / service principle after deployment, they can do so. (Anyway the user should authenticate after deployment, they cannot pick  an exiting connection via the current experience).

anki-narravula:
> > Apologies @anki-narravula, I assumed that was the only step I needed. I've configured my playbooks as outlined per the template.
> > With this managed identity change, I've noticed that additional steps[[1]](https://techcommunity.microsoft.com/t5/azure-sentinel/what-s-new-managed-identity-for-azure-sentinel-logic-apps/ba-p/2068204) are now required in order to have the playbooks run successfully. I need to add an Azure Sentinel Responder role to my workspace and then for each playbook, ensure that it's attached to that role. These steps are not obvious and a bit cumbersome. Is there a way to have all playbooks execute under this responder identity to avoid this configuration step? Additionally, is there anything I can do within the Solution package to automate this?
> > [1] https://techcommunity.microsoft.com/t5/azure-sentinel/what-s-new-managed-identity-for-azure-sentinel-logic-apps/ba-p/2068204
> 
> If playbooks are configured with managed identity, assining permissions to that managed identity on the Sentinel workspace requires to be an Owner on the workspace. I don't think the current Solutions deployment can support autom assignment / we want to add this prerequisite for the solution deployer? @anki-narravula My reccomendation is to make the playbook (connector + workflow) configured with managed identity, so the only post deployment task would be grant permissions on the workspace. If user wants to authenticate with user sign in / service principle after deployment, they can do so. (Anyway the user should authenticate after deployment, they cannot pick an exiting connection via the current experience).

Thank you @lior-tamir. We will analyze more about validating user's permissions while deploying the package itself.

PR #3272, opened 2021-10-22
Remove "target" column that is not used

ep3p:
Remove "target" column that is confusing when analyzing results and it is not used in the rest of the query.

OR, alternatively, **mv-expand TargetResources**, and then **extend** their properties with "TargetResources.property" instead of "TargetResources[0].property".

The change should depend on the length of TargetResources, in my tenant I have only seen it of size 1.

Fixes #

Unuseful column in results.

## Proposed Changes

  Remove "target column", OR, alternatively, apply the query to every element in array "TargetResources"


shainw:
@petebryan - can you have a look at this one.  You did the original and based on others for this, I think this is a typo of not using the 'target' field later in the query.

petebryan:
Looks like a good change. Target appears to be an un-necessary extend.

petebryan (approved):

PR #3292, opened 2021-10-25
Add column "DomainName" in IPEntity_DnsEvents.yaml

ep3p:
Some Threat Intelligence Indicators don't have a "Url" column but a "DomainName" column.

Fixes #

Missing info in results.

## Proposed Changes

Add the "DomainName" column of some Threat Intelligence Indicators.


aprakash13 (approved):
Looks Good.

PR #3323, opened 2021-10-27
[Proofpoint detection] - Fixes on TI match and lookback time.

### LINKED ISSUE #3137: [Proofpoint detection] TI use should take into account additional conditions
https://github.com/Azure/Azure-Sentinel/blob/3177f9141ca3f224ef79da84c91bc3c2fe20bd33/Detections/ProofpointPOD/ProofpointPODEmailSenderIPinTIList.yaml#L21

When querying the TI table, you need to filter out:
- Inactive IoCs
- IoCs that do not represent an IP address
- Check all IP addresses int the IoC. Specifically, malicious IPs relevant to ProofpointPOD are often in the EmailSourceIpAddress field.
- Take only the last copy of the IoC from the table

as a guideline, refer to the IoC list building query in https://github.com/Azure/Azure-Sentinel/blob/master/Detections/ASimDNS/imDns_IPEntity_DnsEvents.yaml

samikroy:
Fixes #3137 

## Proposed Changes

  - Filtered Inactive IoCs.
  - Checked possible IP address fields.
  - Taken only the last copy of the IoC from TI table
  - Updated the time frame for IOC look back from 30days to 14days as per Analytics Rule design.
  
@oshezaf , @Amitbergman, @ShaniFelig - Do we need to update the version number as well for these fixes !

### DIFF
--- a/Detections/ProofpointPOD/ProofpointPODMultipleProtectedEmailsToUnknownRecipient.yaml
+++ b/Detections/ProofpointPOD/ProofpointPODMultipleProtectedEmailsToUnknownRecipient.yaml
@@ -42,5 +42,5 @@ entityMappings:
     fieldMappings:
       - identifier: FullName
         columnName: AccountCustomEntity
-version: 1.0.1
+version: 1.1.0
 kind: Scheduled
--- a/Detections/ProofpointPOD/ProofpointPODMultipleProtectedEmailsToUnknownRecipient.yaml
+++ b/Detections/ProofpointPOD/ProofpointPODMultipleProtectedEmailsToUnknownRecipient.yaml
@@ -42,5 +42,5 @@ entityMappings:
     fieldMappings:
       - identifier: FullName
         columnName: AccountCustomEntity
-version: 1.0.0
+version: 1.0.1
 kind: Scheduled
--- a/Detections/ProofpointPOD/ProofpointPODMultipleProtectedEmailsToUnknownRecipient.yaml
+++ b/Detections/ProofpointPOD/ProofpointPODMultipleProtectedEmailsToUnknownRecipient.yaml
@@ -15,7 +15,7 @@ tactics:
   - Exfiltration
 query: |
   let lbtime = 30m;
-  let lbperiod = 30d;
+  let lbperiod = 14d;
   let knownrecipients = ProofpointPOD
   | where TimeGenerated > ago(lbperiod)
   | where EventType == 'message'
--- a/Detections/ProofpointPOD/ProofpointPODEmailSenderIPinTIList.yaml
+++ b/Detections/ProofpointPOD/ProofpointPODEmailSenderIPinTIList.yaml
@@ -15,31 +15,37 @@ tactics:
   - Exfiltration
   - InitialAccess
 query: |
+  let dt_lookBack = 1h;
   let ioc_lookBack = 14d;
-  let timeframe = 1h;
-  let TI_IP_List =
-  (ThreatIntelligenceIndicator
-  | where TimeGenerated > ago(ioc_lookBack) and ExpirationDateTime > now()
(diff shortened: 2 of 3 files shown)


samikroy:
@Amitbergman - Request you to suggest here.

samikroy:
ShaniFelig & @Amitbergman - please let me know for any modifications.

samikroy:
Request @Amitbergman  & @ShaniFelig to review and suggest further changes or approve.
Thank you.

Detections/ProofpointPOD/ProofpointPODMultipleProtectedEmailsToUnknownRecipient.yaml
```diff
@@ -42,5 +42,5 @@ entityMappings:
     fieldMappings:
       - identifier: FullName
         columnName: AccountCustomEntity
-version: 1.0.0
-kind: Scheduled
\ No newline at end of file
+version: 1.0.1
```
Amitbergman, line 45: version should be updated to 1.1.0. Same in other places.
On the logic of the change, please ask for @shainw  to review.
Thanks!

samikroy: Updated @Amitbergman .
Request @shainw to review for logic change.

shainw (changes requested):
We have an update for the TI queries beyond what you have added here.  We have only done it on a couple, but it will be done on all this week.  You can see what we have done here - https://github.com/Azure/Azure-Sentinel/blame/master/Detections/ThreatIntelligenceIndicator/IPEntity_AzureKeyVault.yaml

Look at the items changed on Nov 20th and you can use the same here.  If you are not sure, let me know and I will give you more specifics.

I also just posted this PR to get the changes to our detections in the TI folder - #3571

samikroy:
> @samikroy - We have an update for the TI queries beyond what you have added here. We have only done it on a couple, but it will be done on all this week. You can see what we have done here - https://github.com/Azure/Azure-Sentinel/blame/master/Detections/ThreatIntelligenceIndicator/IPEntity_AzureKeyVault.yaml
> 
> Look at the items changed on Nov 20th and you can use the same here. If you are not sure, let me know and I will give you more specifics.
> 
> I also just posted this PR to get the changes to our detections in the TI folder - #3571

@shainw - thank you for your inputs.
Have modified based on my understanding. 
Please guide for any more updates.

There are some build failures due to npm package failure.

Regarding #3477, could you please help to understand: 
1. Do we explicitly need to define the innerunique as that is KQL default.
2. Will this arg_max filter be applicable as innerunique will pick only one based on IP or email
or shall we apply the same in TI data !

shainw:
@samikroy
1. Yes, the specification of the join kind is a best practice and we will be updating all queries to make sure this is clearly defined.
2. Yes, in testing, we have seen the join can have multiple matches based on the right side of the join and times associated.  The argmax() is to make sure we have a single representation in the results as this is all that is needed as a starting point for investigation.
3. I will have a look at the build failures.

shainw:
@samikroy - build failures are not specific to this PR, there is some random NPM failure on random validators.  We have sent mail to the owner to have a look.

samikroy:
> @samikroy
> 1. Yes, the specification of the join kind is a best practice and we will be updating all queries to make sure this is clearly defined.
> 2. Yes, in testing, we have seen the join can have multiple matches based on the right side of the join and times associated.  The argmax() is to make sure we have a single representation in the results as this is all that is needed as a starting point for investigation.
> 3. I will have a look at the build failures.

Thanks a lot for clarifying @shainw 
Will be waiting for the merge.

shainw (approved):

shainw (approved):

PR #3329, opened 2021-10-28
Reduce Query Time Range in UserAccountCreatedDeleted_10m.yaml

ep3p:
If you join (2*timeframe) with (timeframe), when the timeframe is 1d, and the spanoftime is 10m, you will have queried 23h50m worth of data that will never be used, wasting computing resources.

Fixes #

Wasting computing resources not used.

## Proposed Changes

Reduce Query Time Range

### DIFF
--- a/Detections/SecurityEvent/UserAccountCreatedDeleted_10m.yaml
+++ b/Detections/SecurityEvent/UserAccountCreatedDeleted_10m.yaml
@@ -58,5 +58,5 @@ entityMappings:
     fieldMappings:
       - identifier: FullName
         columnName: HostCustomEntity
-version: 1.0.0
+version: 1.0.1
 kind: Scheduled
--- a/Detections/SecurityEvent/UserAccountCreatedDeleted_10m.yaml
+++ b/Detections/SecurityEvent/UserAccountCreatedDeleted_10m.yaml
@@ -12,7 +12,7 @@ requiredDataConnectors:
     dataTypes:
       - SecurityEvent
 queryFrequency: 1d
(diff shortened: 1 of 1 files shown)


Detections/SecurityEvent/UserAccountCreatedDeleted_10m.yaml
```diff
@@ -12,7 +12,7 @@ requiredDataConnectors:
     dataTypes:
       - SecurityEvent
 queryFrequency: 1d
-queryPeriod: 2d
+queryPeriod: 25h
```
shainw, line 15: We can set this, doesn't really make any difference in the query because the timeframes are hard coded.  This is used if there is no timeframe specified which is generally preferred when possible.

Detections/SecurityEvent/UserAccountCreatedDeleted_10m.yaml
```diff
@@ -25,17 +25,17 @@ query: |
   let timeframe = 1d;
   let spanoftime = 10m;
   let threshold = 0;
-  SecurityEvent 
-  | where TimeGenerated > ago(2*timeframe) 
+  SecurityEvent
+  | where TimeGenerated > ago(timeframe+spanoftime)
```
shainw, line 29: Agreed, the old value doesn't actually help because the comparison time frame is only 1d. :)  Adding in the extra 10m just provides the ability to check for the absolute full match and based on timings could help pick up a miss on a daily run.

shainw (approved):

PR #3330, opened 2021-10-28
Reduce Query Time Range in UserAccountEnabledDisabled_10m.yaml

ep3p:
If you join (2*timeframe) with (timeframe), when the timeframe is 1d, and the spanoftime is 10m, you will have queried 23h50m worth of data that will never be used, wasting computing resources.

Fixes #

Wasting computing resources not used.

## Proposed Changes

Reduce Query Time Range.


shainw (approved):
Agreed, same update as #3329

PR #3345, opened 2021-10-29
Update solution rules yamls and adding 1 playbook

liranylevy:
Fixes #

## Proposed Changes

  -
  -
  -


aprakash13 (changes requested):
Not sure why we removed relevantTechniques from the query templates. It's required for Detections and Hunting Queries and is generally suggested to have it in there.
https://github.com/Azure/Azure-Sentinel/wiki/Query-Style-Guide#relevanttechniques

Other than that the queries look good to be approved.

liranylevy:
@aprakash13 pushed fixed including techniques back

aprakash13 (approved):
Discussed with author to look at the queryFrequency and queryPeriod parameters as we normally have 1h for high severity detections only. Approving for now.

PR #3351, opened 2021-10-30
AADSecops Detection and hunting Queries

ashwin-patil:
Fixes #

## Proposed Changes

Detections and Hunting Queries for AAD Secops Guide
  -
  -


petebryan:
Looks like you have 2 templates with ID 34c5aff9-a8c2-4601-9654-c7e46342d03b (one detection and one hunting). It shouldn't be an issue technically but it might be best to change one to make identification easier.

petebryan:
Also looks like you'll need to exclude 34c5aff9-a8c2-4601-9654-c7e46342d03b (detection) from KQL validation as its not seeing the correct schema. I have confirmed the query is correct in Sentinel data.

ashwin-patil:
> Looks like you have 2 templates with ID 34c5aff9-a8c2-4601-9654-c7e46342d03b (one detection and one hunting). It shouldn't be an issue technically but it might be best to change one to make identification easier.

Fixed duplicate template Ids.

>Also looks like you'll need to exclude 34c5aff9-a8c2-4601-9654-c7e46342d03b (detection) from KQL validation as its not seeing the correct schema. I have confirmed the query is correct in Sentinel data.

Removed column name for now as its not necessary in most cases

petebryan (approved):

PR #3360, opened 2021-11-02
[Data connectors] Aws-S3 script update

sagamzu:
[Data connectors] Aws-S3 script update according to bugbash inputs:

- add "destination table" to the final output
- update logs default path to put the logs in a dedicated folder ( Logs)
- Update "Write-Log" function to have a Color parameter for info output. (we use it to highlight a new text)
- Remove S3 bucket input. this value is available in aws config, so we don't need to ask the customer for this input
- update vpcTrafficType  value to "Upper case". the validation is insensitive case, but AWS excepted this value to be in upper case 
- Remove `aws config` out of sentinel script according to security team request. 
     -  Update the readme file 
     -  Add aws configuration validation at the benign of the script (this validation was missing also before this change)

### DIFF
--- a/DataConnectors/AWS-S3/Utils/AwsResourceCreator.ps1
+++ b/DataConnectors/AWS-S3/Utils/AwsResourceCreator.ps1
@@ -23,7 +23,7 @@ function New-ArnRole
             Write-Output "`n`n"
             Write-Log "You must specify the the Azure Sentinel Workspace ID. This is found in the Azure Sentinel portal." -LogFileName $LogFileName -Severity Information -LinePadding 1
             
-            $workspaceId = Read-ValidatedHost -Prompt "Please enter your Azure Sentinel Workspace ID (External Id)"
+            $workspaceId = Read-ValidatedHost -Prompt "Please enter your Azure Sentinel External ID (Workspace ID)"
             Write-Log "Using Azure Sentinel Workspace ID: $workspaceId" -LogFileName $LogFileName -Severity Information -Indent 2
 
             $rolePolicy = Get-RoleArnPolicy -WorkspaceId $workspaceId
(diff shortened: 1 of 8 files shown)


DataConnectors/AWS-S3/Utils/AwsResourceCreator.ps1
```diff
@@ -23,7 +23,7 @@ function New-ArnRole
             Write-Output "`n`n"
             Write-Log "You must specify the the Azure Sentinel Workspace ID. This is found in the Azure Sentinel portal." -LogFileName $LogFileName -Severity Information -LinePadding 1
             
-            $workspaceId = Read-ValidatedHost -Prompt "Please enter your Azure Sentinel Workspace ID (External Id)"
+            $workspaceId = Read-ValidatedHost -Prompt "Please enter your Azure Sentinel External ID (Workspace ID)"
```
joelst, line 26: Should this be: "Please enter the External Id (Azure Sentinel Workspace ID)"

I think it is clear as it is, but just want to make sure.

DataConnectors/AWS-S3/Utils/AwsResourceCreator.ps1
```diff
@@ -54,37 +54,40 @@ function New-S3Bucket
         
         # Get s3 bucket name from user and clean up based on naming rules see https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-s3-bucket-naming-requirements.html
         
-        $script:bucketName = (Read-ValidatedHost -Prompt "Please enter S3 bucket name" -MaxLength 64 -MinLength 3)
+        $script:bucketName = (Read-ValidatedHost -Prompt "Please enter S3 bucket name (between 3 and 63 characters long)" -MaxLength 64 -MinLength 3)
 
         Write-Log -Message "Using S3 Bucket name: $bucketname" -LogFileName $LogFileName -Indent 2
-            
+        
+        $regionConfiguration = aws configure get region
```
joelst, line 61: Very nice! I thought this would be a good idea, but had not figured out how to do it.

joelst (approved):
The updates look good.

DataConnectors/AWS-S3/Utils/AwsResourceCreator.ps1
sagamzu: I think it is clear. 
this text is the same as we have in the UI and there is the explanation comment above, and

DataConnectors/AWS-S3/Utils/AwsResourceCreator.ps1
sagamzu: Thanks :)

PR #3384, opened 2021-11-05
Fix Cisco SE loading of data from API

### LINKED ISSUE #3362: Cisco Secure Endpoint API Call Fails
**Describe the bug**
The Cisco Secure Endpoint Function App fails with the error in Screenshots below.  I manually called the endpoint to make sure it wasn't something with my API key and I can successfully pull the data from the URL the script is using with no problem.  I also ensured that the "data" node exists and should be subscriptable.

**To Reproduce**
1. Install Cisco Secure Endpoint connector
2. Enter API Client ID & API Key
3. Deploy Function
4. Error produced

**Expected behavior**
Function should complete successfully

**Screenshots**
image

scottymcraig:
Fixes #3362

## Proposed Changes

  - Add the json library
  - Parse the data from the stream
  - Fixes where the data is actually stored in the API

### DIFF
--- a/Solutions/CiscoSecureEndpoint/Data Connectors/AzureFunctionCiscoSecureEndpoint/main.py
+++ b/Solutions/CiscoSecureEndpoint/Data Connectors/AzureFunctionCiscoSecureEndpoint/main.py
@@ -3,6 +3,7 @@ import datetime
 import logging
 import re
 import requests
+import json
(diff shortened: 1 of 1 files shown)


NikTripathi:
@scottymcraig Please add the updated zip package in this PR. Thank you.

v-rucdu (approved):

PR #3386, opened 2021-11-05
Add CiscoSE High Alert Rule

scottymcraig:
## Proposed Changes

  - Add rule to Cisco Secure Endpoint alerting on High Alerts

### DIFF
--- a/Solutions/CiscoSecureEndpoint/Analytic Rules/CiscoEndpointHighAlert.yaml
+++ b/Solutions/CiscoSecureEndpoint/Analytic Rules/CiscoEndpointHighAlert.yaml
@@ -2,7 +2,7 @@ id: 4683ebce-07ad-4089-89e3-39d8fe83c011
 name: Cisco SE High Events Last Hour
 description: |
   'Find events from Cisco Secure Endpoint that are of High severity in the last hour.'
-severity: Medium
+severity: High
 requiredDataConnectors:
   - connectorId: CiscoSecureEndpoint
     dataTypes:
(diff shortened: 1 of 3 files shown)


Solutions/CiscoSecureEndpoint/Analytic Rules/CiscoEndpointHighAlert.yaml
```diff
@@ -0,0 +1,38 @@
...
+  - connectorId: CiscoSecureEndpoint
+    dataTypes:
+      - CiscoSecureEndpoint_CL
+queryFrequency: 1h
+queryPeriod: 1h
+triggerOperator: gt
+triggerThreshold: 0
+tactics:
+  - Execution
+  - InitialAccess
```
aprakash13, line 16: Do we want to add the relevantTechniques: section after this.

Solutions/CiscoSecureEndpoint/Analytic Rules/CiscoEndpointHighAlert.yaml
```diff
@@ -0,0 +1,38 @@
...
+query: |
+  let endtime = 1h;
```
aprakash13, line 18: Since the Query Frequency is set to 1 hr . Wondering if we need the endtime value.

aprakash13 (changes requested):

Solutions/CiscoSecureEndpoint/Analytic Rules/CiscoEndpointHighAlert.yaml
scottymcraig: We can if you want, but it's going to be a long list - since High alerts can be everything from executed malware to CnC server access, it could be hard to identify specific ones.

Solutions/CiscoSecureEndpoint/Analytic Rules/CiscoEndpointHighAlert.yaml
scottymcraig: We don't if the user doesn't change the timeframe. However, if they do without thinking, it could be a really long-running query so I kept it in there to prevent this.

v-jayakal:
@aprakash13 - Can we please approve? Then we will merge this PR. Thanks.

NikTripathi:
@aprakash13 can you please approve? Thank you.

PR #3388, opened 2021-11-05
Add ShadowByte Aria Solution

dwolosowicz:
Fixes #

## Proposed Changes

  -
  -
  -


anki-narravula (changes requested):
Hi @dwolosowicz  Thanks for creating ShadowByte Aria solution. Overall it looks fine, but I would recommend below changes for smoother deployment for customers
1. For sentinel connector, we advise to use managed identity in the template itself. If required customers can modify as per need - Please find [template ](https://github.com/Azure/Azure-Sentinel/blob/master/Playbooks/.template/incident-trigger/azuredeploy.json)here if required
image
image

2. In the base (connector) playbook default value for "Connector Name" mentioned as ShadowByteAria and in the other playbooks you have mentioned as "ShadowByteAriaConnector", so playbook 2 and 3 fails if customers are using default parameters as it is
3. For playbook 3, i would recommend to accept Log analytics workspace id and key during deployment as parameters for authentication. And of course we can change at anytime with "Edit API Connectors"

dwolosowicz:
Hello @anki-narravula. I've implemented the requested changes.

anki-narravula (approved):
Hi @dwolosowicz  - Looks good.  We can merge this to master, @NikTripathi pls merge

PR #3421, opened 2021-11-10
Corrects parameter misspelling in ASIM parsers.

Korving-F:
Functions as part of ASimAuthentication Parsers are using one parameter name in in the signature (`disable`) and another in the body (`disabled`). This PR corrects it to `disabled` which is how it is defined in the `ParserParams`-section as well as what other such functions are using. 

This does not yet correct the [ARM templates](https://github.com/Azure/Azure-Sentinel/tree/master/Parsers/ASimAuthentication/ARM) which have the same issue embedded.


oshezaf:
Thanks @Korving-F , we will look into this.

YaronFruchtmann (approved):
Thanks!

PR #3438, opened 2021-11-14
DNS parser updated for schema 1.3

oshezaf:


Parsers/ASimDns/ProductParsers/VimDnsGcp.yaml
```diff
@@ -56,7 +56,7 @@ ParserQuery: |
     | where resource_type_s == "dns_query"
     //  Pre-parsing filtering:
     | where
-      (eventtype=='lookup') // -- for now we support only lookup events
+    (eventtype in ('lookup', 'Query')) // -- for now we support only lookup events
```
YaronFruchtmann, line 59: Applies to all vim files: I noticed you did not change the default value.  Do you find it risky?

Parsers/ASimDns/ProductParsers/vimDnsMicrosoftOMS.yaml
```diff
@@ -68,65 +95,69 @@ ParserQuery: |
...
-    , response_has_any_prefix:dynamic=dynamic([]) , eventtype:string='lookup'
-    , disabled:bool=false
+    starttime:datetime=datetime(null), 
+    endtime:datetime=datetime(null),
+    srcipaddr:string='*',
+    domain_has_any:dynamic=dynamic([]),
+    responsecodename:string='*', 
+    response_has_ipv4:string='*',
+    response_has_any_prefix:dynamic=dynamic([]),
+    eventtype:string='*',
```
YaronFruchtmann, line 105: did you mean to change that?

YaronFruchtmann (changes requested):
I think you accidently changed one the default parameters values

Parsers/ASimDns/ProductParsers/VimDnsGcp.yaml
oshezaf: Thanks!

I update for consistent handling:
- Default is always Query
- lookup is accepted as a parameter and handled as Query.

PR #3441, opened 2021-11-15
addressing some feedback around instructions

javiersoriano:
Addressing some feedback on instructions in the lab.

Moving from Azure Sentinel Solutions to Content Hub.


shainw (approved):
Looks good, even though validations are not working right now, image replacement names are consistent, the rest is text so good to go.

Issue #3401, opened 2021-11-08
Wrong entity mapping in Nobelium FoggyWeb rule


javiersoriano:
**Describe the bug**
Detection Nobelium FoggyWeb (https://github.com/Azure/Azure-Sentinel/blob/master/Detections/MultipleDataSources/Nobelium_FoggyWeb.yaml) has incorrect entity mapping for Process entity.

**To Reproduce**
Try to enable Analytic rule Nobelium FoggyWeb. "Set rule logic" tab will show an error about not being able to find the column

It seems like the only available columns would be ProcessCustomEntity_dynamic or ProcessCustomEntity_string, but the rule definition specifies ProcessCustomEntity.

**Screenshots**
![image](https://user-images.githubusercontent.com/13329585/140770586-a2ce6625-1d31-4a39-9947-8c1f9b79b180.png)

PR #3444, opened 2021-11-15
Update Nobelium_FoggyWeb.yaml

shainw:
Fixing ProcessCustomEntity mapping to make sure type is string so we don't get 2 types.

Fixes #
Line 58 - output split (dynamic) to string - ProcessCustomEntity = tostring(split(Image, '\\', -1)[-1])
Line 96 - output split (dynamic) to string - ProcessCustomEntity = tostring(split(Image, '\\', -1)[-1])

Fix for issue #3401


aprakash13 (approved):
ProcessCustomEntity mapping fixes look good.

aprakash13:
Although this PR has validation issues currently, it is because of overall Github validation problem and not because of the template query itself.  This has been verified and looks ok. In the interest of getting the UX fixes in; merging this as of now after discussing with shainw.

Detections/MultipleDataSources/Nobelium_FoggyWeb.yaml
```diff
@@ -188,5 +188,5 @@ entityMappings:
     fieldMappings:
       - identifier: ProcessId
         columnName: ProcessCustomEntity
-version: 1.0.0
-kind: Scheduled
\ No newline at end of file
+version: 1.0.1
```
Amitbergman, line 191: I would do a major version bump as it is a major bug fix.
@shainw

PR #3450, opened 2021-11-16
Asimauth parser correction

Korving-F:
This PR fixes the datatype for the LogonTypes lookup from string to int. This fixes the lookup and brings the `ASimAuthenticationMicrosoftWindowsEvent` parser in line with `vimAuthenticationMicrosoftWindowsEvent` as defined [here](https://github.com/Azure/Azure-Sentinel/blob/master/Parsers/ASimAuthentication/ProductParsers/vimAuthenticationMicrosoftWindowsEvent.yaml).


YaronFruchtmann (approved):
Thanks.

YaronFruchtmann (approved):

PR #3458, opened 2021-11-17
Fix count syntax

rpressburger:
Fixes #
Fix count syntax so avoid syntax errors.

## Proposed Changes
count(<column name>) has the same meaning as count()
It’s an old syntax that was left for backward compatibility.


liemilyg (approved):

liemilyg:
merge

liemilyg (approved):

PR #3469, opened 2021-11-17
Moved hunting queries to workspace deployment saved searches

samikroy:
Fixes # Removed actual tables from hunting queries and analytic rule.

## Proposed Changes

  - Moved Solorigate Inventory check hunting query to workspace.json
  - Removed actual tables from Moved Solorigate Inventory
  - Removed actual tables from Solorigate Network Beacon Analytic rule


samikroy:
@javiersoriano & @v-maudan - please help in merging this as this validation is due to the files outside this PR.

shainw:
@samikroy - I think the validation failure is properly failing, I am just not sure the validation should be configured the way it is.  You are checking in JSON files to the Solutions folder.  The validation indicates there should only be YAML files which is wrong and not only that the validation does not indicate which files should be YAML files.

Your checked in files:
Solutions/Training/Azure-Sentinel-Training-Lab/Artifacts/LinkedTemplates/alertRules.json
Solutions/Training/Azure-Sentinel-Training-Lab/Artifacts/LinkedTemplates/workspace.json
Solutions/Training/Azure-Sentinel-Training-Lab/Package/mainTemplate.json

Validation failure snip

[xUnit.net 00:00:53.44]     Kqlvalidations.Tests.DetectionTemplateSchemaValidationTests.Validate_DetectionTemplates_AllFilesAreYamls [FAIL]
  X Kqlvalidations.Tests.DetectionTemplateSchemaValidationTests.Validate_DetectionTemplates_AllFilesAreYamls [18ms]
  Error Message:
   All the files in detections and solution (Analytics rules) folder are supposed to end with .yaml
Expected: True
Actual:   False

@v-maudan  and @Amitbergman - I think you both worked on these validations and might be able to diagnose the bug in - https://github.com/Azure/Azure-Sentinel/blob/master/.script/tests/detectionTemplateSchemaValidation/DetectionTemplateSchemaValidationTests.cs

samikroy:
Thank you @shainw .
Request request your help @v-maudan & @Amitbergman  here to proceed.
As this PR contains contains changes which will distinguish an actual incident  being raised with current training solution solution .

shainw:
Scratch that, just re-ran the checks and it passed.  I think @petebryan fixed the issue in a recent PR - https://github.com/Azure/Azure-Sentinel/pull/3471/files.  Looks like good to get approved now if @javiersoriano and @v-maudan  agree.

samikroy:
Thank you @shainw.
Request @javiersoriano & @v-maudan your help here.

javiersoriano:
Hi @samikroy , 

We tried to leave the rules as similar as possible to the original ones, that's why we left the rest of the tables in the query even though they are not used. What is the benefit for the user if we remove the other tables?

I agree on moving the hunting query to the workspace instead of having it in mainTemplate.json.

The rest of the changes are just cosmetic.

samikroy:
Agree.
The reason have tried to remove the actual tables is in case this solution is enabled in existing prod Sentinel then this rules might mix up a prod & test data during incident creation.
But, in case if you suggest otherwise, will revert the changes.

javiersoriano:
Got it. If that's the reason, I would change the name of the rule, so it clearly indicates that this is a test rule and that is seen in the incident as well. 

That would involve changing screenshots and text in the guide though

samikroy:
> Got it. If that's the reason, I would change the name of the rule, so it clearly indicates that this is a test rule and that is seen in the incident as well.
> 
> That would involve changing screenshots and text in the guide though

 In addition, we might need to update this in the UI template too, but this this file is not a latest version (please point me to the correct file in that case)
 
 https://github.com/Azure/Azure-Sentinel/blob/master/Solutions/Training/Azure-Sentinel-Training-Lab/Package/createUiDefinition.json

javiersoriano:
@samikroy the intention of the training lab is to be deployed in a clean workspace, and for sure not in a production workspace, so we are hesitant to change the rule name.

can you just leave the move of the savedSearch to workspace.json and remove the other changes so we can approve?

Thanks

samikroy:
> @samikroy the intention of the training lab is to be deployed in a clean workspace, and for sure not in a production workspace, so we are hesitant to change the rule name.
> 
> can you just leave the move of the savedSearch to workspace.json and remove the other changes so we can approve?
> 
> Thanks

@javiersoriano - Done.
Thank you.

javiersoriano:
Thanks @samikroy

All good from my side to approve, but I don't have permissions. @shainw or @v-maudan could you please approve?

samikroy:
Request @shainw & @v-maudan to approve and merge this.
Let me know for any queries.
Thank you.

sreedharande (approved):

PR #3471, opened 2021-11-18
Blog Support Queries

aprakash13:
Fixes #

## Proposed Changes

  -
  -
  -


petebryan (approved):

oshezaf:
@aprakash13 , @petebryan : a few rather important comments in this PR:

**ActingProcessSHA256**

- The need to add ActingProcessSHA256 to the imFileEvent custom validation raises an issue we have to deal with on the validation side. We do promote normalizing as many fields even if we did not add them to the schema s they are not very common in sources, but don't do good in supporting validation for them.
- That said, and expectedly for a non-schema field, using ActingProcessSHA256 in analytics presents a challenge: it is not part of the schema since file event sources do not commonly provide hash values for the process executable:
   - If there is a real need to get ActingProcess hash information, it is really recommended to correlate with a process event, probably based on the Acting Process Name. Maybe in this specific case, you want to correlate only with the process event (having the correct Hash for the process probably is much more important than the file names).

**GuardDuty**
- In general, built-in tables do not need to have custom validation jsons. I have raised the issue with Amit and Alaa (who is in charge of the AWS connectors). Probably an issue with updating the validation system with the latest built-in schemas.

PR #3477, opened 2021-11-19
Save a summarize and use correct Time column in most TI Map Detections

ep3p:
Fixes #

TimeGenerated comes from the table you want to check, NOT from ThreatIntelligenceIndicator (its TimeGenerated was substituted by LatestIndicatorTime).

It does not make sense to compare "Source_TimeGenerated >= TimeGenerated", when **they are the same value**. "Source_TimeGenerated" should be compared to "LatestIndicatorTime".

"**extend** Source_TimeGenerated = TimeGenerated" should not have been used, but "**project-rename** Source_TimeGenerated = TimeGenerated".

Therefore a second summarize computing LatestIndicatorTime is not needed, **multiple events related to the same Indicator are lost** in the results.

More changes might be needed, depending on the results.

Please, could you check this?

## Proposed Changes

  - Removing second summarize.
  - Using correct time column in the comparison


shainw:
@ep3p - 

Appreciate the thoughts here and you have helped find a way to improve the query but not exactly what you are doing as in my testing your change may miss matches.  The portion of the where clause you point as being not useful (where AzureActivity_TimeGenerated >= TimeGenerated) is a valid point. To demonstrate what I see as the change, I will use this query as an example - https://github.com/Azure/Azure-Sentinel/blob/master/Detections/ThreatIntelligenceIndicator/EmailEntity_AzureActivity.yaml

- I overwrote the emailaddress field in the TI results so I know that I will get a match against AzureActivity by doing this in the query:  | extend EmailSenderAddress = '<account we know is in Azure Activity log>'
- When using your changes and specifically this change, I get no results: | where AzureActivity_TimeGenerated >= LatestIndicatorTime
  - This is because LatestIndicatorTime reflects when the entry was last updated, not the timeframe for when it is valid.  By doing the where clause you as you have done, the query miss any matches for when the indicator has a value of Active=True.

The changes after the JOIN that I believe we should do are the following:

- Remove this where clause: AzureActivity_TimeGenerated >= TimeGenerated
- Keep this portion of the where clause as it keeps us from matching on Expired on non-Active indicators: AzureActivity_TimeGenerated < ExpirationDateTime
- We need to keep the summarize with the argmax, otherwise we could get hundreds of matches for each time the indicator was last updated, but we do change it to summarize on the latest AzureActivity time match as that is more useful to the investigtion and we also should include the join condition field, in this case Caller so that we don't lose all the accounts involved: | summarize AzureActivity_TimeGenerated = arg_max(AzureActivity_TimeGenerated, *) by IndicatorId, Caller

When doing the changes above, I get 1 match.

Proposed query:
```yaml  let dt_lookBack = 45m;
  let ioc_lookBack = 10m;
  let emailregex = @'^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$';
  ThreatIntelligenceIndicator
  | where TimeGenerated >= ago(ioc_lookBack) and ExpirationDateTime > now()
  | extend EmailSenderAddress = '<input your own account>'
  | summarize LatestIndicatorTime = arg_max(TimeGenerated, *) by IndicatorId
  | where Active == true
  //Filtering the table for Email related IOCs
  | where isnotempty(EmailSenderAddress)
  //using innerunique to keep perf fast and result set low, we only need one match to indicate potential malicious activity that needs to be investigated
  | join kind=innerunique (
      AzureActivity | where TimeGenerated >= ago(dt_lookBack) and isnotempty(Caller)
      | extend Caller = tolower(Caller)
      | where Caller matches regex emailregex
      | extend AzureActivity_TimeGenerated = TimeGenerated
  )
  on $left.EmailSenderAddress == $right.Caller
  | where AzureActivity_TimeGenerated < ExpirationDateTime
  | summarize AzureActivity_TimeGenerated = arg_max(AzureActivity_TimeGenerated, *) by IndicatorId, Caller
```
Also,  we should declare the join clearly as innerunique and comment as to why.  The comment would be something like this - 
//using innerunique to keep perf fast and result set low, we only need one match to indicate potential malicious activity that needs to be investigated

I also tested this same change on: https://github.com/Azure/Azure-Sentinel/blob/master/Detections/ThreatIntelligenceIndicator/IPEntity_AzureActivity.yaml.

Let me know your thoughts.

ep3p:
@shainw I agree mostly with you.

> * We need to keep the summarize with the argmax, **otherwise we could get hundreds of matches for each time the indicator was last updated**, but we do change it to summarize on the latest AzureActivity time match as that is more useful to the investigtion and we also should include the join condition field, in this case Caller so that we don't lose all the accounts involved: | summarize AzureActivity_TimeGenerated = arg_max(AzureActivity_TimeGenerated, *) by IndicatorId, Caller

I think the first summarize will already get rid of multiple updates of Indicators.

> ```yaml
>   ...
>   | summarize LatestIndicatorTime = arg_max(TimeGenerated, *) by IndicatorId
>   ...
>   on $left.EmailSenderAddress == $right.Caller
>   | where AzureActivity_TimeGenerated < ExpirationDateTime
>   | summarize AzureActivity_TimeGenerated = arg_max(AzureActivity_TimeGenerated, *) by IndicatorId, Caller
> ```
> 
> Also, we should declare the join clearly as innerunique and comment as to why. The comment would be something like this - //using innerunique to keep perf fast and result set low, we only need one match to indicate potential malicious activity that needs to be investigated

If the join key is Caller, I don't think Caller is needed in the 2nd summarize:

```yaml
  | summarize AzureActivity_TimeGenerated = arg_max(AzureActivity_TimeGenerated, *) by IndicatorId
```
should have the same results (because two Indicators with the same ID but different EmailSenderAddress should not exist).

I think generally the join key will not be needed to summarize by.

I think in some cases a second column to summarize by (a column that has not been used as join key) could be useful.

For example, if a device communicates with the IP address of an Indicator, I think we would like to know all the different devices the query finds. To know which rules could profit from something like this, each rule should be checked... it may not be worth it. A client can always edit the rule if needed.

I understand if you prefer to have only **the last event** for each Indicator **in all cases**.

Let me know what you prefer.

ep3p:
Some Detection Rules do not have a 2nd summarize from the beginning, in case you want them all to have it.

ep3p:
Hi @shainw , I have checked the next code exists in all TI map detections:
```yaml
ThreatIntelligenceIndicator
    | where TimeGenerated >= ago(ioc_lookBack) and ExpirationDateTime > now()
```
if Expiration is checked at first, I don't think it makes much sense to check it again later.
```yaml
on $left.EmailSenderAddress == $right.Caller
  | where AzureActivity_TimeGenerated < ExpirationDateTime
```
Do you mind if I remove the last part after the join?

Moreover, out of curiosity. please, when comparing to SecurityAlerts, I don't understand why doing:
```yaml
| extend MSTI = case(AlertName has "TI map" and VendorName == "Microsoft" and ProductName == 'Azure Sentinel', true, false)
| where MSTI == false
```
instead of
```yaml
| extend not(AlertName has "TI map" and VendorName == "Microsoft" and ProductName == 'Azure Sentinel)
```

ep3p:
I am going to close this Pull Request, and open 3 different ones with the changes I wanted to apply here.
Sorry, it's just this one got messy, with the new ones it will easier to discuss and accept or deny.

ep3p:
@PYB-Machine I think there is one TI map rule that uses ingestion time (https://github.com/Azure/Azure-Sentinel/blob/master/Detections/ThreatIntelligenceIndicator/DomainEntity_CommonSecurityLog.yaml), in case you want to apply the same to the rest.

PR #3480, opened 2021-11-19
Create new TI map detection against AzureFirewall

ep3p:
This rule might be useful.

It checks TI indicators against AzureFirewall Application Rule and AzureFirewall Network Rule logs, specifically against the Destination Address, **if it is a public address**.

Please, could you review if it is useful? 

**How can I know the ID an Analytics Rule template should have in the first line?**

### DIFF
--- a/Detections/ThreatIntelligenceIndicator/IPEntity_AzureFirewall.yaml
+++ b/Detections/ThreatIntelligenceIndicator/IPEntity_AzureFirewall.yaml
@@ -1,0 +1,63 @@
+id: tobeadded
+name: TI map IP entity to AzureFirewall
+description: |
+  'Identifies a match in AzureFirewall (NetworkRule & ApplicationRule Logs) from any IP IOC from TI'
+severity: Medium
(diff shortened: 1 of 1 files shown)


Detections/ThreatIntelligenceIndicator/IPEntity_AzureFirewall.yaml
```diff
@@ -0,0 +1,63 @@
+id: 0b904747-1336-4363-8d84-df2710bfe5e7
```
shainw, line 1: Added in the id for you.  You can use new-guid in PowerShell to generate your own...or use any GUID generator in the future. :)

shainw (approved):

Detections/ThreatIntelligenceIndicator/IPEntity_AzureFirewall.yaml
ep3p: Thank you very much for explaining me.

PR #3483, opened 2021-11-19
Check Unusual logins & Use correct time window in RunCommandUEBABreach.yaml

ep3p:
UEBAWindowStart has to be lower than UEBAWindowEnd, or there will be **nothing in between.**

I think the author wanted to write "UEBAWindowEnd = UEBAEventTime _+_ **6h**" instead of "UEBAEventTime _-_ **6h**"

Also, it does not check **unusual** activity in Behaviour Analytics table.

## Proposed Changes

  - Use "ActivityInsights.ActionUncommonlyPerformedByUser == "True"" as condition
  - Extend window of  UEBAevent
  - Reduce Query Period from 7 days to 2 days, could be further reduced.
  
  
 I think a thorough review of Sentinel Detections would be worth it.

### DIFF
--- a/Detections/MultipleDataSources/RunCommandUEBABreach.yaml
+++ b/Detections/MultipleDataSources/RunCommandUEBABreach.yaml
@@ -31,14 +31,14 @@ query: |
   | summarize StartTime=min(TimeGenerated), EndTime=max(TimeGenerated), max(CallerIpAddress), make_list(ActivityStatusValue) by CorrelationId, Authorization, Caller
   // Limit to Run Command executions that Succeeded
   | where list_ActivityStatusValue has "Succeeded"
-  // Extract data from the Authorization field, allowing us to later extract the Caller (UPN) and CallerIpAddress
+  // Extract data from the Authorization field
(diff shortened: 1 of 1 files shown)


thmcelro:
Changes look good, but I wouldn't make the Window lookback so large.

Seeing a Run Command action 6 hours prior to a UEBA event is more likely to be an FP, for this to work the login should be tagged as suspicious before they invoke Run Command. I think the window could start on the UEBA event time, however 1h lookback is a generous buffer.

I'd suggest using:

`| extend UEBAWindowStart = UEBAEventTime - 1h, UEBAWindowEnd = UEBAEventTime + 6h`

ep3p:
@thmcelro Done.

With a query period of 2 days, the same event can trigger 2 alerts (2 days in a row).

Maybe specifying:

```yaml
queryPeriod: 25h
...
let lookback = 1d;
AzureActivity
| TimeGenerated > ago(lookback)
```
could save computational resources.

In case you wanted to further change.

shainw (approved):
Chatted with Tom, looks good.  Approved.

PR #3486, opened 2021-11-19
Update Sysmon-AllVersions_Parser.txt

KustoKing:
Couldn't pivot because Type already existed
Not all columns in project statement exist so that line was removed.

Fixes #

## Proposed Changes

  -
  -
  -


shainw:
@Cyb3rWard0g and @ashwin-patil assigned.

shainw:
### Sysmon will be supported by ASIM in the near future to align on the ease of parsing by a singular method via normalization

PR #3487, opened 2021-11-20
Role Management App permissions grant and Admin Promotion by Adding member to Admin Directory Roles

Cyb3rWard0g:
Detection rule for:
* RoleManagement.ReadWrite.Directory permission grants
* Service principal being granted the Microsoft Graph RoleManagement.ReadWrite.Directory (application) permission before being used to add an Azure AD object or user account to an Admin directory role (i.e. Global Administrators). 

## References:
* https://docs.microsoft.com/en-us/graph/permissions-reference#role-management-permissions
* https://docs.microsoft.com/en-us/graph/api/directoryrole-post-members?view=graph-rest-1.0&tabs=http'
severity


Detections/AuditLogs/AdminPromoAfterRoleMgmtAppPermissionGrant.yaml
```diff
@@ -0,0 +1,74 @@
+id: f80d951a-eddc-4171-b9d0-d616bb83efdc
+name: Admin promotion after Role Management Application Permission Grant
+description: |
+  'This rule looks for a service principal being granted the Microsoft Graph RoleManagement.ReadWrite.Directory (application) permission before being used to add an Azure AD object or user account to an Admin directory role (i.e. Global Administrators). 
+  Ref : https://docs.microsoft.com/en-us/graph/permissions-reference#role-management-permissions, https://docs.microsoft.com/en-us/graph/api/directoryrole-post-members?view=graph-rest-1.0&tabs=http'
```
shainw, line 5: Remove the en-us from the reference links, this is causing the validation failure.

Detections/AuditLogs/AzureADRoleManagementPermissionGrant.yaml
```diff
@@ -0,0 +1,55 @@
+id: 1ff56009-db01-4615-8211-d4fda21da02d
+name: Azure AD Role Management Permission Grant
+description: |
+  'Identifies when the Microsoft Graph RoleManagement.ReadWrite.Directory (Delegated or Application) permission is granted to a service principal.
+  This permission allows an application to read and manage the role-based access control (RBAC) settings for your company's directory.
+  An adversary could use this permission to add an Azure AD object to an Admin directory role.
+  Ref : https://docs.microsoft.com/en-us/graph/permissions-reference#role-management-permissions
```
shainw, line 7: Remove the en-us from the reference links, this is causing the validation failure.

Detections/AuditLogs/AdminPromoAfterRoleMgmtAppPermissionGrant.yaml
```diff
@@ -0,0 +1,74 @@
+id: f80d951a-eddc-4171-b9d0-d616bb83efdc
+name: Admin promotion after Role Management Application Permission Grant
+description: |
+  'This rule looks for a service principal being granted the Microsoft Graph RoleManagement.ReadWrite.Directory (application) permission before being used to add an Azure AD object or user account to an Admin directory role (i.e. Global Administrators).
```
shainw, line 4: Can you add in why this is something to alert on?

Detections/AuditLogs/AdminPromoAfterRoleMgmtAppPermissionGrant.yaml
```diff
@@ -0,0 +1,74 @@
...
+  | extend displayName_ = tostring(TargetResources_modifiedProperties.displayName)
+  | where displayName_ =~ "AppRole.Value"
+  | extend AppRole = tostring(parse_json(tostring(TargetResources_modifiedProperties.newValue)))
+  | where AppRole contains "RoleManagement.ReadWrite.Directory"
+  | extend InitiatingApp = tostring(parse_json(tostring(InitiatedBy.app)).displayName)
+  | extend Initiator = iif(isnotempty(InitiatingApp), InitiatingApp, tostring(parse_json(tostring(InitiatedBy.user)).userPrincipalName))
+  | extend Target = tostring(parse_json(tostring(TargetResources.modifiedProperties[4].newValue)))
+  | extend TargetId = tostring(parse_json(tostring(TargetResources.modifiedProperties[3].newValue)))
+  | summarize by bin(TimeGenerated, 1h), OperationName, Initiator, Target, TargetId, Result
+  | join (
```
shainw, line 39: Please specify the join kind.  Default is innerunique which can reduce the list of full results.

Detections/AuditLogs/AdminPromoAfterRoleMgmtAppPermissionGrant.yaml
```diff
@@ -0,0 +1,74 @@
...
+  | where LoggedByService =~ "Core Directory"
+  | where Category =~ "ApplicationManagement"
+  | where AADOperationType =~ "Assign"
+  | where ActivityDisplayName =~ "Add app role assignment to service principal"
+  | mv-expand TargetResources
+  | mv-expand TargetResources.modifiedProperties
+  | extend displayName_ = tostring(TargetResources_modifiedProperties.displayName)
+  | where displayName_ =~ "AppRole.Value"
+  | extend AppRole = tostring(parse_json(tostring(TargetResources_modifiedProperties.newValue)))
+  | where AppRole contains "RoleManagement.ReadWrite.Directory"
```
shainw, line 33: Seems like we could as a "has" for RoleManagement.ReadWrite.Directory against the unexpanded TargetResources field before the expensive mv-expand to improve perf?  Using "has" does seem to limit the return set in my testing which will improve perf after with mv-expand.  Example:

```yaml
  AuditLogs
  | where LoggedByService =~ "Core Directory"
  | where Category =~ "ApplicationManagement"
  | where AADOperationType =~ "Assign"
  | where ActivityDisplayName =~ "Add app role assignment to service principal"
  | where TargetResources has "RoleManagement.ReadWrite.Directory"
  | mv-expand TargetResources
  | mv-expand TargetResources.modifiedProperties
  | extend displayName_ = tostring(TargetResources_modifiedProperties.displayName)
  | where displayName_ =~ "AppRole.Value"
  | extend AppRole = tostring(parse_json(tostring(TargetResources_modifiedProperties.newValue)))
  | where AppRole has "RoleManagement.ReadWrite.Directory"
```

Detections/AuditLogs/AdminPromoAfterRoleMgmtAppPermissionGrant.yaml
```diff
@@ -0,0 +1,74 @@
...
+    | where LoggedByService =~ "Core Directory"
+    | where Category =~ "RoleManagement"
+    | where AADOperationType in ("Assign", "AssignEligibleRole")
+    | where ActivityDisplayName has_any ("Add eligible member to role", "Add member to role")
+    | mv-expand TargetResources
+    | mv-expand TargetResources.modifiedProperties
+    | extend displayName_ = tostring(TargetResources_modifiedProperties.displayName)
+    | where displayName_ =~ "Role.DisplayName"
+    | extend RoleName = tostring(parse_json(tostring(TargetResources_modifiedProperties.newValue)))
+    | where RoleName contains "Admin"
```
shainw, line 50: You should be able to do the same here with "has" versus "contains"

Detections/AuditLogs/AdminPromoAfterRoleMgmtAppPermissionGrant.yaml
```diff
@@ -0,0 +1,74 @@
...
+entityMappings:
+  - entityType: Account
+    fieldMappings:
+      - identifier: FullName
+        columnName: ServicePrincipal
+  - entityType: Account
+    fieldMappings:
+      - identifier: FullName
+        columnName: AccountCustomEntity
+version: 1.0.1
```
shainw, line 73: New so version should be 1.0.0

Detections/AuditLogs/AzureADRoleManagementPermissionGrant.yaml
```diff
@@ -0,0 +1,55 @@
...
+  | where LoggedByService =~ "Core Directory"
+  | where Category =~ "ApplicationManagement"
+  | where AADOperationType =~ "Assign"
+  | where ActivityDisplayName has_any ("Add delegated permission grant","Add app role assignment to service principal")
+  | mv-expand TargetResources
+  | mv-expand TargetResources.modifiedProperties
+  | extend displayName_ = tostring(TargetResources_modifiedProperties.displayName)
+  | where displayName_ has_any ("AppRole.Value","DelegatedPermissionGrant.Scope")
+  | extend Permission = tostring(parse_json(tostring(TargetResources_modifiedProperties.newValue)))
+  | where Permission contains "RoleManagement.ReadWrite.Directory"
```
shainw, line 36: Same here with usage of "has" versus "contains"

Detections/AuditLogs/AzureADRoleManagementPermissionGrant.yaml
```diff
@@ -0,0 +1,55 @@
...
+entityMappings:
+  - entityType: Account
+    fieldMappings:
+      - identifier: FullName
+        columnName: Initiator
+  - entityType: Account
+    fieldMappings:
+      - identifier: FullName
+        columnName: AccountCustomEntity
+version: 1.0.1
```
shainw, line 54: New so version 1.0.0

Detections/AuditLogs/AdminPromoAfterRoleMgmtAppPermissionGrant.yaml
```diff
@@ -0,0 +1,74 @@
...
+    | extend Target = iif(isnotempty(TargetUser), TargetUser, tostring(TargetResources.displayName))
+    | extend TargetType = tostring(TargetResources.type)
+    | extend TargetId = tostring(TargetResources.id)
+    | summarize by bin(TimeGenerated, 1h), OperationName,  RoleName, Initiator, InitiatorId, Target, TargetId, TargetType, Result
+  ) on $left.TargetId == $right.InitiatorId
+  | extend TimeRoleMgGrant = TimeGenerated, TimeAdminPromo = TimeGenerated1, ServicePrincipal = Initiator1, ServicePrincipalId = InitiatorId,
+    TargetObject = Target1, TargetObjectId = TargetId1, TargetObjectType = TargetType
+  | where TimeRoleMgGrant < TimeAdminPromo
+  | project TimeRoleMgGrant, TimeAdminPromo, RoleName, ServicePrincipal, ServicePrincipalId, TargetObject, TargetObjectId, TargetObjectType
+  | extend timestamp = TimeRoleMgGrant, AccountCustomEntity = TargetObject
```
shainw, line 63: No need for timestamp and AccountCustomEntity any longer here but not really hurting anything.   You can also just point to TargetObject in the entity mappings below.

Detections/AuditLogs/AzureADRoleManagementPermissionGrant.yaml
```diff
@@ -0,0 +1,55 @@
...
+  | extend Permission = tostring(parse_json(tostring(TargetResources_modifiedProperties.newValue)))
+  | where Permission contains "RoleManagement.ReadWrite.Directory"
+  | extend InitiatingApp = tostring(parse_json(tostring(InitiatedBy.app)).displayName)
+  | extend Initiator = iif(isnotempty(InitiatingApp), InitiatingApp, tostring(parse_json(tostring(InitiatedBy.user)).userPrincipalName))
+  | extend Target = tostring(parse_json(tostring(TargetResources.modifiedProperties[4].newValue)))
+  | extend TargetId = iif(displayName_ =~ 'DelegatedPermissionGrant.Scope',
+    tostring(parse_json(tostring(TargetResources.modifiedProperties[2].newValue))),
+    tostring(parse_json(tostring(TargetResources.modifiedProperties[3].newValue))))
+  | summarize by bin(TimeGenerated, 1h), OperationName, Initiator, Target, TargetId, Result
+  | extend AccountCustomEntity = Target
```
shainw, line 44: Again, no need for this, you can point at Target in the entity mapping below.

shainw (changes requested):

Detections/AuditLogs/AdminPromoAfterRoleMgmtAppPermissionGrant.yaml
Cyb3rWard0g: Thank you Shain! I always wondered what those error messages were :) Fixed.

Detections/AuditLogs/AdminPromoAfterRoleMgmtAppPermissionGrant.yaml
Cyb3rWard0g: This is the current description now:

This rule looks for a service principal being granted the Microsoft Graph RoleManagement.ReadWrite.Directory (application) permission before being used to add an Azure AD object or user account to an Admin directory role (i.e. Global Administrators).
  This is a known attack path that is usually abused when a service principal already has the AppRoleAssignment.ReadWrite.All permission granted. This permission Allows an app to manage permission grants for application permissions to any API.
  A service principal can promote itself or other service principals to admin roles (i.e. Global Administrators). This would be considered a privilege escalation technique.

Detections/AuditLogs/AdminPromoAfterRoleMgmtAppPermissionGrant.yaml
Cyb3rWard0g: ok. Inner JOIN works in here since we want both actions to execute. I will specify inner join unique.

Detections/AuditLogs/AzureADRoleManagementPermissionGrant.yaml
Cyb3rWard0g: Done. Thank you Shain!

Detections/AuditLogs/AdminPromoAfterRoleMgmtAppPermissionGrant.yaml
Cyb3rWard0g: Ahhh I use a few templates and I forget to change the versions. Sorry! Thank you!

Detections/AuditLogs/AdminPromoAfterRoleMgmtAppPermissionGrant.yaml
Cyb3rWard0g: This will not work in here because `has` looks for exact string values. With `contains` we want to capture `X Admin Group` and `Global Administrators`. I tested it and `has` did not show any results. However, `contains` did.

Detections/AuditLogs/AzureADRoleManagementPermissionGrant.yaml
Cyb3rWard0g: Done

Detections/AuditLogs/AzureADRoleManagementPermissionGrant.yaml
Cyb3rWard0g: Done. Thank you!

Cyb3rWard0g:
Hello @shainw ! let me know if there is anything I need to add or modify :) Thank you for taking the time!

Detections/AuditLogs/AdminPromoAfterRoleMgmtAppPermissionGrant.yaml
shainw: Doesn't look like this changed - if you want inner, you have specify it here - join kind=inner.

Detections/AuditLogs/AdminPromoAfterRoleMgmtAppPermissionGrant.yaml
shainw: Got it, so because "Admin" is part of "Administrator" then 'has' misses it.

shainw (changes requested):
Looks like the join kind did not get changed, update that and then good to go.

shainw (approved):

PR #3505, opened 2021-11-22
Create AnomalousAzureOperationModel.yaml

thmcelro:
Create anomaly model hunting query to detect suspicious operations being executed in Azure.


shainw (approved):
Approved, build was passing before parallelism issue hit us again.  Okay to merge.

PR #3508, opened 2021-11-22
Use different time periods on UserAccounts-CABlockedSigninSpikes.yaml

ep3p:
If the query period is 1d, you won't get results from 14 days ago that you try to query.

Fixes #

Not getting any meaningful results due to several time periods.

HourlyCount was computed in timeframes of 1 day, maybe this was intended, but then it shouldn't be called HourlyCount or compared to DateHour (which was computed in timeframes of 1 hour).

## Proposed Changes

  - Use query period of 14 days, as other Detection Rules have it when they search events from 14 days ago.
  - Reuse ```let allSignins = union isfuzzy=true aadSignin, aadNonInt ;``` or remove it.
  - Use ```TimeSeriesAlerts
  | where TimeGenerated > startofday(ago(timeframe)) ``` or is it intended to be 2d when the queryfrequency is 1d?
  - Compute HourlyCount with timeframe of 1h, instead of 1d
  - several other changes to make it work
  
  The anomalies were intented to be computed by hour or by day? This alert could be noisy.


ep3p:
I see now **Privileged Accounts - Sign in Failure Spikes** has similar logic. Maybe I will commit in this branch too.

shainw:
@ashwin-patil - This needs your review to make sure the changes are in line with intent of the original idea.

ep3p:
I would not recommend accepting this Pull Request in the current state, because the detection rule may be noisy.

I think some further changes, like increasing a threshold or summarizing the results could be beneficial. I have not commited these hypothetical changes, because I think you will have better expertise/datasets for this noise problem than me.

NikTripathi:
@ashwin-patil Can you please check and approve?

ashwin-patil:
> I would not recommend accepting this Pull Request in the current state, because the detection rule may be noisy.
> 
> I think some further changes, like increasing a threshold or summarizing the results could be beneficial. I have not commited these hypothetical changes, because I think you will have better expertise/datasets for this noise problem than me.

Thank you for the changes and notes. It seems hourly timeseries do not have consistent values (for many are 0) hence it is flagging lot of anomalies. For now, i have changed it to daily timeseries, increased baseline thresholds to 50 which has reduced results drastically.

aprakash13 (approved):
Thanks @ep3p & @ashwin-patil .

PR #3523, opened 2021-11-23
Create MissingDCHearbeat.yaml

ep3p:
This rule should detect when the most recent Heartbeat of a specified DomainController happened more than 1 hour ago (or the time specified).

**It only triggers once**, while the Heartbeat has not been reestablished.

Please, could you check if this is useful? At least it has been in my tenant, with legacy Log Analytics (Microsoft Monitoring Agent).


ep3p:
Which is the name of the new tactic ```Impair Process Control```?

shainw (approved):

PR #3530, opened 2021-11-24
Exclude TI map Alerts from TI Map Security Alert

shainw:
Fixes #
TI mapping based on entities showing up in Security Alerts was causing 2 rules to fire over and over because their own Alert was retriggering.

## Proposed Changes
  - Exclude Alerts based on rules from MS that have names with TI Map in them
 ```yaml
| extend MSTI = case(AlertName has "TI map" and VendorName == "Microsoft" and ProductName == 'Azure Sentinel', true, false)
| where MSTI == false
```


ashwin-patil (dismissed):

ashwin-patil (commented):
Looks like version needs to be updated, rest looks good.

petebryan (dismissed):

PR #3531, opened 2021-11-24
Updated texts for Sysmon for Linux connector

oshezaf:
Texts required updating (it actually included "Sysmon for linux link: update when available.")


oshezaf:
@preetikr , @v-jayakal ? We need to update those texts.

NikTripathi (approved):

Issue #3455, opened 2021-11-16
Rule error Chia_Crypto_Mining - Domain, Process, Hash and IP IOCs - June 2021


anfisher1967:
Rule creation fails with Set rule logic step.

anfisher1967:
![Chia](https://user-images.githubusercontent.com/33937240/142067859-9da6f92a-3a3a-44e1-a5b8-d1751b461feb.jpg)

shainw:
@anfisher1967 - test out the query in this PR, ProcessId was never really there, it should be File and FilePath.  These are not part of the automap, so you will need to add them specifically, but they are now easily found - FileCustomEntity and FilePathCustomEntity columns.  You will still need to add in the Value mapping for File Hash as this is not automapped either.  You will map it to the FileHashCustomEntity column.

PR #3542, opened 2021-11-25
Update ChiaCryptoMining.yaml

shainw:
Based on issue #3455: Chia rule entity mapping issues - 

Fixes #
 - Fixing entity mappings in general
 - Changing process to file as we do not have a process name or process path entity mapping, it is done thru file.
 - Adding in FilePath and we are missing a column value to indicate the hash algorithm so added that in.  
 - UX will look for xxxCustomEntity to auto map some values as part of the entity mapping feature so making sure those are available still. 
 - Reducing extending names so it is done once outside of the join, this also reduces characters below 10000 which the rule creator errors on if more than.
 - Moving sysmon comment to top and only having one instance
 - Added tostring() for items in this format - EventDetail.[4].["#text"]


shainw:
@aprakash13 - Please have a look at this one, test to make sure we are good and approve.  Would like to get this merged asap to fix a bug.

PR #3545, opened 2021-11-25
ASim fixes

rjaakke:
Fixes #
- Fixed ASim DnsEmpty ARM by removing the  duplicate EventSeverity column
- Fixed Barium and Iridium rules to call imDns correctly with parameter response_has_any_prefix

## Proposed Changes
  - Changed watchlist alias of ASimDisabledParsers in ARM to be plural as since alle rules call _GetWatchlist('ASimDisabledParsers')


oshezaf:
Thanks @rjaakke! Your conrtibution to ASIM is very welcomed. Keep it coming.

@YaronFruchtmann is the expert and will review, but two comments from me:
- Are you sure Dvc is the value you want to map to host? It includes the DNS server information rather than the client issuing the DNS request. I think that SrcIpAddr is a better value. DNS events usually do not have hostname information for the source system.
- Thanks for noticing the duplicate EventSeverity. I think @YaronFruchtmann fixed it a few days back, so you might be working not of the most recent branch. 

~ Ofer

rjaakke:
Hi Ofer,

I reverted host mapping.
The duplicate EventSeverity was indeed fixed in the parser but the ARM template wasn't updated.

Detections/MultipleDataSources/BariumIPIOC112020.yaml
```diff
@@ -87,7 +87,7 @@ query:  |
   | where  DestinationIPAddress has_any (IPList)  
   | extend timestamp = TimeGenerated, IPCustomEntity = DestinationIPAddress, HostCustomEntity = Host 
   ), 
-  (imDns (response_has_any=IPList)
+  (imDns (response_has_ipv4=IPList)
```
YaronFruchtmann, line 90: Thanks for noticing an notifying the error. since we are passing a dynamic list (IPList) I believe the fix should be: response_has_any_prefix.

Parsers/ASim/Watchlists/ASimDisabledParsers.json
```diff
@@ -11,12 +11,12 @@
...
       "resources": [
           {
-          "name": "[concat(parameters('workspaceName'), '/Microsoft.SecurityInsights/ASimDisabledParser')]",
+          "name": "[concat(parameters('workspaceName'), '/Microsoft.SecurityInsights/ASimDisabledParsers')]",
           "type": "Microsoft.OperationalInsights/workspaces/providers/Watchlists",
           "kind": "",
           "properties": {
-              "displayName": "ASim Disabled Parser",
-              "source": "ASimDisabledParser.csv",
+              "displayName": "ASim Disabled Parsers",
```
YaronFruchtmann, line 18: I accept

Parsers/ASim/Watchlists/ASimDisabledParsers.json
```diff
@@ -11,12 +11,12 @@
     },
       "resources": [
           {
-          "name": "[concat(parameters('workspaceName'), '/Microsoft.SecurityInsights/ASimDisabledParser')]",
+          "name": "[concat(parameters('workspaceName'), '/Microsoft.SecurityInsights/ASimDisabledParsers')]",
```
YaronFruchtmann, line 14: Since this doesn't cause any interruption in functionality, and a change might cause problems for users who already deployed the watchlist if deploying again - I prefer not to change the resource name

Parsers/ASim/Watchlists/ASimDisabledParsers.json
```diff
@@ -11,12 +11,12 @@
...
           "type": "Microsoft.OperationalInsights/workspaces/providers/Watchlists",
           "kind": "",
           "properties": {
-              "displayName": "ASim Disabled Parser",
-              "source": "ASimDisabledParser.csv",
+              "displayName": "ASim Disabled Parsers",
+              "source": "ASimDisabledParsers.csv",
```
YaronFruchtmann, line 19: for the same reason as above, this has no functional impact, and a change might. I prefer not to make that change at the moment,

YaronFruchtmann (commented):
Thanks!

Parsers/ASim/Watchlists/ASimDisabledParsers.json
rjaakke: There are 13 Asim parsers that use _GetWatchlist('ASimDisabledParsers') so providing a watchlist template with a different name as guidance seems confusing to me. People who might have deployed the watchlist now might not even know the parsers are looking for a different watchlist.

Detections/MultipleDataSources/BariumIPIOC112020.yaml
rjaakke: Fixed it.

Parsers/ASim/Watchlists/ASimDisabledParsers.json
YaronFruchtmann: You are right! Sorry!

YaronFruchtmann: you are right! sorry!

YaronFruchtmann (approved):
Thanks!

PR #3549, opened 2021-11-25
Use correct time columns to check Expiration and to summarize in TI Map Detections.

ep3p:
This Pull Request comes from a closed unmerged pull request that got messy (https://github.com/Azure/Azure-Sentinel/pull/3477)

Fixes #

Saving some unnecessary conditions.

Expiration Date is already getting compared at
```yaml
ThreatIntelligenceIndicator
| where TimeGenerated >= ago(ioc_lookBack) and ExpirationDateTime > now()
```

2nd summarize was getting its TimeGenerated from the event table, not from the Indicator Table.

## Proposed Changes

  - Remove conditions about the time of the indicator, as they are already fulfilled.
  - Use correct column in 2nd summarize.
  - Use project-rename for Event columns that conflict with Indicator columns.

### DIFF
--- a/Detections/ThreatIntelligenceIndicator/DomainEntity_CommonSecurityLog.yaml
+++ b/Detections/ThreatIntelligenceIndicator/DomainEntity_CommonSecurityLog.yaml
@@ -17,6 +17,7 @@ triggerThreshold: 0
 tactics:
   - Impact
 query: |
+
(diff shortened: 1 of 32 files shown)


shainw:
As I said in other comments, we are using this PR with multiple changes across all TI - #3571.  Thanks much.

PR #3550, opened 2021-11-25
Update condition of "Alert not generated by TI Map" 

ep3p:
Fixes #

Saving one step in the tabular expression.

## Proposed Changes

  - Using ```not``` instead of ```case```.


shainw:
Appreciate this, but the preference is to leave this as an option for users.  You are welcome to modify your local version in Azure Sentinenl. :)

ep3p:
@shainw Thank you very much for explaining me and everything.

PR #3551, opened 2021-11-25
Specify join kind in TI Map Detections

ep3p:
This Pull Request comes from a closed unmerged Pull Request that got messy (https://github.com/Azure/Azure-Sentinel/pull/3477).

## Proposed Changes

  - Specify join kind as per indication of @shainw , nowadays the default kind is already ```innerunique```


shainw:
@ep3p - I have done a broader PR that includes making this and other items consistent across all TI queries.  We will use that and close this one - #3571

PR #3552, opened 2021-11-26
Update Details in PrivlegedRoleAssignedOutsidePIM.yaml

ep3p:
## Proposed Changes

  - Write the full word of "Privileged"

**ALSO**, please, I have not seen this operation ```Add member to role (permanent)```, but ```Add member to role``` as specified in https://github.com/Azure/Azure-Sentinel/blob/master/Detections/AuditLogs/UseraddedtoPrivilgedGroups.yaml

Could you check if this ActivityDisplayName ```Add member to role (permanent)``` really exists?


aprakash13 (approved):
Changes looks good. Approving it now.

PR #3553, opened 2021-11-26
Increase Queryperiod in AuthenticationMethodsChangedforPrivilegedAccount.yaml

ep3p:
IdentityInfo does not have all records every 2 hours.

Fixes #

Not getting wanted results.

## Proposed Changes

  - Use greater query period, ideally should be 21 days as I think IdentityInfo is resynced every 21 days.


aprakash13 (approved):
Changes looks good. Approving this now.

PR #3556, opened 2021-11-26
Detection : External User Added to Team and Immediately Uploads File

samikroy:
## Proposed Changes

This detection identifies an external user is added to a Team or Teams chat
 and within 1 minute of being added upload a file via the chat.


ep3p:
You may want to specify the kind of join to not lose events of the left table.

samikroy:
> You may want to specify the kind of join to not lose events of the left table.

Thank you for the review @ep3p . 
Have added inner as a join kind to avoid default innerunique picking a random one.
Please let know further.

shainw:
Adding @petebryan as he did the Hunting query that this is based on and he may have tested to see if this type of detection was too noisy for a detection - https://github.com/Azure/Azure-Sentinel/blob/master/Hunting%20Queries/OfficeActivity/UserAddToTeamsAndUploadsFile.yaml..  @petebryan - thoughts?

samikroy:
> Adding @petebryan as he did the Hunting query that this is based on and he may have tested to see if this type of detection was too noisy for a detection - https://github.com/Azure/Azure-Sentinel/blob/master/Hunting%20Queries/OfficeActivity/UserAddToTeamsAndUploadsFile.yaml.. @petebryan - thoughts?

@petebryan  - please share your thoughts

petebryan:
@samikroy so when I first created this we didn't have an official Teams connector publicly available yet so we had limited data to determine FP rate. However in the data did show this was potentially quite noisy. It would be good to re-evaluate to see if that is still the case.

samikroy:
> @samikroy so when I first created this we didn't have an official Teams connector publicly available yet so we had limited data to determine FP rate. However in the data did show this was potentially quite noisy. It would be good to re-evaluate to see if that is still the case.

Agree @petebryan , this detection now applies few more filters

External User Added > File Uploaded > File is accessed by many users > External user is removed.
Please have a look and let me know 
Thank you.

shainw (dismissed):
I did some testing and looks like looking only for #EXT# (external users) and checking for within 1m has very low results.  Approving and we will watch if alerts start to flood for some customers.

shainw (approved):

PR #3564, opened 2021-11-29
Remove unnecessary condition in UseraddedtoPrivilgedGroups.yaml

ep3p:
```Add member to role in PIM requested (permanent)``` is ```LoggedByService == "PIM"```, not by ```Core Directory```

Fixes #

I think this rule has never detected the operation ```Add member to role in PIM requested (permanent)```, or at least not nowadays.

## Proposed Changes

  - The rule already filters by OperationName, it should not filter by LoggedByService if you want to detect ```Add member to role in PIM requested (permanent)```


aprakash13 (approved):
Changes look good. Approving this.

PR #3571, opened 2021-11-29
Updating TI queries based on feedback and discussions

shainw:
Discussions as part of - #3477 - This includes generic changes that need to be done.  Customers may still want to customize.

## Proposed Changes

  -  Specifying join kind explicitly and commenting as to why we use innerunique. 
  - Making changes to filtering after the join and fixing the 2nd argmax to use the activity time instead of indicator time as you want the most recent activity that matches
  - Include the primary join token value in the 2nd argmax() so we do not lose visibility on the friendly name of the IOC and only use the IndicatorId.
  - Added in some additional entity mappings


petebryan (approved):
All make sense

PR #3595, opened 2021-12-01
Use decoded command in the regex comparison in powershell_empire.yaml

ep3p:
Fixes #

Decoded commands were not being compared against the Regex.

```Set-Speaker(\$Volume){\$wshShell``` parentheses were not correctly escaped.

Please, could you review the query against a good dataset?


shainw (approved):
These changes impact performance very highly, and I am pretty sure this is why we used a function to begin with.  I will updated.

ep3p:
@shainw @aprakash13

Please, could I know why it was merged into master without a where condition against the regex like ```| where EffectiveCommand matches regex regexEmpire```? would not trigger a lot of alerts with any command?

PR #3597, opened 2021-12-01
Update link in AzurePortalSigninfromanotherAzureTenant.yaml

ep3p:
Fixes #

Outdated link.

## Proposed Changes

  - There should be another way for this to be updated. last time it was only 21 days ago.

I know there is an Azure Function in the repository, to update regularly a Watchlist with Cloud Ranges.

Maybe you could provide a Watchlist with Azure Ranges in all Sentinel workspaces, and this rule would use that hypotetical Watchlist.

Or create a permanent link that updates its content by itself.


shainw (approved):

shainw:
@ep3p - we will look at how we can resolve going forward, thanks for the suggestions.

ep3p:
@timpailthorpe the link renews every 21 days. Check commented lines 21 and 22 in the template.

The new link is https://download.microsoft.com/download/7/1/D/71D86715-5596-4529-9B13-DA13A5DE5B63/ServiceTags_Public_20211220.json

PR #3607, opened 2021-12-03
Update timeframe & details in SuspiciousServicePrincipalcreationactivity.yaml

ep3p:
Fixes #

1.
If the detection rule executes between these operations have started but not finished, before the deletion, it will never trigger.
A wait_period could be used, the query frequency can be the same.

2.
```
Update application – Certificates and secrets management 
Update application - Certificates and secrets management
```
do not have the same hyphen. And **here there might be a background problem**, in my tenant this OperationName contains a non-ASCII character. If I search by the ASCII version, it can't find the results, but if I use the non-ASCII version then the detection rule does not pass the nonasciivalidation...

image

3.
I don't think it is needed to ```summarize by AppID, AppID```
## Proposed Changes

  - Fix above details

I understand this detection is wanted to trigger **only if the service principal is deleted** and all the operations have ocurred, otherwise a ```kind=leftouter``` should be needed.

### DIFF
--- a/Detections/AuditLogs/SuspiciousServicePrincipalcreationactivity.yaml
+++ b/Detections/AuditLogs/SuspiciousServicePrincipalcreationactivity.yaml
@@ -44,7 +44,6 @@ query: |
     | extend ipAddress_deleter = tostring(parse_json(tostring(InitiatedBy.user)).ipAddress);
   let account_credentials =
     AuditLogs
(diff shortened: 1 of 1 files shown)


NikTripathi:
@aprakash13 Can you please review and update? Thanks.

aprakash13 (approved):
Thanks @ep3p.

PR #3616, opened 2021-12-03
added snowflake content

spsocprime:
## Proposed Changes

Added analytic content for Snowflake data connector


v-sabiraj:
Hey @spsocprime, We have resolved the issue for PR Validations failure and request you to please update your PR branch with the latest changes from **Azure:master** so that the PR Validations can be executed.... Thanks!!!

v-sabiraj:
@aprakash13 please review Analytic Rules, thanks.

Solutions/Snowflake/Analytic Rules/SnowflakeUserAddAdminPrivileges.yaml
```diff
@@ -0,0 +1,33 @@
+id: f258fa0c-e26c-4e2b-94fb-88b6cef0ca6e
```
ritika-msft, line 1: There should not be 2 templates with the same ID, but the id f258fa0c-e26c-4e2b-94fb-88b6cef0ca6e is duplicated. Please update it.

ritika-msft:
Please fix workbook queries. Thanks.

image
image
image

spsocprime:
> Please fix workbook queries. Thanks.
> 
> image image image

Do you have custom table with Snowflake events in your environment because in our Sentinel deployment everything works fine
image

ritika-msft:
@spsocprime 
Yes we had the custom table "Snowflake" in our environment. But as highlighted in workbook query failures the issue is with the column names. 

**For ex**: Query used in workbook is "Snowflake| summarize s=sum(EXECUTION_TIME_d)| project s_time = strcat(s/1000, ' sec')" which fails with below msg and same is highlighted in workbook query results 
image

However when I correct the column the query is executed successfully. Below is the example:
Snowflake| summarize s=sum(EXECUTION_TIME_d_d)| project s_time = strcat(s/1000, ' sec')

Same is the issue with all other workbook queries as highlighted in previous screenshots.

spsocprime:
image
Here is what we have in our environment (this is default field name ingested by Snowflake data connector). Seems like field names  in your environment are not named as they should be by default

shainw (commented):
Queries look good.  I would recommend maybe looking at custom alert names and descriptions to reduce the number of queries - https://github.com/Azure/Azure-Sentinel/wiki/Query-Style-Guide#alertdetailsoverride

ritika-msft:
> image Here is what we have in our environment (this is default field name ingested by Snowflake data connector). Seems like field names in your environment are not named as they should be by default

I verified that the table schema is having issues in our environment [probably due to redundant ingestion], when I run the queries by updating column name, the workbook looks good. We can merge the PR once comment from @shainw are resolved.

ritika-msft (approved):
Workbook looks good. Approving the PR for workbooks.

shainw (approved):

PR #3648, opened 2021-12-08
Updating Purview Workbook

rheabansal:
Fixes #

## Proposed Changes

  -
  -
  -


v-jayakal:
Fallback resource id has hard coded subscription id, resulting into validation failure. please fix. Thanks.

v-jayakal (changes requested):
SInce we have Azure Purview solution, can you please add this workbook to solution instead of a standalone workbook?

libarson:
We will also be adding this workbook to the solution. In addition, we wanted to make this update to the existing workbook because we changed the schema and we don't want the workbook to break for our existing private preview customers.

PR #3664, opened 2021-12-09
Updated ansi-regex version to fix Dependabot Security Alert

v-rucdu:
Fixes #

## Proposed Changes

  -
  -
  -


v-rucdu:
KqlTest is failing because there is a recent merge in analytics rule which uses **_GetWatchList** as an function in one of the query... Thanks!!!

Tichandr (approved):
This looks good with latest version of ansi-regex@6.0.1

Issue #3669, opened 2021-12-09
Logic error in AWSConsoleAADCorrelation rule


javiersoriano:
**Describe the bug**
Detection AWSConsoleAADCorrelation located here: https://github.com/Azure/Azure-Sentinel/blob/master/Detections/MultipleDataSources/AWSConsoleAADCorrelation.yaml it's supposed to look for AWS login failures but success in AAD, but the rule says:

` where ResultType !in ("0", "50125", "50140")`

which means to look for non successful login events. It seems like the `!` needs to be removed from line 40 in the rule logic: https://github.com/Azure/Azure-Sentinel/blob/ff558fe2d557af6b659b5ba97e5cb3535444e64e/Detections/MultipleDataSources/AWSConsoleAADCorrelation.yaml#L40

PR #3672, opened 2021-12-10
Fix for #3669

samikroy:
Fix for #3669 and addition of AADNonInteractiveUserSignInLogs to the query.


samikroy:
@javiersoriano - please review here.

javiersoriano:
@shainw @petebryan please check if this the new logic is correct

shainw:
@javiersoriano - I don't think we want to include non-interactive signins in Azure.  The AWS login is interactive to the "console".

samikroy:
> @javiersoriano - I don't think we want to include non-interactive signins in Azure. The AWS login is interactive to the "console".

@shainw & @javiersoriano - Updated the query as per the suggestion & review comments.
Thank you.

shainw (approved):

PR #3677, opened 2021-12-10
Use different time periods in PrivilegedAccountsSigninFailureSpikes.yaml

ep3p:
**SIMILAR PULL REQUEST AS https://github.com/Azure/Azure-Sentinel/pull/3508**

If the query period is 1d, you won't get results from 14 days ago that you try to query.

Fixes #

Not getting any meaningful results due to several time periods.

HourlyCount was computed in timeframes of 1 day, maybe this was intended, but then it shouldn't be called HourlyCount or compared to DateHour (which was computed in timeframes of 1 hour).

## Proposed Changes

  - Use query period of 14 days, as other Detection Rules have it when they search events from 14 days ago.
  - Reuse ```let allSignins = union isfuzzy=true aadSignin, aadNonInt ;``` or remove it.
  - Use ```TimeSeriesAlerts
  | where TimeGenerated > startofday(ago(timeframe)) ``` or is it intended to be 2d when the queryfrequency is 1d?
  - Compute HourlyCount with timeframe of 1h, instead of 1d
  - several other changes to make it work
  
  The anomalies were intented to be computed by hour or by day? This alert could be noisy.

### DIFF
--- a/Detections/MultipleDataSources/PrivilegedAccountsSigninFailureSpikes.yaml
+++ b/Detections/MultipleDataSources/PrivilegedAccountsSigninFailureSpikes.yaml
@@ -13,7 +13,7 @@ requiredDataConnectors:
     dataTypes:
       - AADNonInteractiveUserSignInLogs
 queryFrequency: 1d
-queryPeriod: 1d
+queryPeriod: 14d
 triggerOperator: gt
 triggerThreshold: 0
 tactics:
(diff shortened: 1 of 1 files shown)


shainw:
@ep3p - I think you may be changing the intent of the query, but will need @ashwin-patil to review as he is the original author.

ep3p:
Just to shed some light, ```AnomalyHours``` ***always*** has the format ```date 01:00:00```, but ```DateHour``` has the format ```date XX:00:00```.

Unless the SignInLogs happen between 01:00 and 02:00 at night, it will never trigger.

NikTripathi:
@ashwin-patil Can you please check and approve?

ashwin-patil (approved):
This looks good. I have tested with the changes and happy with results. Aggregations are helping to reduce results without loosing context. The timeseries alerts are relatively small but since we are joining again back on base dataset to pull all the related logins it increases result set.

PR #3690, opened 2021-12-12
Pushing query to consume IP IOC list

shainw:
## Proposed Changes

  - adding query to lookup IP associated with attack infra for Log4j vuln, aka Log4Shell
  - https://cve.mitre.org/cgi-bin/cvename.cgi?name=2021-44228
  -


oshezaf:
@shainw : can you share the PR that failed validation because of imDns? imDns is defined as should have worked and I want to look into that.

tagging @YaronFruchtmann  as well.

PR #3711, opened 2021-12-14
Fix Action not being parsed & Compare Allowed Public Source Address in IPEntity_AzureFirewall.yaml

ep3p:
Fixes #

Action column was not being parsed due to Ungreedy regex flag.

Source Address may be a public address, and it could also be compared to TI Indicators, specially when the traffic was not denied (publicly exposed machines).
____
Maybe you prefer the Source Address to not be compared. In my case it did not add much noise except for a machine that serves as a WAF.

### DIFF
--- a/Detections/ThreatIntelligenceIndicator/IPEntity_AzureFirewall.yaml
+++ b/Detections/ThreatIntelligenceIndicator/IPEntity_AzureFirewall.yaml
@@ -48,6 +48,7 @@ query: |
       | project-rename AzureFirewall_TimeGenerated = TimeGenerated
       )
       on $left.TI_ipEntity == $right.RemoteIP
+  | where AzureFirewall_TimeGenerated < ExpirationDateTime
   | summarize AzureFirewall_TimeGenerated = arg_max(AzureFirewall_TimeGenerated, *) by IndicatorId, RemoteIP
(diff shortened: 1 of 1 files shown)


Detections/ThreatIntelligenceIndicator/IPEntity_AzureFirewall.yaml
```diff
@@ -38,16 +38,18 @@ query: |
...
-      | extend SourceAddress = extract(@'([\.0-9]+)(:[\.0-9]+)?',1,SourceHost)
-      | extend DestinationAddress = extract(@'([\.0-9]+)(:[\.0-9]+)?',1,DestinationHost)
-      | where not(ipv4_is_private(DestinationAddress))
+      | where OperationName in ("AzureFirewallApplicationRuleLog", "AzureFirewallNetworkRuleLog")
+      | parse kind=regex flags=U msg_s with Protocol 'request from ' SourceHost 'to ' DestinationHost @'\.? Action: ' Action @'\.' Rest_msg
+      | extend SourceAddress = extract(@'([\.0-9]+)(:[\.0-9]+)?', 1, SourceHost)
+      | extend DestinationAddress = extract(@'([\.0-9]+)(:[\.0-9]+)?', 1, DestinationHost)
+      | extend RemoteIP = case(not(ipv4_is_private(DestinationAddress)), DestinationAddress, not(ipv4_is_private(SourceAddress)), SourceAddress, "")
+      // Traffic that involves a public address, and in case this is the source address then the traffic was not denied
+      | where isnotempty(RemoteIP) and not(RemoteIP == SourceAddress and Action == "Deny")
```
shainw: I am good with the other changes, but we do not want to exclude Deny's in the default query as some users want this included so they can see attempts.  I will remove that piece and approve.

shainw (approved):
Thanks for catching the ungreedy issue.  Many of these have been around for quite some time and good to have a fresh look at them especially when underlying data schema/fields change. :)

PR #3718, opened 2021-12-15
Differentiate Action column from Indicators and from Azure Firewall in IPEntity_AzureFirewall.yaml

ep3p:
Fixes #

Sorry for not catching up this yesterday.

Some Indicators may have an "Action" column that collides with the Action column I specified for the Azure Firewall action.


aprakash13 (approved):
The changes look good.  Thanks @ep3p. Indeed, some TI feeds may have an "Action" column populated. Approved.

PR #3724, opened 2021-12-15
Checking in a hunting and a detection query for Log4j Vulnerability t…

aprakash13:
Checking in a hunting and a detection query for Log4j Vulnerability that looks at user agent and network connections to find malicious activity.

### DIFF
--- a/Detections/MultipleDataSources/UserAgentSearch_log4j.yaml
+++ b/Detections/MultipleDataSources/UserAgentSearch_log4j.yaml
@@ -1,0 +1,104 @@
+id: 29283b22-a1c0-4d16-b0a9-3460b655a46a
+name: User agent search for log4j exploitation attempt
+description: |
+  'This query uses various log sources having user agent data to look for log4j CVE-2021-44228 exploitation attempt based on user agent pattern. Log4j is an open-source Apache logging library that is used in 
+   many Java-based applications. The regex and the string matching look for the most common attacks. This might not be comprehensive to detect every possible user agent variation.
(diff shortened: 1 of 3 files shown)


Detections/MultipleDataSources/UserAgentSearch_log4j.yaml
```diff
@@ -0,0 +1,104 @@
...
+    - CVE2021-44228
+    - Schema: ASimWebSession
+    - SchemaVersion: 0.2.1
+    - Schema: ASimNetworkSessions
+    - SchemaVersion: 0.2.1
+query: |
+  let UserAgentString = dynamic (["${jndi:ldap:/", "${jndi:rmi:/", "${jndi:ldaps:/", "${jndi:dns:/", "${jndi:iiop:/","${jndi:","${jndi:nds:/","${jndi:corba/"]);
+  (union isfuzzy=true
+  (OfficeActivity
+  | where UserAgent has_any (UserAgentString) or UserAgent matches regex '\\$\\{j\\$\\{::-\\}n\\$\\{::-\\}d\\$\\{::-\\}[a-zA-Z]'
```
shainw, line 47: Move the regex out of each query into a let statement.

shainw (changes requested):
Just one change and then good.

Detections/MultipleDataSources/UserAgentSearch_log4j.yaml
aprakash13: Thanks, Shain.  Made the changes as suggested.

shainw (approved):

PR #3743, opened 2021-12-17
Update IPs in Log4j IOC list

shainw:
Fixes #

Updating latest IPs for Log4j IOC list


aprakash13 (approved):
Addition of new IP addresses to the IOC list. Approved.

PR #3747, opened 2021-12-17
New hunting query looking for external LDAP connections

petebryan:
# Proposed Changes
PR Adds a new hunting query that looks for connections to an external LDAP server where that server has not been connected to previously.

Query is for common network security log types and equivalent for ASIM datasets is underway but need to resolve some parsing issues first.


aprakash13 (approved):
Query looks good to be approved.

PR #3758, opened 2021-12-20
Adding imNetwork to IP Activity

yaronMSFT:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/Insights/IP/IP_SessionCounts.yaml
+++ b/Insights/IP/IP_SessionCounts.yaml
@@ -93,6 +93,12 @@ BaseQuery: |
   //| where not(DeviceAction has_any ('built','deny', 'denied', 'rst', 'blocked', 'teardown'))
   | extend Direction = iff(CommunicationDirection !in~ ('outbound','0') or CommunicationDirection !in~ ('inbound','1'), 'NotAvailable', CommunicationDirection)
   | summarize StartTime = min(TimeGenerated), EndTime = max(TimeGenerated), InboundCount = countif(Direction in~ ('Inbound','1')), OutboundCount = countif(Direction in~ ('Outbound', '0')), UnknownDirection = countif(Direction =~ 'NotAvailable') by IPAddress = DestinationIP, Type = strcat(Type,':', DeviceVendor,'-', DeviceProduct), RemoteIPAddress = SourceIP, Direction, SentBytes = tolong(SentBytes), ReceivedBytes = tolong(ReceivedBytes)
+  ),
+  (
+  imNetworkSession(starttime=start_time, endtime=end_time, dstipaddr=Ip_Address) // , dvcaction='Allow')
+  | where SrcIpAddr != DstIpAddr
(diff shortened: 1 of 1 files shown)


Insights/IP/IP_SessionCounts.yaml
```diff
@@ -93,6 +93,12 @@ BaseQuery: |
   //| where not(DeviceAction has_any ('built','deny', 'denied', 'rst', 'blocked', 'teardown'))
   | extend Direction = iff(CommunicationDirection !in~ ('outbound','0') or CommunicationDirection !in~ ('inbound','1'), 'NotAvailable', CommunicationDirection)
   | summarize StartTime = min(TimeGenerated), EndTime = max(TimeGenerated), InboundCount = countif(Direction in~ ('Inbound','1')), OutboundCount = countif(Direction in~ ('Outbound', '0')), UnknownDirection = countif(Direction =~ 'NotAvailable') by IPAddress = DestinationIP, Type = strcat(Type,':', DeviceVendor,'-', DeviceProduct), RemoteIPAddress = SourceIP, Direction, SentBytes = tolong(SentBytes), ReceivedBytes = tolong(ReceivedBytes)
+  ),
+  (
+  imNetworkSession(starttime=start_time, endtime=end_time, dstipaddr=Ip_Address) // , dvcaction='Allow')
+  | where SrcIpAddr != DstIpAddr
```
oshezaf, line 99: How do you think unknown direction would work?
How do we handle duplicates?

oshezaf:
@YaronFruchtmann : note that I commented on the PR.

Insights/IP/IP_SessionCounts.yaml
shainw: Direction: I think unknown direction needs to be used on conjunction with the source data type.  In other words, not all data types provide direction or it is implied based on the data type.  I think we make that clear in the result to help the user.  Thoughts?

Duplicates: Seems like duplicates is a loaded term here.  For example, duplicates of an IP between a variety of logs is expected, but I don't think we should obfuscate that.  I think we could do a makeset() or pack() or the like on the Type column to aggregate but still show that this IP showed up in x,y,z data types.  Thoughts?

shainw:
@YaronFruchtmann and @oshezaf - I also commented. ;)

oshezaf:
Comments from Yaron and my conversation: 
- First, not a high priority.
- Will wait to inNetwork LA solution functions deployment

Duplication was meant in the context of ASIM and direct source use.

shainw:
@YaronFruchtmann and @oshezaf - do we want to revisit this one now that ASIM is released?

v-mchatla:
Hi @oshezaf
Can we close this and reopen whenever required as you have labelled it as differed.
Thanks

oshezaf:
As I wrote, those are intentionally kept open as we intend to get back to them soon.

PR #3764, opened 2021-12-20
Removing some dupes and adding a couple.

shainw:
Fixes #
Update to fix some dupes and a couple new IPs


aprakash13 (approved):
Approving the updated IP List

PR #3773, opened 2021-12-22
Update link and condition in AzurePortalSigninfromanotherAzureTenant.yaml

ep3p:
Updated link, as noted by @timpailthorpe.


ep3p:
Also, I think this rule does not currently work. Now there is a column called AADTenantId, that may be equal to ResourceTenantId, but not TenantId.

This rule was created by @petebryan , please, could you review the different TenantId columns?

shainw:
@aprakash13 - the latest range list is always here - https://www.microsoft.com/download/details.aspx?id=56519.  We should look at a method of auto-updating the list.

aprakash13 (approved):
Updating the new Azure ranges json file location.

PR #3777, opened 2021-12-22
add missing id to rule

Phrozyn:
Fixes #

Adds an ID to the following rule for consistency:

- Hunting Queries/SigninLogs/SuccessfulAccount-SigninAttemptsByIPviaDisabledAccounts.yaml


shainw (approved):
Odd, not sure how that got thru the scrub we did a long time back, thanks for catching. ;)

PR #3785, opened 2021-12-23
analytic content for Google Workspace

spsocprime:
## Proposed Changes

  - add analytic rules
  - add hunting queries
  - add workbook


vmanojreddy:
@aprakash13 ,please review Analytic Rules , thanks.

Solutions/GoogleWorkspaceReports/Analytic Rules/GWorkspacePossibleMaldocFileNamesInGDRIVE.yaml
```diff
@@ -0,0 +1,33 @@
...
+triggerOperator: gt
+triggerThreshold: 0
+tactics:
+  - InitialAccess
+relevantTechniques:
+  - T1566
+query: |
+  GWorkspaceActivityReports
+  | where isnotempty(DocTitle)
+  | where DocTitle contains "invoice" or DocTitle contains "payment" or DocTitle contains "order" or DocTitle contains "fax" or DocTitle contains "scan" or DocTitle contains "transfer" or DocTitle contains "report" or DocTitle contains "bill"
```
aprakash13, line 21: Instead of numerous 'contains' we can probably use the 'has_any' operator here. 
Eaxample use of 'has_any' in a query: https://github.com/Azure/Azure-Sentinel/blob/master/Hunting%20Queries/ASimProcess/imProcess_powershell_downloads.yaml

Solutions/GoogleWorkspaceReports/Hunting Queries/GWorkspaceSuspendUser.yaml
```diff
@@ -0,0 +1,23 @@
+id: 26cea052-5da9-11ec-bf63-0242ac130002
+name: GWorkspace - Suspend users
+description: |
+  'Query searches suspend users.'
```
aprakash13, line 4: Wondering if it should be suspend vs suspended users.

aprakash13 (changes requested):
The queries overall look good.

One small suggestion - It seems a lot of analytic queries is using 'contains' operator that could probably be replaced by 'has'. 'Has' is more performant than 'contains' operator, since it doesn't look for substrings. Can you please take a look if these can be changed?

A few other minor changes have also been suggested.

Solutions/GoogleWorkspaceReports/Analytic Rules/GWorkspacePossibleMaldocFileNamesInGDRIVE.yaml
spsocprime: These are not full terms (e.g. file names are like invoice2021-12-30.ext or CompanyXXXpayment.doc, etc.), that's why we use `contains` in query instead on `has_any` to cover more cases

Solutions/GoogleWorkspaceReports/Hunting Queries/GWorkspaceSuspendUser.yaml
spsocprime: fixed name/description text

aprakash13:
> The queries overall look good.
> 
> One small suggestion - It seems a lot of analytic queries is using 'contains' operator that could probably be replaced by 'has'. 'Has' is more performant than 'contains' operator, since it doesn't look for substrings. Can you please take a look if these can be changed?
> 
> A few other minor changes have also been suggested.

@spsocprime  Team.... Thanks for getting some of the  other suggestions addressed. Can you please look into the use of 'has' vs 'contains' and confirm that these are needed in the analytic queries?  Once this is done, we can go ahead and merge these.

NikTripathi (approved):
Workbook looks good.

shainw (commented):
For Detections, I see 'contains' used a bunch, we should use 'has' unless it won't work.  Please be sure to test this.

Additionally, it seems that many detections only use 1 condition which means we could have 1 query for many of these that simply uses customer alert name and description based on the field that the condition is matching on - https://github.com/Azure/Azure-Sentinel/wiki/Query-Style-Guide#alertdetailsoverride

v-rucdu:
@spsocprime : Can you please resolve the comments provided by Shain.. Thanks!!!

spsocprime:
> @spsocprime : Can you please resolve the comments provided by Shain.. Thanks!!!

fixed where it was reasonable

PR #3796, opened 2021-12-27
Perf improvements to Log4J_IPIOC_Dec112021.yaml

shainw:
Fixes #

- Take just IPs from each source, match against ioc IPList and then only filter on matches
- Include Type


aprakash13 (approved):
Looks great. Approved.

PR #3807, opened 2021-12-29
Analytic content for Jira

spsocprime:
## Proposed Changes

  - analytic rules
  - hunting queries
  - workbook


v-jayakal:
@aprakash13  - Please review analytics.

aprakash13 (commented):
Approved from a detection perspective.  Please feel free to approve/merge if other things look good.

One feedback for future is that there is an opportunity here to simplify the number of distinct queries by using custom alert names that looks for all of the different matches and labels the Alert name and Description based on the value that is triggered. See - https://github.com/Azure/Azure-Sentinel/wiki/Query-Style-Guide#alertdetailsoverride

PR #3870, opened 2022-01-10
Preparations for SF 

yaronMSFT:
* Adding EquivalentBuiltInParser tag
* Adding Parsers tag to meta parsers
* Updating Yaml2ARM script to support parameters


ASIM/dev/ASimYaml2ARM/ASimYaml2ARM.py
```diff
@@ -23,12 +23,19 @@
...
+params=parserYaml.get('ParserParams')
 
 data_section=arm_template['resources'][0]['resources'][0]
 data_section['name'] = alias
 data_section['properties']['query'] = query
 data_section['properties']['FunctionAlias'] = alias
 data_section['properties']['displayName'] = title
+for param in params:
+    if param['Type']=='string':
+        param['Default'] = f"\'{param['Default']}\'"
```
oshezaf, line 35: We don't need to modify that.

oshezaf (changes requested):

PR #3903, opened 2022-01-12
Fix join failnoSuccess for Detection

PawelLiebich:
Change(s):
   - Updated syntax for FailedLogonToAzurePortal.yaml

   Reason for Change(s):
   - Not working detection

**Testing Completed:**
  Yes/ No :


shainw (dismissed):
Agreed, this is causing misses since the Id can be different and is not related to UPN.  We also don't use ResultType or Id in the eventual result, so I updated the PR to remove those also.

shainw (approved):

PR #3910, opened 2022-01-13
Merge master to insights_preview

rpressburger:
Merge master to insights_preview


shainw:
Both of the failures are okay.  WorkbooksMetadata.json has the correct version.  AdvancedKQL.json has URL paths with locales in them.  79 instances of en-us, which need to be removed, but this will be done in the main branch and insights_preview will get it on the next merge back.

shainw (approved):

PR #3916, opened 2022-01-13
Hunting Query Automation Fix

elforb:
Change:
Updating the automation to find the index of the workspace object, if available, and utilize that workspace object.

Reasoning:
In the automation, if there are both Parsers and Hunting Queries, if the Parser resources aren't declared directly before the Hunting Queries in the input file then duplicate workspace resources will be created.


Tichandr:
@elforb - Changes looks good to me. Can you confirm that every time there will be only single resource type "Microsoft.OperationalInsights/workspaces" created in the MainTemplate.json file for both parsers, hunting queries etc.

With this change we must ensure that Parsers should be followed by Hunting Queries in the input step. If not then the additional resource type "Microsoft.OperationalInsights/workspaces" for parser will get created. Why don't we check the resource type existing condition check in the Parsers logic as well? Please confirm. Thanks

elforb:
@Tichandr Good catch, I've updated the parser section as such.

v-rucdu (approved):

PR #3921, opened 2022-01-14
Corrects EntityMapping in ASim detection rule for Solarwinds

Korving-F:
Detection rule `SUNBURST and SUPERNOVA backdoor hashes (Normalized File Events)` uses non-normalized field (MD5) in entity mappings breaking rule deployment.
   
   Change(s):
   - This PR makes the rule use the correct field as used in the imFileEvent parser (TargetFileMD5).

   Reason for Change(s):
   - Incorrect entity field breaks deployment

   Testing Completed:
   - Yes

### DIFF
--- a/Detections/ASimFileEvent/imFileESolarWindsSunburstSupernova.yaml
+++ b/Detections/ASimFileEvent/imFileESolarWindsSunburstSupernova.yaml
@@ -29,7 +29,8 @@ query:  |
       timestamp = TimeGenerated,
       AccountCustomEntity = User, 
       HostCustomEntity = DvcHostname,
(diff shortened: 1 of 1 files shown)


Korving-F:
In hindsight this is not correct. Algorithm is an enum and the MD5 field needs to be replaced with an "MD5"-string instead.
See also:
- https://github.com/Azure/Azure-Sentinel/commit/8768b916756b827da02d1dfd95ece8fbe27049c4
- https://docs.microsoft.com/en-us/azure/sentinel/entities-reference#file-hash

oshezaf (approved):

PR #3922, opened 2022-01-14
Corrects imAuthBruteForce scheduled alert rule's EntityMapping

Korving-F:
Detection rule `Brute force attack against user credentials (Uses Authentication Normalization)` uses non-normalized fields (AccountCustomEntity / IPCustomEntity) in entity mappings breaking rule deployment.
   
   Change(s):
   - This PR makes the rule use the correct fields as defined in the imAuthentication parser / rule's own field mappings.

   Reason for Change(s):
   - Incorrect entity fields breaks deployment

   Testing Completed:
   - Yes

### DIFF
--- a/Detections/ASimAuthentication/imAuthBruteForce.yaml
+++ b/Detections/ASimAuthentication/imAuthBruteForce.yaml
@@ -41,6 +41,6 @@ entityMappings:
   - entityType: IP
     fieldMappings:
       - identifier: Address
-        columnName: IPAddress
+        columnName: SrcDvcIpAddr
 version: 1.1.1
 kind: Scheduled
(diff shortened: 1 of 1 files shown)


oshezaf (approved):

Detections/ASimAuthentication/imAuthBruteForce.yaml
```diff
@@ -37,10 +37,10 @@ entityMappings:
...
-        columnName: AccountCustomEntity
+        columnName: TargetUsername
   - entityType: IP
     fieldMappings:
       - identifier: Address
-        columnName: IPCustomEntity
-version: 1.1.0
-kind: Scheduled
\ No newline at end of file
+        columnName: IPAddress
```
oshezaf, line 44: Should be SrcDvcIpAddr

oshezaf (changes requested):
I was too fast to approve. Requires a correction.

And thanks for finding this bug!

Detections/ASimAuthentication/imAuthBruteForce.yaml
Korving-F: Column should now have the correct value.

oshezaf (approved):

Issue #3256, opened 2021-10-20
Duo Security Data connector


AlexeyMihalich:
**Describe the bug**
This is not exactly a bug, but more of a compatibility issue.
Duo security data connector uses Get-Date -AsUTC in his powershell script. AsUTC was introduced in powershell 7.1, while azure functions currently operate in powershell 7.0. So the function simply wont work, I've got a workaround which i will post here (in the additional context). 

**To Reproduce**
1. Go to https://aka.ms/sentinelduosecurityazuredeploy ,Its the template suggested in: https://github.com/Azure/Azure-Sentinel/blob/master/DataConnectors/Duo%20Security/DuoSecurity_API_FunctionApp.json
3. Deploy as described in template.
4. go to functions, click your function and open the run.ps1 file in online editor
5. click test/run . then click run on the right side of the window.
4. See error in error log after running the function (you will get it prompted after clicking on run).
5. **Optional** -you might see another error which says %Schedule% isnt defined or doesnt have a value, in this case upload the files manually and modify %schedule% to the cron expression format - 0 */5 * * * *. To edit the files you need to delete the environment variable which makes the function readonly as he pulls the zip directly from github.

**Expected behavior**
The function wont run and wont perform its function.

**Screenshots**
Don't have any.

**Desktop (please complete the following information):**
 - Not Relevant

**Smartphone (please complete the following information):**
 - N/A

**Additional context**
suggested solution (may require optimization): 
  $New = Get-Date
  $New = (Get-Date).AddHours(-$hrs)
  $New2 = ($New|Get-Date -UFormat %s).ToString() +"000"
  $New = $New.AddMinutes(-$time)
  $New = $New | Get-Date -UFormat %s
  $New3 = $New.ToString()
  $New = $New.ToString() +"000"
  $minTime = $New
  $maxTime = $New2
  $minTimeSeconds = $New3
I've got it working with the mentioned modifications.

v-rucdu:
Hey.. we are looking into this issue and would get back to you asap... Thanks!!!

v-rucdu:
Hey @AlexeyMihalich, we have raised a [PR](https://github.com/Azure/Azure-Sentinel/pull/3923) with the fix. Can you please check and confirm if it works for you?.. Thanks!!!

v-rucdu:
@AlexeyMihalich, can you please confirm if the fix mentioned in the PR works for you, so that we can merge the changes and close this issue... thanks!!!

[...]

PR #3923, opened 2022-01-14
Duo Security Function Compatibility Fix

elforb:
Changes:
Replacing use of `-AsUTC` flag with `ToUniversalTime()` method.
Reasoning:
In the PowerShell run script of the Cisco Duo Security function app, there existed a compatibility issue due to use of a PowerShell 7.1 functionality, the `-AsUTC` flag, while function apps are currently utilizing PowerShell 7.0 where that does not exist.
This fix is made due to the following issue: #3256


v-rucdu:
@elforb : Please update the zip file as well with this change... Thanks!!

v-rucdu:
AlexeyMihalich confirmed on the issue that the fix would work, hence approving and merging the PR... Thanks!!

v-rucdu (approved):

PR #3932, opened 2022-01-17
Asim/dns/add zscaler to union

oshezaf:


YaronFruchtmann (approved):
Approving. 
Has no functional significance but maybe rename files too

PR #3934, opened 2022-01-17
Fix some Domain Entities not being checked in DomainEntity_SecurityAlert.yaml

ep3p:
Change(s):
   1. Check all possible domains in Entities, instead of just one per SecurityAlert, I have lost alerts because of this.
   2. Take just one hostname and one address from each alert, instead of expanding their entities (useless with innerunique kind).
   3. Extend the domain name regex to accept punycode.
   4. Add the domain name and entities in the projected columns.

   Reason for Change(s):
   1. An alert can have multiple Url Entities. Some analytic rules integrate all events in one alert, or create an alert for each event.
   2. The join kind is innerunique, so which Hostname or Address end up in the results cannot be controlled.
   3. There are punycode top level domains, which were not being considered in the query.
   4. Without them you don't know what are you looking for.

   Testing Completed: Yes but Need Help

  I don't have a good dataset to check this query against, please, could you check it?

  I have a problem with ```extract_all```. If I don't ```todynamic(dynamic_to_json(``` the results, expanding them sometimes returns a dynamic and other times a string.
_________________________________
**The same thing happens with https://github.com/Azure/Azure-Sentinel/blob/master/Detections/ThreatIntelligenceIndicator/URLEntity_SecurityAlerts.yaml**

Does not happen with **https://github.com/Azure/Azure-Sentinel/blob/master/Detections/ThreatIntelligenceIndicator/EmailEntity_SecurityAlert.yaml**

I think these 3 queries should try to be more common or be normalized to a more similar pattern.

Also, there could be a "IPEntity_SecurityAlert.yaml"

### DIFF
--- a/Detections/ThreatIntelligenceIndicator/DomainEntity_SecurityAlert.yaml
+++ b/Detections/ThreatIntelligenceIndicator/DomainEntity_SecurityAlert.yaml
@@ -47,25 +47,27 @@ query: |
         | extend MSTI = case(AlertName has "TI map" and VendorName == "Microsoft" and ProductName == 'Azure Sentinel', true, false)
         | where MSTI == false
         //Extract domain patterns from message
-        | extend domain = extract("(([a-z0-9]+(-[a-z0-9]+)*\\.)+[a-z]{2,})", 1, tolower(Entities))
-        | where isnotempty(domain)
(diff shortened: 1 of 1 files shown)


v-sabiraj:
Hey @aprakash13 , will you please review and approve the changes, thanks.

v-sabiraj:
Hey @aprakash13 , will you please review and approve the changes, thanks.

vmanojreddy:
Hey @aprakash13 , will you please review and approve the changes, thanks.

vmanojreddy:
Hey @aprakash13 , will you please review and approve the changes, thanks.

petebryan (approved):

PR #3948, opened 2022-01-19
Replacing imDns with _Im_Dns, version update

yaronMSFT:


Detections/MultipleDataSources/Log4J_IPIOC_Dec112021.yaml
```diff
@@ -70,8 +70,8 @@ query:  |
   let CSlogDestIP = CommonSecurityLog | summarize by IPAddress = DestinationIP, Type;
   let CSlogMsgIP = CommonSecurityLog | extend MessageIP = extract(IPRegex, 0, Message) | summarize by IPAddress = MessageIP, Type;
   let DnsIP = DnsEvents | summarize by IPAddress = IPAddresses, Type;
-  // If you have enabled the imDNS and/or imNetworkSession normalization in your workspace, you can uncomment one or both below.  Reference - https://docs.microsoft.com/azure/sentinel/normalization
-  //let imDnsIP = imDns (response_has_any_prefix=IPList) | summarize by IPAddress = ResponseName, Type;
+  // If you have enabled the _Im_Dns and/or imNetworkSession normalization in your workspace, you can uncomment one or both below.  Reference - https://docs.microsoft.com/azure/sentinel/normalization
```
oshezaf, line 73: _Im_Dns is already enabled..... you can make this active by default here.

oshezaf (changes requested):
Two more:
- Did you test at least one to check it works, for example when defining the rule in the UI?
- Find the rules that use DnsEvents but not ImDns and for those due to length and for those I do suggest doing the replace. It adds values.

oshezaf (dismissed):

oshezaf (approved):

YaronFruchtmann:
Fails due to _Im_Dns not known to tests

PR #3964, opened 2022-01-20
Add Microsoft 365 Defender queries

orshe4:
Add Microsoft 365 Defender queries


aprakash13 (approved):
Thanks for all the work on this one @orshe4 .

(Note: There is some parallelism issue going on with the Microsoft Sentinel Github that is giving weird issues. However, all the validation checks have passed, and this is good to be approved. )

PR #3978, opened 2022-01-21
Fix conditions in UnusualGuestActivity.yaml

ep3p:
Change(s):
1. Take the correct part (```[0]```) of #EXT# users in AuditLogs.
2. Replace "@" by "_" to make correct comparisons.
3. Add ```AADNonInteractiveUserSignInLogs``` to make comparisons.
4. Make the join comparing the InvitedUser, NOT the user that made the invitation (InitiatedByUser).
5. Comment the line where the user who made the invitation has to be a Guest.
6. Remove Office365 Shell WCSS-Client app.
~~7. Change OperationName conditions.~~
8. Specify queryfrequency to 1h.
__________________________________________________
   Reason for Change(s):
1. In my tenant these Guest users have the following format ```external.user_external.domain#EXT#@my.domain```
2. #EXT# users have their "@" character automatically replaced by "_", I prefer not to bring "@" back to make comparisons.
3. An attacker does not need to signin interactively in a Powershell app as their first step.
4. The description link talks about a guest user who is invited AND **this invited user** signins with a Powershell app, NOT that a guest user invites a user AND **the user who made the invitation** signins with a Powershell app.
5. I think the interesting part is the invited user signins with a Powershell app, not the nature of the inviting user. The reference link does not talk about this scenario.
6. I think this app is logged by usual web browser activities.
~~7. I have observed in my tenant some bulk OperationNames do not start by the "Bulk" string, even if Microsoft documentation says so.~~
8. I guess you don't want to wait a day to know this activity.
   Testing Completed:
   - Yes but RECHECK

Please, if you only want to detect **if the invitation was done by a Guest user**, uncomment line ~~31~~ 34.


ep3p:
I see now that "Bulk invitation" does not have any TargetResources (it expands to "Invite external user" operations). So I "restored" the detection to check if the initiator user is a guest and signins with a PowerShell App.

Now this dectection would check both invitator guest users and invited guest users using PowerShell Apps.

I have also added the app Pnp Management Shell.

ep3p:
I did not realize I have opened another Pull Request (https://github.com/Azure/Azure-Sentinel/pull/4312) for the app PnP Management Shell. I will take out that part in this PR. Sorry.

aprakash13 (dismissed):
Thanks @ep3p

aprakash13 (approved):
Thanks @ep3p

PR #3980, opened 2022-01-21
Solution Automation: Workspace Filter Update

elforb:
Change(s):
   - Updated handling of casing for workspace dropdown

   Reason for Change(s):
   - An update to the Azure API has caused resource IDs to no longer be guaranteed to be lowercase in certain workspace locations (Ex. France Central).


rushriva (approved):
PR changes approved

PR #3981, opened 2022-01-22
Use of param domain_has_any

yaronMSFT:
Change(s):
   - Modified the parameter passed to dynamic parameter domain_has_any

   Reason for Change(s):
   - Syntax error

> The code should have been tested in a Microsoft Sentinel environment that does not have any custom parsers, functions or tables, so that you validate no incorrect syntax and execution functions properly.  If your submission requires a custom parser or function, it must be submitted with the PR.

   Testing Completed:
   - Yes/No/Need Help


oshezaf (approved):

oshezaf (changes requested):
I think you need to update the version number. While probably not critical as the previous version was probably not deployed yet, it fails verification. And if you are unlucky it was deployed....

PR #4003, opened 2022-01-25
reverting _Im_dns

yaronMSFT:
reverting #3948, until next version builtin dns is released


Detections/MultipleDataSources/ExchangeServerVulnerabilitiesMarch2021IoCs.yaml
```diff
@@ -144,7 +144,7 @@ query: |
   | where Name has_any (domains)
   | extend timestamp = TimeGenerated, IPCustomEntity = ClientIP, HostCustomEntity = Computer
   ),
-  (
```
oshezaf, line 147: You got it wrong here.

PR #4067, opened 2022-02-01
Solution Automation: Template Specs

elforb:
Change(s):
- Updated solution automation to handle Template Specs, in accordance with the internal template spec experiment.

Reason for Change(s):
- Template Spec solutions will be necessary during the upcoming solution migration push.

> Note: This is based on the initial version of the template spec experiment, and will be finalized further based on any updates.


v-sabiraj:
Hey @elforb, will you check and close this if not needed, thanks

PR #4081, opened 2022-02-02
Update CODEOWNERS

shainw:
Change(s):
   - Adding Preeti to proper folders

Reason for Change(s):
   - Due to how codeowners works, global code owners must be part of defined folders to show up as reviewer in PR

Testing Completed:
   - Yes

   - Yes/No/Need Help


aprakash13 (approved):
Some issue with parallelism that is causing the validations to act weird and failing. However, this should be good to be approved.

PR #4130, opened 2022-02-08
asim/convert-network-sizes-to-long

oshezaf:
   
   Change(s):
   - Update network parsers to use long for traffic sizes fields

   Reason for Change(s):
   - Updated network schema

   Version Updated:
   - N/A

   Testing Completed:
   - N/A

   - Post PR creation


yaronMSFT (approved):
Maybe we need a safety net on the metaparser, in case anyone uses an older version of vendor specific parser

PR #4151, opened 2022-02-11
Update FirstAppOrServicePrincipalCredential.yaml

KustoKing:
new_value_set contains multiple values, therefore i used mv-expand. 
bumped version to 1.1.1

   Change(s):
   - Splitted new_value_set with mv-expand

   Reason for Change(s):
   - new_value_set had 2 records KeyUsage: Sign and Verify, the parse only returned Sign and the query did not gave the expected result

   Version Updated:
   - Yes (1.1.1)

   Testing Completed:
   - Yes

   - Yes


aprakash13 (approved):
Thanks @KustoKing

PR #4156, opened 2022-02-13
Asim/fix filter for panos and empty updates

oshezaf:
Change(s):
   - Update Palo Alto PanOS parser to require EventProduct to be "PAN-OS"
   - Add fields to empty WebSession and NetworkSession tables.


yaronMSFT (approved):
Do we change the version of meta-parser when any of its element changes, or only when there's a direct change to meta-parser?
Remember to update arm before merging

oshezaf:
> Do we change the version of meta-parser when any of its element changes, or only when there's a direct change to meta-parser? 

I don't think so.

> Remember to update arm before merging

Remember that I prefer to stage PR, balancing consistency and prudence.

PR #4164, opened 2022-02-14
CrowdstrikeReplicator:fixes

averbn:
Change(s):
- now the launch will be regulated by the internal trigger of the function itself, and not by timeouts in the code. This will allow us to easily and independently scale - we can simply add a new similar function if necessary(or several functions ), while they will work and gather messages independently of each other.
- made the categorizing of the message into required and optional fields(to decrease the count of using fields in the table).

   Reason for Change(s):
   - Performance and ingestion issues.


anki-narravula (dismissed):
Looks good. Approved.

PR #4179, opened 2022-02-15
Update table name in DomainEntity_DnsEvents.yaml

ep3p:
Change(s):
   - Employ correct name of the Table being compared.

   Reason for Change(s):
   - Coherence with other TI map rules.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes
   
 I think Asim parsers instead of DnsEvents could be used in the next versions of TI map DNS rules.


aprakash13 (approved):
Thanks @ep3p.

Issue #4168, opened 2022-02-14
Sentinel Lab - Module 5 - No HighRiskApps Watchlist 


lemonsprig:
**Describe the bug**
There is no HighRiskApps watchlist. Causing step 13 and 14 to fail. 

**To Reproduce**
1. Follow the module instructions to step
2. At step 13 run the query _GetWatchList('HighRiskApps')

**Expected behavior**
The query should execute and return a list of results from the watchlist. 

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 98.0.4758.82 (Official Build) (64-bit)

**Additional context**
Looks like the creation of the watchlist is a missing step as part of the solution installation or Module 1 environement setup.

Lodewyk-Git:
Hey @lemonsprig, looks like the solution skips the deployment of the watchlist, you can deploy it manually using [the following as a template](https://github.com/Azure/Azure-Sentinel/blob/master/Solutions/Training/Azure-Sentinel-Training-Lab/Artifacts/LinkedTemplates/watchlist.json), remember to select the correct sentinel workspace and resource group:

Steps:
1. Search deploy a custom template and select it, you should see:
![image](https://user-images.githubusercontent.com/36856692/154067709-1d3ebe6f-c019-4032-bb51-8c92fa20ea6a.png)
2. Click "build your own template in the editor" - replace everything with the json contents from the link like so: 

![image](https://user-images.githubusercontent.com/36856692/154067915-f5f09097-9a4c-4de5-a060-baaa68a53246.png)

Click save and select the correct resource group and workspace. The Watchlist will then be in the sentinel workspace after the template has been deployed.

lemonsprig:
Thank you @Lodewyk-Git that has done the trick.

PR #4181, opened 2022-02-15
Update mainTemplate.json

Lodewyk-Git:
Added the watchlist, I've not tested this 

Also:
How do I test this? 

The Azure deploy file looks like it should work, this is the only file that doesn't mention the watchlist and everything else in this file did get deployed from the solution in sentinel


Lodewyk-Git:
#4168

v-sabiraj:
Hey @Lodewyk-Git , This branch has conflicts that must be resolved, thanks.

PR #4210, opened 2022-02-17
Detection to explore users reading the local device identity key

Cyb3rWard0g:
This detection uses Windows security events to look for users opening/reading the local Device Identity Key (Machine Key).
 This information can be correlated with other events for additional context and get to use-cases where a
 machine key with a transport key together can be used to impersonate an AAD joined or registered machine.

 Reference: https://o365blog.com/post/deviceidentity/

image


aprakash13 (approved):
Thanks @Cyb3rWard0g .

PR #4211, opened 2022-02-17
asim/nsg-network-parser

oshezaf:
   
   Change(s):
   - ASIM Network parser for NSG

### DIFF
--- a/Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionAzureNSG.yaml
+++ b/Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionAzureNSG.yaml
@@ -1,0 +1,138 @@
+Parser:
+  Title: Network Session ASIM parser for Azure NSG logs
+  Version: '0.1'
+  LastUpdated: Feb 16, 2021
+Product:
+  Name: Azure NSG
+Normalization:
+  Schema: NetworkSession
+  Version: '0.2.2'
+References:
+  - Title: ASIM Network Session Schema
+    Link: https://aka.ms/ASimNetworkSessionDoc
+  - Title: ASIM
+    Link: https://aka.ms/AboutASIM
+Description: |
+  This ASIM parser supports normalizing Azure NSG logs to the ASIM Network Session normalized schema.
+ParserName: ASimNetworkSessionAzureNSG
+EquivalentBuiltInParser: _ASim_NetworkSession_AzureNSG
+ParserParams:
+  - Name: disabled
+    Type: bool
+    Default: false
+ParserQuery: |
+  let DvcActionLookup = datatable(FlowStatus_s:string, DvcAction:string, EventResult:string) [
+      'A', 'Allow', 'Success',
+      'D', 'Deny', 'Failure',
+  ];
+  let NetworkDirectionLookup = datatable(FlowDirection_s:string, NetworkDirection:string, isOutBound:bool) [
+      'I', 'Inbound', false,
+      'O', 'Outbound', true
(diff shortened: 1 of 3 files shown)


Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionAzureNSG.yaml
```diff
@@ -0,0 +1,138 @@
...
+  let NetworkProtocolLookup = datatable(L4Protocol_s:string, NetworkProtocol:string)[
+      'T', 'TCP',
+      'U', 'UDP'
+  ];
+  let parser = (disabled:bool=false) 
+  {
+    let AzureNetworkAnalytics = 
+      AzureNetworkAnalytics_CL
+      | where not(disabled) and isnotempty(FlowType_s)
+      | lookup NetworkDirectionLookup on FlowDirection_s
```
yaronMSFT, line 41: Th function is called twice, and lookup runs twice on both directions. If you first filter the  on flowdirection_s, and then lookup, it would perform better

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionAzureNSG.yaml
```diff
@@ -0,0 +1,138 @@
...
+          SrcDomain = tostring(hostelements[0]),
+          SrcDomainType = "ResourceGroup"
+      | extend Hostname = SrcHostname
+      | project-away hostelements
+    ;
+    union AzureNetworkAnalyticsInbound, AzureNetworkAnalyticsOutbound
+    | project-rename
+        Dvc = NSGList_s,
+        DvcSubscriptionId = Subscription_g,
+        EventEndTime = FlowEndTime_t,
```
yaronMSFT, line 89: For some schemas docs say EventEndTime is alias to Time Generated. For nw the csv treats it as an alias.
Parser is ok, doc to be updated

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionAzureNSG.yaml
```diff
@@ -0,0 +1,138 @@
...
+        NetworkApplicationProtocol = L7Protocol_s,
+        NetworkRuleName = NSGRule_s,
+        NetworkSessionId = ConnectionName_s,
+        EventOriginalSubType = FlowType_s
+    | extend
+        DstPortNumber = toint(DestPort_d),
+        EventProduct = 'NSGFlow',
+        EventSchema = 'NetworkSession',
+        EventSchemaVersion='0.2.2',
+        EventSeverity = 'Informational', //??
```
yaronMSFT, line 100: If Deny than low, no?

yaronMSFT (changes requested):
Lookup comment is the most important

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionAzureNSG.yaml
oshezaf: Jason insisted otherwise. I did not want to change now....

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionAzureNSG.yaml
oshezaf: Good. Point. Updated. I woder if given that I anyway test for direction here, I should move all the contact part to here rather than the end. However, did not want to make big changes just for style.

PR #4231, opened 2022-02-21
Create ChiaCryptoMining_WindowsEvent

yaelrbergman:
   
   Change(s):
Added ChiaCryptoMining detection for WindowsEvent 
   
Reason for Change(s):
Adding WindowsEvent as a source

   Version Updated:
Yes
   Testing Completed:
Yes

Yes


Detections/WindowsEvent/ChiaCryptoMining_WindowsEvent.yaml
```diff
@@ -0,0 +1,64 @@
+id: 4d173248-439b-4741-8b37-f63ad0c896ae
```
oshezaf, line 1: We use tags to refer to the original for code maintenance reasons. See [here](https://github.com/Azure/Azure-Sentinel/blob/master/Detections/ASimWebSession/PossibleDGAContacts.yaml)

Detections/WindowsEvent/ChiaCryptoMining_WindowsEvent.yaml
```diff
@@ -0,0 +1,64 @@
...
+triggerThreshold: 0 
+tactics: 
+  - Impact
+relevantTechniques:
+  - T1496
+tags:
+  - Chia 
+query:  |  
+  let iocs = externaldata(DateAdded:string,IoC:string,Type:string,TLP:string) [@"https://raw.githubusercontent.com/Azure/Azure-Sentinel/master/Sample%20Data/Feeds/ChiaCryptoIOC.csv"] with (format="csv", ignoreFirstRecord=True);
+  let process = (iocs | where Type =~ "process" | project IoC);
```
oshezaf, line 22: No need to keep the lists you don't use.

oshezaf (changes requested):

aprakash13 (commented):
@yaelrbergman Just wondering if we want a new query for this or do we want to include this in the existing ChiaCryptoMining detection.

oshezaf:
> @yaelrbergman Just wondering if we want a new query for this or do we want to include this in the existing ChiaCryptoMining detection.

@aprakash13 : as noted in another PR - Yael added a new version when the original was too long to add to.

oshezaf (approved):

PR #4233, opened 2022-02-21
Fix QueryFrequency in AnomalousIPUsageFollowedByTeamsAction.yaml

ep3p:
Change(s):
   1. Establish a queryfrequency parameter for OfficeActivity.
   2. Check activities from non-prevalent countries even if that activity is not the "lightest" activity.
   3. Remove other unneeded code.

   Reason for Change(s):
   1. If not, the same event will trigger an alert every day for 14 days.
   2. This line ```| where min_count_ == count_``` is leaving out rare activity that could possibly match OfficeActivity events.
   3. The query is a bit chaotic to understand, there are some oversteps, and it checks signinlogs even if there are not Office activities...

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes

If you want to keep it just as it is, just remove completely my commit and add:

```
let projectedEndTime = 60min;
let queryfrequency = 1d;
...
OfficeActivity
| where TimeGenerated > ago(queryfrequency)
```
So an event only triggers an alert once.

### DIFF
--- a/Detections/MultipleDataSources/AnomalousIPUsageFollowedByTeamsAction.yaml
+++ b/Detections/MultipleDataSources/AnomalousIPUsageFollowedByTeamsAction.yaml
@@ -56,12 +56,12 @@ query: |
           | extend TotalSignin = toscalar(signinData | summarize count())
           | extend CountryPrevalence = toreal(CountCountrySignin) / toreal(TotalSignin) * 100;
       // Count signins by user and IP address
-      let userIPSignin =
+      let userIpSignin =
           signinData
           | summarize CountIPSignin = count(), Country = any(Country), ListSigninTimeGenerated = make_list(TimeGenerated) by IPAddress, UserPrincipalName;
       // Calculate delta between the IP addresses with the most and minimum activity by user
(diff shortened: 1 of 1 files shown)


vmanojreddy:
@aprakash13 Please review detections.thanks.

ep3p:
I think ```tostring(todynamic(LocationDetails)['countryOrRegion'])``` could be substituted by ```Location```

vmanojreddy:
@aprakash13  Please review detections. thanks!!!

vmanojreddy:
@aprakash13 Please review detections. thanks!!!

aprakash13:
Reached out to @Shain for help with this.

thmcelro (commented):
Reviewed the changes, the original intention of the query is unaltered but with good improvements to performance. Addition of the query frequency and period make sense.

In testing this query performs well and produces useful results.

aprakash13 (approved):
Thanks @ep3p and @thmcelro

shain:
> 

@aprakash13 , did you mean to tag @shainw instead of @shain

PR #4235, opened 2022-02-22
Remove Duplicate Query in Filter for Known Domains Using Long DNS

ben12385:
Extra Name !has "wpad"

   Change(s):
   - Updated filter for DNS_LongURILookup.yaml
  
   Reason for Change(s):
   - Duplicate domain name with known long DNS lookup

   Version Updated:
   - No as no change in logic
 
   Testing Completed:
   - Need Help

   - Need Help


aprakash13 (approved):
Thanks @ben12385

PR #4247, opened 2022-02-23
Solution Automation: Enable SavedSearch Inputs

elforb:
Change(s):
   - Enabling SavedSearch inputs for Solution Packaging Tool

   Reason for Change(s):
   - Requested by ILDC team for internal use.

### DIFF
--- a/Tools/Create-Azure-Sentinel-Solution/createSolution.ps1
+++ b/Tools/Create-Azure-Sentinel-Solution/createSolution.ps1
@@ -718,12 +718,10 @@ foreach ($inputFile in $(Get-ChildItem $path)) {
                         $searchData = $json # Assume input is basic array of SavedSearches to start
                         # Check if SavedSearch input file uses direct structure given by export
                         if ($searchData -isnot [System.Array] -and $searchData.value) {
-                            Write-Host "Value prop exists"
                             $searchData = $searchData.value 
                         }
                         # Check if SavedSearch input file uses standard template structure
                         if ($searchData -isnot [System.Array] -and $searchData.resources) {
-                            Write-Host "Resources prop exists"
                             $isStandardTemplate = $true
                             $searchData = $searchData.resources
                         }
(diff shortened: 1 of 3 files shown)


Tools/Create-Azure-Sentinel-Solution/createSolution.ps1
```diff
@@ -712,6 +713,46 @@ foreach ($inputFile in $(Get-ChildItem $path)) {
                         # Update Connector Counter
                         $connectorCounter += 1
                     }
+                    elseif ($objectKeyLowercase -eq "savedsearches") {
+                        $isStandardTemplate = $false
+                        $searchData = $json # Assume input is basic array of SavedSearches to start
+                        # Check if SavedSearch input file uses direct structure given by export
+                        if ($searchData -isnot [System.Array] -and $searchData.value) {
```
Tichandr, line 720: @elforb - Can you give us sample input to understand more about the changes? because I can see in the changes that we are using searchdata as array sometimes and sometimes not. Please provide me the sample inputs to understand more about the changes related to savedsearches

elforb: The expected input format is either of the following:
- Raw array
- Direct format from export at https://docs.microsoft.com/en-us/rest/api/loganalytics/saved-searches/list-by-workspace
- Standard ARM Template

Tichandr: @elforb - Got you. In that case If input is of direct structure given by export template and it value prop exists then we don't need to add it to the the basemaintemplate file?

elforb: @Tichandr No?? We have to add the resources to the template no matter what format they come in.

elforb:
> $isStandardTemplate = $true is not required if value prop exists condition check. If you don't add have that check then how we are adding it to the basemaintemplate. Please check and let me know. Rest of all looks good to me. Thanks

@Tichandr I'm very confused on what you mean here due to odd phrasing. Could you please phrase better?
The 'isStandardTemplate' check helps us add the resources to the template since there are multiple formats in can come in.

PR #4383, opened 2022-03-10
Add detection for Password Spray against Seamless SSO

ep3p:
Change(s):
   - Add a specific detection for Password Spray against Seamless SSO.

   Reason for Change(s):
   - Theoretically this could be detected by Password Spray against Azure AD application, but the actor may change IPAddresses very quickly.

   Version Updated:
   - Does not apply

   Testing Completed:
   - Yes

   - Yes

Please, could you check if 81XXX ResultTypes are uncommon or not?

Please, could you check if there are more ResultTypes other than 81016 when an actor is trying unsuccessful passwords, or just checking which accounts might exist in a tenant?

image


aprakash13 (approved):
Thanks @ep3p

PR #4388, opened 2022-03-11
Update MFARejectedByUser to include AAD IDs

malexmave:
Change(s):
- Include `AadUserId` and `AadTenantId` in the `Account` entityMapping for the MFARejectedByUser rule.

Reason for Change(s):
- In my experience, Sentinel will not properly link the alerts / incidents created by the MFARejectedByUser rule to an AAD account unless extra `fieldMappings` for the `AadUserId` and `AadTenantId` are set. Changing the rule like this allowed the data to be linked correctly.
- I assume that most of the other rules in this folder could benefit from this change as well, but I have no way of testing them, so I did not modify them.

Version Updated:
- Yes

Testing Completed:
   - Yes

   - No (did not understand the documentation on running the validation, so I'll let the PR CI take care of it)

### DIFF
--- a/Detections/SigninLogs/MFARejectedbyUser.yaml
+++ b/Detections/SigninLogs/MFARejectedbyUser.yaml
@@ -29,9 +29,13 @@ entityMappings:
     fieldMappings:
       - identifier: FullName
         columnName: AccountCustomEntity
+      - identifier: AadUserId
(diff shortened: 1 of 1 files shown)


aprakash13 (approved):
Thanks @malexmave

PR #4402, opened 2022-03-13
ASIM Dns - Rearrangement

yaronMSFT:
- Rearranging files
- Generating uniform readme files


oshezaf (changes requested):
Overall OK. Just make sure that EventProduct and EventVendor are in the doc (see registry notes).

oshezaf (changes requested):
- You need to change the file location also in the full deployers.
- I think you did not review yet the EventVendor/EventProduct fields.

oshezaf (approved):

PR #4404, opened 2022-03-13
WebSession - Rearrangement

yaronMSFT:
- Rearranging files
- Generating uniform readme files


Parsers/ASimWebSession/Parsers/vimWebSessionSquidProxy.yaml
```diff
@@ -111,7 +118,12 @@ ParserQuery: |
     | extend    
       DstHostname = tostring(DstFQDNparts[0]),
       DstDomain = strcat_array(array_slice(DstFQDNparts,1,-1),"."),
-      DstDomainType = "FQDN"
+      DstDomainType = "FQDN",
+      ASimMatchingIpAddr = case(array_length(ipaddr_has_any_prefix) == 0 ,"",
```
oshezaf, line 122: 1. If the address is not in Src or Dst, the record should not be returned.  ipaddr_has_any_prefix is for Src and Dst, not just anything in the record. For example, it should not trigger if the URL has a parameter with the IP address.
2. Since this becomes post filtering, it should be higher up.

Parsers/ASimWebSession/Parsers/vimWebSessionzScalerZIA.yaml
```diff
@@ -82,6 +82,7 @@ ParserQuery: |
     and (isnull(endtime) or TimeGenerated <= endtime) 
     and ((array_length(httpuseragent_has_any) == 0) or (RequestClientApplication has_any (httpuseragent_has_any)))
     and ((array_length(srcipaddr_has_any_prefix) == 0) or has_any_ipv4_prefix(SourceIP, srcipaddr_has_any_prefix))
+    and ((array_length(ipaddr_has_any_prefix) == 0) or has_any_ipv4_prefix(SourceIP, ipaddr_has_any_prefix) or has_any_ipv4_prefix(DestinationIP, ipaddr_has_any_prefix) )
```
oshezaf, line 85: I wonder whether array_concat in the previous line would not be more efficient.

Parsers/ASimWebSession/Parsers/vimWebSessionzScalerZIA.yaml
```diff
@@ -149,7 +150,12 @@ ParserQuery: |
     EventSeverity = case (ThreatRiskLevel > 90, "High", ThreatRiskLevel > 60, "Medium", ThreatRiskLevel > 10, "Low",  "Informational"),
     DstAppType = "SaaS application",
     DstDomainType = iff (DstHostnameNotAddr, "FQDN", ""),
-    SrcUsernameType = "UPN"
+    SrcUsernameType = "UPN",
+    ASimMatchingIpAddr = case(array_length(ipaddr_has_any_prefix) == 0 ,"",
```
oshezaf, line 154: See note above about post-filtering.

oshezaf (changes requested):
Make sure you update the templates using this one (AsimFullDeployed is the only one I think) to use the new parameters when calling the Web Deployer.

Parsers/ASimWebSession/Parsers/vimWebSessionSquidProxy.yaml
YaronFruchtmann: You are right. not sure why I commented line 96. now uncommented.

Parsers/ASimWebSession/Parsers/vimWebSessionSquidProxy.yaml
```diff
@@ -111,7 +117,12 @@ ParserQuery: |
...
```
oshezaf, line 121: You should really do the case as post filtering and filter on ASimMatchingIpAddr != "", to reduce the number of calls to has_any_ipv4_prefix

Parsers/ASimWebSession/Parsers/vimWebSessionzScalerZIA.yaml
oshezaf, line 154: again, do this is the pre-filter to reduce calls to has_any_ipv4

oshezaf (changes requested):
I think that the handling of the new parameter can be optimized. See comments.

Parsers/ASimWebSession/Parsers/vimWebSessionzScalerZIA.yaml
YaronFruchtmann: what lists would you concatenate?

Parsers/ASimWebSession/Parsers/vimWebSessionzScalerZIA.yaml
YaronFruchtmann: This is not a filter. The events are already filtered at line 85. BUT: since these are already filtered I CAN improve this a bit. 
I could: extend SrcMatch = has_any_ipv4_prefix(SrcIpAddr,ipaddr_has_any_prefix), DstMatch = has_any_ipv4_prefix(DstIpAddr,ipaddr_has_any_prefix)
and then use this column instead of calling has_any_ipv4.

oshezaf: Exactly, use your extend message in line 85 (breaking the where clause), and use the values in the extend in 154.

Parsers/ASimWebSession/Parsers/vimWebSessionSquidProxy.yaml
oshezaf: Sure, that said, using the same method for running has_any_ipv4_prefix less times you suggested for Zscaler works here as well.

oshezaf (changes requested):
Continuing the thread as a new review. If we are still not in agreement let's talk.

oshezaf (approved):

PR #4443, opened 2022-03-18
Renamed the "Workspace Region" parameter to "location"

fidelcasto:
   
   Change(s):
   - Renamed the "Workspace Region" parameter to "location" in each file where it was used.

   Reason for Change(s):
   - Causes issue when attempting to automate the deployment using Azure DevOps since the parameter have spaces in it.  It also seems to be in best practice to use CamelCase.  

   Version Updated:
   -  N/A

   Testing Completed:
   - Yes

   - Yes

### DIFF
--- a/Parsers/ASimProcessEvent/ARM/FullDeploymentProcessEvent.json
+++ b/Parsers/ASimProcessEvent/ARM/FullDeploymentProcessEvent.json
@@ -25,7 +25,7 @@
       "properties": {
         "mode": "Incremental",
         "templateLink": {
-          "uri": "https://raw.githubusercontent.com/fidelcasto/Azure-Sentinel/master/Parsers/ASimProcessEvent/ARM/imProcess/imProcess.json",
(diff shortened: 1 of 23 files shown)


oshezaf:
Hi @fidelcasto ,

The change to Workspace Region is intentional and needed to solve repeating issues when templates are deployed interactiely. Since "loation" is ambiguous, users where either lost as to what to input, or povided the wrong region.

Now, obviously, we need to address the DevOps challenge, so i suggest to remove the space, but keep the text, i.e. WorspageRegion.

~ Ofer

fidelcasto:
Hi @oshezaf ! 
Thanks for the reply.  Of course that works too.  I used "location" since the other parsers ARM Templates used "location" as parameter name.  Example : https://github.com/Azure/Azure-Sentinel/blob/master/Parsers/ASimAuthentication/FullDeploymentAuthentication.json

That could probably be a request to use the same parameter name in every FullDeployment scripts.

I will make the change tomorrow to the Pull Request.

Thanks a lot !

fidelcasto:
Hi @oshezaf,
I just pushed the changes to the Pull Request to use 'WorkspaceRegion' as parameter name in the ARM Templates.
Thanks a lot !

oshezaf (approved):

PR #4452, opened 2022-03-20
Update-VIPUsers-Watchlist-from-AzureAD-Group

BenjiSec:
Playbook that is syncing members from Azure AD group to the watchlist.


lior-tamir (approved):
Approved but let's add playbook template metadata and light screenshot so we can add this to templates tab

PR #4520, opened 2022-03-28
Updating api version and version

mariussm:
Change(s):

- Changed ARM apiVersion of Microsoft.OperationalInsights/workspaces to 2021-06-01
- "Version" of saved seraches to 2, as it seems to be what is actually present in Azure (what if always wants to overwrite version 2 with 1)

Reason for Change(s):

- Remove what-if noise
- Remove preview api

Testing Completed:

- Deployed successfully

Version updated:

- All contentversion updated from 1.0.0.0 to 1.0.0.1

- Yes


oshezaf:
Hi @mariussm : the templates are derived files generated by [ASimYaml2ARM](https://github.com/Azure/Azure-Sentinel/tree/master/ASIM/dev/ASimYaml2ARM). The changes should be applied to the "templates of templates" of this tool and will propagate to the ARM templates as we rebuild. 

There are two reasons for that:
- Ensuring that your changes remain on the next build.
- When we build, we run a test cycle.

PR #4521, opened 2022-03-29
Updated Infoblox parser

rushriva:
New solution created for Infoblox NIOS with updated Parsers, Workbook and watchlist. 

Reason for Change(s):
Infoblox parser is updated for performance enhancement. 

Testing Completed:
Yes
Yes

### DIFF
--- a/Solutions/Infoblox NIOS/Data Connectors/Connector_Syslog_Infoblox.json
+++ b/Solutions/Infoblox NIOS/Data Connectors/Connector_Syslog_Infoblox.json
@@ -14,24 +14,24 @@
         "sampleQueries": [ 
         { 
             "description" : "Total Count by DHCP Message Types", 
-            "query": "InfobloxNIOS \n| where LogType == \"dhcpd\"  \n| where Type startswith \"DHCP\" \n| summarize count() by Type"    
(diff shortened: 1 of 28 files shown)


anki-narravula (commented):
- Watchlist name should not be "datasource" it should be in resemblance with data source, like InfloboxDevices or InflobloxComputerList make sense

- Please change that in watchlist and all of the parsers.

- Can we create one common query which consolidate data from all other individual parsers so that we can use this if we required to query all the data 

` InfloboxNIOSEvents = union isfuzzy=true Infoblox_dhcpexpire, Infoblox_dhcpinform, Infoblox_dhcpoption etc etc`

- Since connectivity criteria query causing issue, we can change query to use new parser (This chage must be for solution version ONLY, should NOT made any chnage for the connector gallery version. otherwise it will impact existing customers.  - 

`InfloboxNIOSEvents \n      | summarize LastLogReceived = max(TimeGenerated)\n            | project IsConnected = LastLogReceived > ago(7d)"`

anki-narravula:
@NikTripathi  pls review workbook

anki-narravula (approved):
Looks good. Approved. @NikTripathi pls merge this and help with package publish

PR #4538, opened 2022-03-30
Add more PowerShell apps in UnusualGuestActivity.yaml

ep3p:
   
   Change(s):
   - Add one more PowerShell app
   
   Reason for Change(s):
   - It may be useful to check attempts to use this app by guests. This may trigger when connecting to Azure Information Protection service.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes
   
   Please, could you check if this attempts by guests (failed or not) could be possible or not? Thank you in advance.


aprakash13 (approved):
Thanks for adding this @ep3p.

Issue #4541, opened 2022-03-31
Reset-AADPassword playbook not working with default OOB template


mgijo:
Hi Team,

I would like to provide a feedback in regards to the automation template called Reset-AADPassword. 

This logic app when deployed OOB through Sentinel Playbooks blade fails at 2 different places:-

1)	At the HTTP connector in the for-each loop 

![image](https://user-images.githubusercontent.com/83291280/160958370-8eed91cb-c5ea-4d0b-ab03-b6f82fb1f3c2.png)

While testing the password to be sent to user’s manager it doesn’t meet minimum complexity requirements based on the GUID value in the ‘Set variable’ connector.

![image](https://user-images.githubusercontent.com/83291280/160958395-8134bb5d-e90e-4200-90dc-53a0ea09fe3b.png)

If we change the value to 9 or 10 the logic app works fine as the length of the password meets minimum complexity. Could you review and then update the value here on the template to 9 or 10 atleast?

2)	Logic app also seems to fail at the final section where comment needs to be added to the triggered incident

 
![image](https://user-images.githubusercontent.com/83291280/160958413-3f23d794-5d0b-4ff1-a121-5ca29b32c15d.png)

The PowerShell script provided here Azure-Sentinel/Playbooks/Reset-AADUserPassword at master · Azure/Azure-Sentinel · GitHub assigns “Password Administrator” role. Could you add also another role such as “**Microsoft Sentinel Responder**” role to the managed identity? So that this final connector in the logic app has sufficient privileges to execute this action?

BenjiSec:
Hi @mgijo, it is fixed now. It will take couple of days until it's replicated in the Playbooks Template in the Sentinel, but you can deploy it from GitHub as soon as my PR is approved - https://github.com/Azure/Azure-Sentinel/pull/4659
@sarah-yo - you can close this Issue.

PR #4543, opened 2022-03-31
Possible Fix for #4541

samikroy:
Possible Fix for #4541 

Reset-AADPassword playbook not working with default OOB template

- Update PS script
- Updated the logic app / playbook definition.

### DIFF
--- a/Playbooks/Reset-AADUserPassword/alert-trigger/azuredeploy.json
+++ b/Playbooks/Reset-AADUserPassword/alert-trigger/azuredeploy.json
@@ -236,7 +236,7 @@
                         "Set_variable": {
                             "inputs": {
                                 "name": "Password",
-                                "value": "@{substring(guid(), 0, 8)}"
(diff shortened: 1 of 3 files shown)


samikroy:
@anki-narravula & @v-spadarthi  - Please let me know for further changes to this.

samikroy:
@lior-tamir & @BenjiSec - Could you please help merging this !

BenjiSec:
@samikroy - I'll have to change few things and will make new PR. Need to add control if user doesn't have a manager and will change connection for checking manager to HTTP. Also this is template so need to change template versioning as well as add release notes.

Yaniv-Shasha:
template error

samikroy:
> @samikroy - I'll have to change few things and will make new PR. Need to add control if user doesn't have a manager and will change connection for checking manager to HTTP. Also this is template so need to change template versioning as well as add release notes.

@BenjiSec - Thank you for sharing this.

PR #4552, opened 2022-04-01
Improve SigninBruteForce-AzurePortal.yaml

ep3p:
Change(s):
   1. Comment lines that were not used in the results (because of the summarize)
   2. Use correct column for Status (instead of DeviceDetail)
   3. Summarize by UserId, instead of UserPrincipalName
   4. Put "Azure Portal" condition ahead of extends.

   Reason for Change(s):
   1. They were not used.
   2. Although it was not used, it was not correct.
   3. Column UserPrincipalName does not show always the UPN, sometimes it shows the UserId, specially in erroneous ResultTypes
   4. CPU may be wasted on unwanted events

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes
   
   Additionally, maybe the ResultTypes accepted as "Success" should increase or be reviewed, in comparison with other detections that define ResultTypes lists.


aprakash13 (approved):
Thanks @ep3p.

PR #4565, opened 2022-04-04
Update ExplicitMFADeny detection

wez3:
Change(s):
   - If a user selects "Report Fraud" after declining a MFA push notification, the Status is "MFA denied; Phone App Reported Fraud" instead of "MFA Denied; user declined the authentication". Added a OR statement to detect both situations.

   Reason for Change(s):
   - Expands detection when Fraud Report is enabled in the tenant.

   Version Updated:
   - Yes (1.0.1)

   Testing Completed:
   - Yes

   - Yes


aprakash13 (approved):
Thanks @wez3

PR #4592, opened 2022-04-07
Updating workbook AKS for one ICM

NikTripathi:
Updating Workbook for ICM Incident 295158206 : Enabled Defender for Containers. However ,unable to observe in the Workbook 'Azure Kubernetes Service (AKS) Security'

   Change(s):
   - Added OR condition in query.

   Reason for Change(s):
   - Old query only works in subscriptions where still the old separate values for Kubernetes and Containers are there. So updated the query so that it can work for all the subscription.
   
   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


NikTripathi:
Hi @yoweiz I have made these changes regarding one ICM "[Incident 295158206](https://portal.microsofticm.com/imp/v3/incidents/details/295158206/home) : Enabled Defender for Containers. However ,unable to observe in the Workbook 'Azure Kubernetes Service (AKS) Security'" .
Please take a look and let me know if something else needs to be updated. 
Thanks.

yoweiz (approved):

PR #4620, opened 2022-04-12
Queries related to Tarrask malware related blog post.

aprakash13:
Change(s):
   - Queries related to Tarrask malware related blog post.

   Reason for Change(s):
   -New queries 

   Version Updated:
   -  Not applicable (New)

   Testing Completed:
   -  Yes
   - 
   -


Detections/SecurityAlert/AVTarrask.yaml
```diff
@@ -0,0 +1,41 @@
...
+  - Persistence
+relevantTechniques:
+  - T1053
+query: |
+  let Tarrask_threats = dynamic(["HackTool:Win64/Tarrask!MS", "HackTool:Win64/Ligolo!MSR", "Behavior:Win32/ScheduledTaskHide.A"]);
+  DeviceInfo
+  | extend DeviceName = tolower(DeviceName)
+  | join ( SecurityAlert
+  | where ProviderName == "MDATP"
+  | extend ThreatFamilyName = tostring(parse_json(ExtendedProperties).ThreatFamilyName)
```
petebryan, line 27: So you want to project out ExtentedProperties.ThreatName here to match on those values. ThreatFamilyName in this instance would just be "Tarrask" so wouldn't match

Detections/SecurityAlert/AVTarrask.yaml
```diff
@@ -0,0 +1,41 @@
...
+triggerThreshold: 0
+tactics:
+  - Persistence
+relevantTechniques:
+  - T1053
+query: |
+  let Tarrask_threats = dynamic(["HackTool:Win64/Tarrask!MS", "HackTool:Win64/Ligolo!MSR", "Behavior:Win32/ScheduledTaskHide.A"]);
+  DeviceInfo
+  | extend DeviceName = tolower(DeviceName)
+  | join ( SecurityAlert
```
petebryan, line 25: You might want to make this a rightouter join type in case users don't have DeviceInfo data onboarded - that way it will still return results.

petebryan (changes requested):
Can you double check that CSV as well, as it doesn't seem to be a regular CSV (not sure what format it is).

petebryan (approved):

thmcelro (approved):

PR #4635, opened 2022-04-13
Kyle yuasa/officeactivity query tuning

Kyle-Yuasa:
Change(s):
   - Updated new_sharepoint_downloads_by_UserAgent to use historical ASN info from SigninLogs 
   -Updated new_sharepoint_downloads_by_IP to use historical ASN info from SigninLogs
   
Reason for Change(s):
   - Too many results to be useful for hunting when only looking at 14 days of IP and UserAgent OfficeActivity data
   - Incorporating an additional signal in AutonomousSystemNumbers with high block rates helps reduce result set considerably

   Testing Completed:
   - Yes

   - Yes


aprakash13 (approved):
Thanks @Kyle-Yuasa

PR #4637, opened 2022-04-13
Created detect-uac-elevation query

mjmelone:
   
   Change(s):
   - new query created

   Reason for Change(s):
   - new query, no other example known

   Version Updated:
   - Not required as this is not a detection or analytic query

   Testing Completed:
   - Yes

   - Yes


aprakash13 (approved):
Thanks @mjmelone

PR #4654, opened 2022-04-15
Put constraints on number of assets in export chunk. Enforce limit of 30 MB on data uploaded to sentinel

Mangusingh-te:
Change(s):
   - Put limit on tenable.io assets and vulns exports chunk size.
   - Divides individual export chunk into sub-chunk to enforce 30 MB size of each sub-chunk.

   Reason for Change(s):
   - Some containers with huge amount of data, were returning export chunks in few hundreds of MBs(overall export size more than few GBs).
   - It resulted into
       -  Taking too much time to download export chunk in activity function.
       -  OOM result causing function app to crash.
       -  Chunk failing to get uploaded to sentinel(due to 30 MB limit).

   Version Updated:
   - N/A

   Testing Completed:
   - Yes

   - Yes


NikTripathi (changes requested):
Hi @Mangusingh-te .py files are modified. So I think we need to update the .zip file also with the latest init files. Please make the required changes in zip also.

Mangusingh-te:
> Hi @Mangusingh-te .py files are modified. So I think we need to update the .zip file also with the latest init files. Please make the required changes in zip also.

Hi @NikTripathi I have updated the PR with zip file containing latest durable function code as well as bumped the solution version too.

NikTripathi (approved):
As confirmed by customer it is working fine for them. Approving and merging.

PR #4665, opened 2022-04-17
Update azuredeploy.json

AcceleryntSecurityDev:
Change(s):
- updated sensitive parameters to secure string

   Reason for Change(s):
   - security improvement

   Version Updated:
   - No

   Testing Completed:
   - No

   - Yes

### DIFF
--- a/Playbooks/Guardicore-Import-Incidents/azuredeploy.json
+++ b/Playbooks/Guardicore-Import-Incidents/azuredeploy.json
@@ -11,12 +11,16 @@
             "type": "String"
         },
         "GCURL": {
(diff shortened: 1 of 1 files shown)


sreedharande (changes requested):
@AcceleryntSecurityDev - Delete "defaultValue" to make it Required Column, as GCPassword is required value for the Playbook

sreedharande (changes requested):
@AcceleryntSecurityDev - Remove Default values for GCURL, GCUserame and GCPassword. and add defaultvalue text to "metadata": {
            "description": ""
        }

sreedharande (approved):

PR #4671, opened 2022-04-18
Add more cases to Office_MailForwarding.yaml

ep3p:
Change(s):
   1. Add more operation cases where mails can be forwarded to other mail addresses.
   2. In case of the scheduled query specify a queryfrequency.
   3. Remove the final ```distinct```.
   4. Put a limit to make_sets.

   Reason for Change(s):
   1. Other rules have these cases.
   2. If not, the scheduled query will generate the same alert every day for 7 days straight.
   3. It is not necessary, it was already summarized, and the dynamic of UserId was a set, not a list.
   4. Just as a good practice.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes
   
   Are ```in~``` and ```=~``` operators necessary instead of ```in``` and ```==```?
   Also, I don't understand why to summarize by ClientIP, is there that much noise?

### DIFF
--- a/Detections/OfficeActivity/NRT_Office_MailForwarding.yaml
+++ b/Detections/OfficeActivity/NRT_Office_MailForwarding.yaml
@@ -15,7 +15,6 @@ relevantTechniques:
   - T1114
   - T1020
 query: |
-
   OfficeActivity
   | where OfficeWorkload =~ "Exchange"
   | mv-apply DynamicParameters = todynamic(Parameters) on (summarize ParsedParameters = make_bag(pack(tostring(DynamicParameters.Name), DynamicParameters.Value)))
@@ -33,7 +32,6 @@ query: |
   | where DistinctUserCount > 1
   | mv-expand UserId
   | extend UserId = tostring(UserId)
-
 entityMappings:
   - entityType: Account
     fieldMappings:
--- a/Detections/OfficeActivity/NRT_Office_MailForwarding.yaml
+++ b/Detections/OfficeActivity/NRT_Office_MailForwarding.yaml
@@ -15,47 +15,25 @@ relevantTechniques:
   - T1114
   - T1020
 query: |
+
   OfficeActivity
   | where OfficeWorkload =~ "Exchange"
-  | where Parameters has "ForwardingSmtpAddress"
-  | extend parsed = parse_json(Parameters)
-  | mv-expand parsed
-  | where parsed.Name == "ForwardingSmtpAddress"
-  | extend parameterName = tostring(parsed.Name), fwdingDestination = tostring(parsed.Value)
-  | where isnotempty(fwdingDestination)
-  | extend ClientIPOnly = case(
-  ClientIP has "." and ClientIP has ':', tostring(split(ClientIP,":")[0]),
-  ClientIP has "." and ClientIP has '-', tostring(split(ClientIP,"-")[0]),
-  ClientIP has ']-', tostring(trim_start(@'[[]',tostring(split(ClientIP,"]")[0]))),
-  ClientIP has ']:', tostring(trim_start(@'[[]',tostring(split(ClientIP,"]")[0]))),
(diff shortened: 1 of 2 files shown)


aprakash13:
Requesting @petebryan to help with these as he probably helped build some of these original queries.

vmanojreddy:
@petebryan  Please review and provide updates. thanks!!!

vmanojreddy:
@petebryan Please review and provide updates. thanks!!!

Detections/OfficeActivity/Office_MailForwarding.yaml
```diff
@@ -19,49 +19,28 @@ relevantTechniques:
...
-  isnotempty(ClientIP), ClientIP,
-  isnotempty(ClientIP_), ClientIP_,
-  "IP Not Available"
-  )
-  | extend UserId = iff(isempty(UserId), UserId_, UserId)
-  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), DistinctUserCount = dcount(UserId), UserId = make_set(UserId), 
-  Ports = make_set(Port), EventCount = count() by fwdingDestination, ClientIP = ClientIPOnly 
-  | where DistinctUserCount > 1
+  | where TimeGenerated > ago(queryperiod)
+  //| where OfficeWorkload =~ "Exchange"
```
petebryan, line 26: Is there a reason to comment out this line?

Detections/OfficeActivity/NRT_Office_MailForwarding.yaml
```diff
@@ -17,45 +17,21 @@ relevantTechniques:
...
-  isnotempty(ClientIP), ClientIP,
-  isnotempty(ClientIP_), ClientIP_,
-  "IP Not Available"
-  )
-  | extend UserId = iff(isempty(UserId), UserId_, UserId)
-  | summarize StartTimeUtc = min(TimeGenerated), EndTimeUtc = max(TimeGenerated), DistinctUserCount = dcount(UserId), UserId = make_set(UserId),
-  Ports = make_set(Port), EventCount = count() by fwdingDestination, ClientIP = ClientIPOnly
+  | mv-apply DynamicParameters = todynamic(Parameters) on (summarize ParsedParameters = make_bag(pack(tostring(DynamicParameters.Name), DynamicParameters.Value)))
+  | evaluate bag_unpack(ParsedParameters, columnsConflict='replace_source')
+  | extend DestinationMailAddress = tolower(case(
```
petebryan, line 22: Given this is an NRT rule it would be good have a check early on to see if Parameters has any of ForwardTo, RedirectTo, or ForwardingSmtpAddress before parsing out to improve performance.

petebryan (changes requested):

Detections/OfficeActivity/NRT_Office_MailForwarding.yaml
ep3p: @petebryan Added.

Detections/OfficeActivity/Office_MailForwarding.yaml
ep3p: @petebryan Maintenance of these two versions of the rule should be as easy as possible, so I preferred to keep both versions as identical as possible. This condition ```| where Operation in ("Set-Mailbox", "New-InboxRule", "Set-InboxRule")``` is not possible in the NRT version, so the condition ```"| where OfficeWorkload =~ "Exchange"``` had to be  added in the first NRT version and did not exist in the scheduled version. As this line was not strictly needed in the scheduled version, I kept it commented.

After adding the condition ```| where Parameters has_any ("ForwardTo", "RedirectTo", "ForwardingSmtpAddress")``` I think the commented line should instead be ```| where Operation in ("Set-Mailbox", "New-InboxRule", "Set-InboxRule")```. I think both versions should try to be similar, if you think this line should be uncommented or deleted, let me know, feel free to commit any change.

petebryan (approved):

PR #4684, opened 2022-04-20
Palo Alto Threat signatures from Unusual IP addresses - Fusion scenario rule

ashwin-patil:
   
   Change(s):
   - new rule for PAN threat signatures from unusual IP addresses historically not seen
   - Also it will be used to correlated with Tor alert for Fusion scenario
   - Hourly frequency is required to correlate with MDE alert and create fusion incident

   Reason for Change(s):
   - New scheduled rule for PAN-MDE fusion scenario 

   Version Updated:
   - See guidance below

   Testing Completed:
   - Tested on sample workspaces for noise and performance efficiency

   - See guidance below


aprakash13 (approved):
Thanks @ashwin-patil .

PR #4689, opened 2022-04-21
Ignore expired token due to signin frequency expired

mariussm:
Change(s):
   - Added ignore for two very noisy result types

   Reason for Change(s):
   - Noise

   Version updated:
   - Yes

   Testing Completed:
   - Yes


aprakash13 (approved):
Thanks @mariussm

PR #4703, opened 2022-04-24
renaming non filtering parsers to ASim

YaronFruchtmann:
Non filtering parser:
Adding disabled parameter
renaming to ASim_*


oshezaf (changes requested):
We can't do it that way. You need to do it one parser by one. This is production code and we can't merge without vim functions. Convert eash vim to ASIM when creating its new vim.

oshezaf (approved):

PR #4743, opened 2022-04-28
Add filter params to Process Terminate MS Sec Event

YaronFruchtmann:
Add filter params to Process Terminate MS Sec Event


oshezaf (changes requested):
Same comments: order of fliters, plus filter on actoruser.

oshezaf (approved):

PR #4772, opened 2022-05-02
Fixing Process parameters

YaronFruchtmann:


Parsers/ASimProcessEvent/Parsers/vimProcessCreateMD4IoT.yaml
```diff
@@ -79,16 +79,19 @@ ParserQuery: |
     // --------------------------------------------------------------------------------------
     | extend
       EventDetails = todynamic(EventDetails)
+    | where EventDetails.EventType != 'EXIT'
```
oshezaf, line 82: Filtering has to be done earlier as to not work on process terminate so much. As you did in terminate....

oshezaf (changes requested):

Parsers/ASimProcessEvent/Parsers/vimProcessCreateMD4IoT.yaml
YaronFruchtmann: with terminate I require inclusion, and later verified it is included in the correct place.
Here the requirement is NOT to include, so I am risking omitting cases where "EXIT" appears elsewhere

oshezaf: Not if you check for the entire fragment that includes EXIT, after all it is JSON and not free text.

YaronFruchtmann: After offline discussion, will not change now. 
TODO: in prefilter using the full key value segment of the EXIT value

oshezaf (approved):

PR #4782, opened 2022-05-04
Update AuthenticationMethodsChangedforPrivilegedAccount.yaml

TheAlistairRoss:
Commented out the final summarise line as Target entity mapping wasn't occurring. Provided a description explaining this. Also added additional entity mapping for the Initiator

   Change(s):
   - Updated presentation logic and entity mapping for AuthenticationMethodsChangedforPrivilegedAccount.yaml

   Reason for Change(s):
   - Target entities were not being mapped due to aggregation. Provided an alternative with comments, using the same output schema (Except Targets changes between dynamic and string depending on which line is used). This allows for Target users to be mapped to entities if required.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


aprakash13 (approved):
Thanks.

PR #4796, opened 2022-05-05
Better MTD Template Spec Package

elforb:
Change(s):
   - Creating Better MTD Template Spec package with Data Connector and Workbook

   Reason for Change(s):


anki-narravula (changes requested):
Data Connector version is coming as 2.0.0 but it should not be the case, please regenerate the package using latest changes
Same case for workbook as well

In mainTemplate pls change {publisherID}.{offerID} to publisherId_test.offerId_test  (brackets are not required)
author - currently it is Microsofr, pls change to <<Provider Name>>
email - currently it is support@microsoft.com,  pls change to isv@email.com

So that while we send instructions to ISVs to change these values it will be uniform

anki-narravula (approved):
Looks good. Approved.

PR #4840, opened 2022-05-09
Watchguard Firebox Template Spec

elforb:
Change(s):
- Creating Watchguard Firebox Template Spec package with Data Connector and Parser

Reason for Change(s):


anki-narravula (commented):
Waiting for publisher Id from Eric. Everything else is good.

anki-narravula (changes requested):
2.0.0.zip mainTemplate is not update 
image

anki-narravula (approved):
Looks good Approved

PR #4856, opened 2022-05-09
Forcepoint DLP Template Spec Package

elforb:
Change(s):
- Creating Forcepoint DLP Template Spec package with Data Connector and Workbook
Reason for Change(s):


anki-narravula (approved):
Looks Good. Approved.

PR #4857, opened 2022-05-09
Fixing missing day due to midtime usage

shainw:
Change(s):
   - Removed usage of midtime value for time frame

   Reason for Change(s):
   - midtime filter on timegenerated was causing a gap in days for historical comparison

   Version Updated:
   - Not needed for hunting queries

   Testing Completed:
   - Yes

   - Yes


ashwin-patil (approved):
Changes look good across queries

PR #4942, opened 2022-05-16
Update CODEOWNERS order

shainw:
Change(s):
   - adjust code owner order

   Reason for Change(s):
   - to see if it resolves issues with who is considered a code owner

   Version Updated:
   - N/A

   Testing Completed:
   - This is the test. ;)

   - Will do


aprakash13 (approved):
Thanks Shain.

PR #4982, opened 2022-05-17
Migrate to Teams as code owners

shainw:
Change(s):
   - Code owners file now uses Teams references instead of individuals.

   Reason for Change(s):
   - This allows for better control and updating.

   Version Updated:
   - N/A

   Testing Completed:
   - Yes, file is valid
   - Tested various scenarios when building teams and assigning code owner and role level for each team

   - Yes


aprakash13 (approved):
Thanks Shain.

PR #5008, opened 2022-05-19
Fix little errors in detection MaliciousWAFSessions.yaml

ep3p:
Change(s):
   1. BlockvsSuccessRatio now is forced to be a real number.
   2. mv-expand the summarized left side of the join, instead of the right-side
   3. Put a limit size to make_set.
   4. Project the columns in a certain order.
   5. Harmonize some column names.

   Reason for Change(s):
   1. It was divided by an integer, and the operation returned an integer, not a ratio.
   2. I think this is more efficient. It was triplicating all the right records to be compared.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes, but, please could you test the efficiency of the query against a good dataset?

   - Yes

### DIFF
--- a/Detections/AzureDiagnostics/MaliciousWAFSessions.yaml
+++ b/Detections/AzureDiagnostics/MaliciousWAFSessions.yaml
@@ -20,41 +20,46 @@ tactics:
 relevantTechniques:
   - T1190
 query: |
-  let mode = 'Blocked'; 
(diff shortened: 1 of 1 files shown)


vmanojreddy:
@aprakash13 : Could you please review. Thanks!!

aprakash13 (approved):
Thanks @ep3p.

PR #5037, opened 2022-05-20
Workplace from Facebook Template Spec

elforb:
Change(s):
- Creating Workplace from Facebook Template Spec package with Data Connector and Parser

Reason for Change(s):


mkchiliveri (approved):
Looks good, Approved.

Tichandr (approved):

PR #5080, opened 2022-05-24
Add new Cisco Meraki CCP CLv2 data connector and updated content

devikamehra:
Added new Meraki data connector
   
   Change(s):
   - Added new CCP CLv2 based data connector for Cisco Meraki. This is using the REST API as the source.
   - Updated the Parser with the required changes to support the new data type from REST API based data connector.
   - Updated the Workbook to support both old and new data type

   Reason for Change(s):
   - Requirement to have a REST API based data connector

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes


anki-narravula (approved):
Looks good. Approved

PR #5086, opened 2022-05-24
Update UserAssignedPrivilegedRole.yaml

Lodewyk-Git:
Adding a section to not alert for PIM when uncommented

   Change(s):
   - | where Initiator != "MS-PIM"

   Reason for Change(s):
   - Alerts when user activates PIM assignment, which is not of high severity in my opinion.


aprakash13 (approved):
Thanks @Lodewyk-Git

PR #5087, opened 2022-05-24
[Fix] Extract ClientIp for proper entity extraction

rjaakke:
Change(s):
   - Updated Detections/OfficeActivity/RareOfficeOperations.yaml

   Reason for Change(s):
   - Extract ClientIp for proper entity extraction, ClientIPs with port number were not mapped to an entity
   
   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


rjaakke:
Somehow the IP entities aren't always added to the alert in our environment.
I tested the rule to make sure the column values are correct and only show the IP.
For some reason entities still are not always mapped, is this some background process that will run eventually?

v-spadarthi:
@aprakash13 :Please review detections.Thanks !

aprakash13 (approved):
Thanks @rjaakke

PR #5122, opened 2022-05-25
Microsoft Project Template Package

elforb:
Change(s):
- Creating Microsoft Project Template Spec package with Data Connector.
- _Data Connector sourced from connector json downloaded via sentinel, using special feature flag, as connector was unavailable here in github._

Reason for Change(s):


mkchiliveri (approved):
Looks good, Approved.

PR #5133, opened 2022-05-26
Updating the Azure Portal Brute Force Query 

aprakash13:
Updating the Azure Portal Brute Force Query bunch of failures then a success

   Change(s):
   - Updating the Azure Portal Brute Force Query to enforce sequence i.e bunch of failures then a success

   Reason for Change(s):
   - Reduce FP

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

### DIFF
--- a/Detections/SigninLogs/SigninBruteForce-AzurePortal.yaml
+++ b/Detections/SigninLogs/SigninBruteForce-AzurePortal.yaml
@@ -72,3 +72,8 @@ entityMappings:
       - identifier: FullName
         columnName: AccountCustomEntity
   - entityType: IP
+    fieldMappings:
+      - identifier: Address
+        columnName: IPCustomEntity
+version: 1.0.2
+kind: Scheduled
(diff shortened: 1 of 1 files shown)


Detections/SigninLogs/SigninBruteForce-AzurePortal.yaml
```diff
@@ -76,5 +75,5 @@ entityMappings:
     fieldMappings:
       - identifier: Address
         columnName: IPCustomEntity
-version: 1.0.1
+version: 1.0.2
```
shainw, line 78: This is a pretty major change, let's go with 2.0.

shainw (changes requested):
Change the version to a major version change, otherwise good.

shainw (approved):

Detections/SigninLogs/SigninBruteForce-AzurePortal.yaml
aprakash13: Thanks @shainw. Updated the version as suggested.

PR #5148, opened 2022-05-27
Snowflake analytic content update

spsocprime:
Change(s):
   - Rule severity level adjusted

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


aprakash13 (approved):
Thanks, Looks OK from a query perspective.

PR #5188, opened 2022-06-02
Make the "-EncodedCommand " parameter case insensitive while parsing

princemathew:
Change(s):
   - Made the parsing of encoded command case insensitive

   Reason for Change(s):
   - Encoded PowerShell scripts like "powershell.exe -encodedcommand <cmd>" will not get detected for the current rule implementation.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


aprakash13 (approved):
Thanks.

PR #5247, opened 2022-06-08
Add TimeRange parameter instead of defined time context

Official-James:
   
   Change(s):
   - Updated [Solutions/GitHub/Workbooks/GithubWorkbook.json](https://github.com/Azure/Azure-Sentinel/compare/master...Official-James:patch-1#diff-191566d9e044781fe2e98ff89f0ac0416e89548bcf9fe5e7f57f7b9953b2fb72) to replace timeContext with TimeRange

   Reason for Change(s):
   - If used as it is, it will filter based on the timeContext defined by the creator and would require manual editing. By replacing with timeContextFromParameter and setting it to TimeRange, it enables users to use the time range filter within the workbook (no editing required).

   Version Updated:
   - Yes (from v1.0 to v1.1)

   Testing Completed:
   - Yes

   - Yes

### DIFF
--- a/Solutions/GitHub/Workbooks/GithubWorkbook.json
+++ b/Solutions/GitHub/Workbooks/GithubWorkbook.json
@@ -1,5 +1,5 @@
 {
-  "version": "Notebook/1.1",
+  "version": "Notebook/1.0",
   "items": [
     {
       "type": 1,
--- a/Solutions/GitHub/Workbooks/GithubWorkbook.json
+++ b/Solutions/GitHub/Workbooks/GithubWorkbook.json
@@ -1,5 +1,5 @@
 {
-  "version": "Notebook/1.0",
+  "version": "Notebook/1.1",
   "items": [
     {
       "type": 1,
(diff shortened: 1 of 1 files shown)


vmanojreddy:
Hi @Official-James When we using version 1.1  in the work book it is not working it displaying below error like below.

image

But we changed to version 1.0 query's are failing. Provided screen shot below can you please check it once. thanks!!!

image

Official-James:
Hi @vmanojreddy I have changed the version (apologies) and this is now working as intended:

image

I am unsure why you are getting the error after changing the version to 1.0. The time context is the only change in this, not affecting the query. I can confirm that copying and pasting the workbook from this file works. The second screenshot you provide seems to be querying "GitHubAuditLogPolling CL", whereas the queries are actually using GitHubAuditData.

v-spadarthi:
@Official-James : Still we are getting below error 
image
Could you please provide the sample data then we will ingest and let you know the status.Thanks!

v-spadarthi:
Workbook looks good
image

v-spadarthi (approved):

NikTripathi (approved):

PR #5260, opened 2022-06-09
Take last versions of indicators in some TI map rules

ep3p:
Change(s):
   - Take last version of Indicator

   Reason for Change(s):
   - If not, the rules could be using outdated indicators.
   - Rest of TI map IP entity rules use this step.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


aprakash13 (approved):
Thank you so much @ep3p

PR #5269, opened 2022-06-10
Fix "Vlaue" dictionary key in URLEntity_OfficeActivity.yaml

ep3p:
Change(s):
   - Call the correct key "Value"

   Reason for Change(s):
   - If not, it will never resolve correctly. Although a user doesn't have to be in position 0 of the array... So a bigger fix might be needed.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


aprakash13 (approved):
Thanks @ep3p

PR #5273, opened 2022-06-10
Fix mistake in CrossServiceADXQueries.yaml

ep3p:
Change(s):
   - Take all elements from extract_all

   Reason for Change(s):
   - I was leaving behind possible results

   Testing Completed:
   - Yes

   - Yes


aprakash13 (approved):
Thanks @ep3p

PR #5321, opened 2022-06-15
Fix column names from external source

b401:
Change(s):
   - Removed no longer used external csv column names

   Reason for Change(s):
   - External source no longer includes Ipv6 addresses & some other fields

   Version Updated:
   - No
  
   Testing Completed:
   - Yes

   - Yes


aprakash13 (approved):
Thanks @b401

PR #5400, opened 2022-06-21
Fix small condition in UnusualGuestActivity.yaml

ep3p:
Change(s):
   - Distinguish between starting with live.com# ```live.com#example1@example2.com``` and containing live.com# ```example1_live.com#EXT#example2.onmicrosoft.com```

   Reason for Change(s):
   - As live.com is a valid domain, the parsing of the invited user might be incorrect in the current state.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


aprakash13 (approved):
Thanks @ep3p

PR #5403, opened 2022-06-21
New solution package Version 1.0.13 (analytic rule adjustments and new playbook)

amitcohen25:
Change(s):
   - For all 15 analytic rules: added new parameter called AlertManamgentUri to the project section + add it to the custom details section. Also, remove existing Entity Mapping and rely on the new Sentinel Entities mapping
   - New playbook called AD4IoT-AutoAlertStatusSync

   Reason for Change(s):
   - Customer request of Defender for IoT

   Version Updated:
   - Yes
   
   Testing Completed:
   - Yes

   - Yes


aprakash13 (approved):
Thanks @amitcohen25

PR #5442, opened 2022-06-23
Synack integration - latest changes before solution is published

andriykozynets:
Change(s):
   - improved logging
   - bug fixes

   Reason for Change(s):
   - insufficient logging
   - failure to create incident if description length is > 5000 symbols

   Version Updated:
   - no, as we didn't yet release any versions

   Testing Completed:
   - Yes

   - No


v-sabiraj (approved):
Looks good, as the changes are reflected in zip too.

PR #5470, opened 2022-06-27
Fix error in expanding IPAddresses in IPEntity_DnsEvents.yaml

ep3p:
Change(s):
   - Split multiple ip addresses by ```, ``` instead of ```,```

   Reason for Change(s):
   - Spaces are important when comparing the join key, so until now there were only matches with the first ip address (the one that did not start with a space), the rest of ip addresses would not match.

   Version Updated:
   - Yes

   Testing Completed:
   - No, please, could you check if DnsEvents tables **always** has ip addresses separated by ```, ``` instead of ```,``` ?

   - Yes


aprakash13 (approved):
Thanks @ep3p

PR #5501, opened 2022-06-29
Fix github workbook in the GH solution

wanpeng-cds:
   
   Change(s):
   - Uses TimeRange parameter in the queries

   Reason for Change(s):
   - The workbook has a TimeRange picker but the queries are not using it

   Version Updated:
   - No Need

   Testing Completed:
   - No

   - Yes


vmanojreddy:
@wanpeng-cds Please provide preview images. thanks!!!

wanpeng-cds:
image

This PR changes the value of Time Range in the red circle to 'TimeRange' which is the parameter defined previously.

v-spadarthi:
@wanpeng-cds : We could see same workbook changes were there in another [5247](https://github.com/Azure/Azure-Sentinel/pull/5247) because of that we are closing this PR.

PR #5509, opened 2022-06-30
Fix missing parentheses in MaliciousWAFSessions.yaml

ep3p:
I'm sorry, with the previous pull request accepted I saw this error. If needed, return the version to 1.0.1 , please.
   
   Change(s):
   - Add parentheses to clarify condition logic.

   Reason for Change(s):
   - If ```httpStatus_d``` is used by other events with ```Category``` different from ```ApplicationGatewayAccessLog```, they will effectively be added to the join unnecessarily, when the original rule only wanted ```ApplicationGatewayAccessLog``` events.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


aprakash13 (approved):
Thanks @ep3p for catching this and all your contributions.

PR #5579, opened 2022-07-08
Parse ClientIP addresses in IPEntity_OfficeActivity.yaml

ep3p:
   
   Change(s):
   - Extract the IP address value from ClientIP column.

   Reason for Change(s):
   - In specific workloads some addresses can end with a port (:\d) or a percent (%\d), they will not match TI records if not parsed.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


aprakash13 (approved):
Thanks @ep3p

PR #5735, opened 2022-07-27
Removed dependency of query and kind field check.

devikamehra:
Changed the validation query to support dummy analytical rule and hunting query files.
   
   Change(s):
   - Repositioned the skip code so that query and kind fields are not needed.

   Reason for Change(s):
   - Many analytical rules and hunting queries are moved to Solutions folder but are required to be present in individual galleries to ensure URLs in blogs are not broken. These changes ensure we have no dependency on query and kind fields.

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes


anki-narravula (approved):
Looks good. Approved

PR #5751, opened 2022-07-28
Fix errors in KeyvaultMassSecretRetrieval.yaml

ep3p:
Change(s):
   1. Remove ```isnotempty(``` conditions, that are removing desired results.
   2. Put a limit in make_set and make_list.
   3. Do not show by default all columns from ```arg_max(CalledIPAddress, *)```

   Reason for Change(s):
   1. Different operations do not have the same columns. For example ```SecretGet``` uses ```identity_claim_oid_g``` instead of ```identity_claim_http_schemas_microsoft_com_identity_claims_objectidentifier_g```. The same happened with ```clientInfo_s```, not every operation has this column.
   2. Put a limit on results size.
   3. It's difficult to interpret results.

The same happens with:

https://github.com/Azure/Azure-Sentinel/blob/ee97399b426a21878776c4c7b4835efc04b47393/Detections/AzureDiagnostics/TimeSeriesKeyvaultAccessAnomaly.yaml
https://github.com/Azure/Azure-Sentinel/blob/ee97399b426a21878776c4c7b4835efc04b47393/Detections/AzureDiagnostics/KeyVaultSensitiveOperations.yaml
https://github.com/Azure/Azure-Sentinel/blob/d11b7c5e8610b075bd0284044153c7a1e75d720d/Detections/AzureDiagnostics/NRT_KeyVaultSensitiveOperations.yaml
...

   Version Updated:
   - Yes

   Testing Completed:
   - No, this query needs to be checked against tenants that do not have some of the columns.

   - Yes


v-spadarthi:
@aprakash13 : Please review the Detections. Thanks!

v-mchatla:
Hi @aprakash13 
Can you please review and provide your feedback on the PR changes.
Thanks

v-spadarthi:
Hi @aprakash13
Can you please review and provide your feedback on the PR changes.
Thanks

NikTripathi:
@aprakash13 Please have a look.
Thanks

shainw:
@ep3p - We are having the original author, @ashwin-patil,  review these.  Thanks for the feedback as the queries originally written 3 years back and logging has changed for sure.

Detections/AzureDiagnostics/KeyvaultMassSecretRetrieval.yaml
```diff
@@ -1,58 +1,72 @@
...
-  | where requestUri_s !~ "None" and isnotempty(requestUri_s)
-  | where OperationName in~ (OperationList)   
-  ) on identity_claim_http_schemas_microsoft_com_identity_claims_objectidentifier_g 
-  | summarize EventCount=sum(count_), StartTimeUtc=min(TimeGenerated), EndTimeUtc=max(TimeGenerated), TimeTriggered=makelist(TimeGenerated),OperationNameList=make_set(OperationName), RequestURLList=make_set(requestUri_s), CallerIPList = make_set(CallerIPAddress),  CallerIPMax= arg_max(CallerIPAddress,*) by ResourceType, ResultType, Resource, id_s, identity_claim_http_schemas_microsoft_com_identity_claims_objectidentifier_g, clientInfo_s
-  | extend timestamp = EndTimeUtc, IPCustomEntity = CallerIPMax, AccountCustomEntity = identity_claim_http_schemas_microsoft_com_identity_claims_objectidentifier_g
+  let AllowedAppId = dynamic(["509e4652-da8d-478d-a730-e9d4a1996ca4","8cae6e77-e04e-42ce-b5cb-50d82bce26b1"]);
+  let OperationList = dynamic(["SecretGet", "KeyGet", "VaultGet"]);
+  let Retrievals = materialize(
+      AzureDiagnostics
+      | where OperationName in (OperationList) and ResourceType =~ "VAULTS"
```
ashwin-patil, line 30: can you please add TimeFilter since we are using materialize dont want to hold all the records in memory.. It was missing in original query too.

Detections/AzureDiagnostics/KeyvaultMassSecretRetrieval.yaml
```diff
@@ -1,58 +1,72 @@
...
+  | extend
+      requestUri_s = column_ifexists("requestUri_s", ""),
+      id_s = column_ifexists("id_s", ""),
+      CallerIPAddress = column_ifexists("CallerIPAddress", ""),
+      clientInfo_s = column_ifexists("clientInfo_s", "")
+  | summarize
+      EventCount = count(),
+      StartTime = min(TimeGenerated),
+      EndTime = max(TimeGenerated),
+      //TimeTriggered = make_list(TimeGenerated, 250),
```
ashwin-patil, line 60: can you reduce soft limit from 250 to 25/50 for better readability in UI and avoid perf issues..

ashwin-patil (changes requested):
Apologies for the delay.. I have added couple of comments,, tested the query and looks good. We will also review other queries you mentioned in the notes for similar updates.  

Thanks for your contribution.

Detections/AzureDiagnostics/KeyvaultMassSecretRetrieval.yaml
ep3p: @ashwin-patil this Scheduled rule has a query frequency and query period of 1d, if I am not mistaken, this query would only have access to data from the last day by the query scheduler, in that case the "TimeFilter" would be already applied, is this correct?

I am going to change the flow of the query a little bit, use the ```as``` operator, and remove ```materialize``` functions. Please, wait a bit.

Detections/AzureDiagnostics/KeyvaultMassSecretRetrieval.yaml
ep3p: Done.

v-spadarthi:
@ep3p : Please resolve the conflicts. Thanks!!

PR #5880, opened 2022-08-10
Update hunt query Possible Ransomware Related Destruction Activity

AaronHoffmannRL:
Change(s):
   - Adds a filter to include WMIC events where shadow copies are deleted

   Reason for Change(s):
   - Updating query for better detections

   Version Updated:
   - N/A

   Testing Completed:
   - Yes

   - Yes


NikTripathi (approved):
@v-spadarthi It looks good. Please go ahead and do the needful. Thanks

shainw (approved):

PR #5941, opened 2022-08-19
Update PotentialKerberoast.yaml

garis:
Removed the filter "EventData has 'krbtgt'", it was probably an error.
   
   Change(s):
   - Updated PotentialKerberoast.yaml
 
   Reason for Change(s):
   - A kind customer noticed that the filter "and EventData has 'krbtgt'" in line 54 doesn't make sense. I believe it was maybe intended as something like "and EventData !has 'krbtgt'" but I removed it because a few lines after we are filtering out krbtgt ServiceNames based on EventData.

   Version Updated:
   - Yes

   Testing Completed:
   - No, I don't think is needed for this change.

### DIFF
--- a/Detections/SecurityEvent/PotentialKerberoast.yaml
+++ b/Detections/SecurityEvent/PotentialKerberoast.yaml
@@ -51,7 +51,7 @@ query: |
   (
   WindowsEvent
   | where TimeGenerated >= ago(starttime)
-  | where EventID == 4769 and EventData has '0x17' and EventData has '0x40810000' and EventData has 'krbtgt'
+  | where EventID == 4769 and EventData has '0x17' and EventData has '0x40810000'
   | extend TicketEncryptionType = tostring(EventData.TicketEncryptionType)
   | where TicketEncryptionType == '0x17'
   | extend TicketOptions = tostring(EventData.TicketOptions)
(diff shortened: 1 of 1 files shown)


v-atulyadav:
Hi @aprakash13 , please review above detection PR and provide your feedback accordingly. Thanks

NikTripathi (approved):
@v-atulyadav It looks good. Please go ahead and do what else is needed.

Detections/SecurityEvent/PotentialKerberoast.yaml
```diff
@@ -51,7 +51,7 @@ query: |
   (
   WindowsEvent
   | where TimeGenerated >= ago(starttime)
-  | where EventID == 4769 and EventData has '0x17' and EventData has '0x40810000' and EventData has 'krbtgt'
+  | where EventID == 4769 and EventData has '0x17' and EventData has '0x40810000'
```
shainw, line 54: We will not remove the matching here. The query is including this and then only excluding if it is the machine account.  The query is much more complex than a simple krbtgt match or not.  If there is a need to modify to a specific environment, then simply modify the local template, the primary template does not need this change.

shainw:
Closing as this change is not accepted.

PR #5900, opened 2022-08-15
Add Checkpoint Firewall and Cisco ASA Parsers

devikamehra:
Change(s):
   - Added 2 new ASIM Parsers

   Reason for Change(s):
   - Increase ASIM Network Session Parser support for 2 more products

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   Checked that the validations are passing and have addressed any issues that are present:
   - Yes

PR #5961, opened 2022-08-22
ASIM Parser for Check Point Firewall

devikamehra:
ASIM Parser for Check Point Firewall

Raising separate PR as told by Ofer for only Check Point ASim Parser (non-parameterized version). More comments are present in #5900. Previous PR will be closed soon.
   
   Change(s):
   - Adding new parser for Check Point Firewall

   Reason for Change(s):
   - Increase support for Network Session 

   Version Updated:
   - Done

   Testing Completed:
   - Yes

   - Yes


Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCheckPointFirewall.yaml
```diff
@@ -0,0 +1,260 @@
...
+  [
+      "Accept","Allow", 
+      "Allow","Allow", 
+      "Drop","Drop", 
+      "Reject","Deny", 
+      "Encrypt","Encrypt", 
+      "Decrypt","Decrypt",
+      "Bypass","Allow", 
+      "Block","Deny", 
+      "Detect","Deny",
```
oshezaf, line 191: I don't know if Detect->Deny. Not sure how to check that. I suspect that all the values below may not belong in a NetworkSession. Did you see them in events?

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCheckPointFirewall.yaml
```diff
@@ -0,0 +1,260 @@
...
+    | parse AdditionalExtensions with * "rule_action=" DeviceAction ";" * "rule_uid=" NetworkRuleName ";" *
+    | parse AdditionalExtensions with * "loguid=" EventOriginalUid ";" * "origin=" Dvc ";" * "originsicname=CN\\=" DvcHostname "," *
+    | parse AdditionalExtensions with * "inzone=" SrcZone ";" * "outzone=" DstZone ";" * 
+    | parse AdditionalExtensions with * "conn_direction=" CommunicationDirection ";" 
+    | parse AdditionalExtensions with * "alert=" AlertName ";" *
+    | parse AdditionalExtensions with * "inspection_category=" InspectionCategory ";" * "inspection_item=" ThreatName ";"                   
+    | extend
+            ThreatCategory = coalesce(AlertName, InspectionCategory),
+            EventResult = case(
+                                isempty(DeviceAction), "",
```
oshezaf, line 216: EventResult cannot be empty. It can be NA. Also, many of the new DeviceActions you added are also Success. The best way is to add a column to the DeviceAction lookup with the matching EventResult.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCheckPointFirewall.yaml
```diff
@@ -0,0 +1,260 @@
...
+    | lookup ProtocolLookup on Protocol
+    | lookup DirectionLookup on CommunicationDirection
+    | lookup ActionLookup on DvcOriginalAction
+    | extend 
+            EventEndTime = EventStartTime,
+            IpAddr = SrcIpAddr,
+            Dst = DstIpAddr,
+            Src = SrcIpAddr,
+            EventMessage = Message,
+            EventSeverity = case(
```
oshezaf, line 246: I would make the default "Informational". Also, the new DvcAction's probably need their own mapping tot "Low" or "Informational". Best to also add this to the lookup.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCheckPointFirewall.yaml
```diff
@@ -0,0 +1,260 @@
...
+                                DvcAction == "Drop", "Low",
+                                DvcAction == "Allow", "Informational",
+                                ""
+                            ),
+            NetworkDirection = case(
+                                isnotempty(NetworkDirection), NetworkDirection,
+                                SrcZone == "Internal" and (DstZone == "Internal" or DstZone == "Local"), "Local",
```
oshezaf, line 253: I think you skipped my comment on DstZone and SrcZone. You should not map them but only use the parsed values ot determine NetworkDirection.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCheckPointFirewall.yaml
```diff
@@ -0,0 +1,260 @@
...
+    | parse AdditionalExtensions with * "inspection_category=" InspectionCategory ";" * "inspection_item=" ThreatName ";"                   
+    | extend
+            ThreatCategory = coalesce(AlertName, InspectionCategory),
+            EventResult = case(
+                                isempty(DeviceAction), "", 
+                                DeviceAction == "Accept", "Success",
+                                "Failure"
+                            ), 
+            NetworkRuleName = coalesce(DeviceCustomString2, NetworkRuleName),
+            EventStartTime = unixtime_milliseconds_todatetime(tolong(ReceiptTime))
```
oshezaf, line 221: As mentioned, if you don't have start in the AdditionalExetnsions, map to TimeGenerated.

oshezaf (commented):
Did you run the tests (schema, data) on the parser? If so, what warnings did you get?

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCheckPointFirewall.yaml
devikamehra: No, I can only see Accept and Deny values. Removing all after Block for now.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCheckPointFirewall.yaml
devikamehra: Lookup update and "NA" is included as well.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCheckPointFirewall.yaml
devikamehra: Added in the same lookup.
Question: In case we have Threat details, should we not have high severity levels like "Medium", "High" and "Critical"?

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCheckPointFirewall.yaml
devikamehra: I commented a question on the previous PR. Sorry for multiple locations, posting it here - 

**Sample values are Internal, External, Local. Should we consider them as Interfaces?** 

If yes, should we include watchlist here as well?

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCheckPointFirewall.yaml
devikamehra: Updated.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCheckPointFirewall.yaml
oshezaf: As documented, the values (inzone, outzone) are not interfaces or zones, but indications whether the source (inzone) or destinaion (outzone) networks are internal or external. As such it should be used only to determine NetworkDirection and not set either zones or interfaces.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCheckPointFirewall.yaml
oshezaf: Make sense, let's do High. Did you actually see Threat details?

devikamehra: I found one type of record where alert=spoofalert value was found.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCheckPointFirewall.yaml
devikamehra: Understood. I found one more column "CommunicationDirection" which can be used to enrich Direction details. Replacing the logic to reduce complexity.

oshezaf: Fine as is.

oshezaf (dismissed):

oshezaf:
Merging with validation error as parse-kv is not supported in validation yet.

PR #5970, opened 2022-08-23
Add Device Inventory Queries

nstylepro:
   
   Change(s):
   - Added new Directory of Device Inventory for M365D

   Reason for Change(s):
   - New Content

   Testing Completed:
   - tested in customer AH
   - didnt run dotnet test

### DIFF
--- a/Hunting Queries/Microsoft 365 Defender/Device Inventory/Anomalous Device Models.yaml
+++ b/Hunting Queries/Microsoft 365 Defender/Device Inventory/Anomalous Device Models.yaml
@@ -6,7 +6,8 @@ requiredDataConnectors:
 - connectorId: MicrosoftThreatProtection
   dataTypes:
   - DeviceInfo
-queryPeriod: 7d
+tactics: []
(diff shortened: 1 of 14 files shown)


aprakash13 (changes requested):
The queries look good. Thank you.
One small suggestion would be to remove **queryPeriod** field from these queries. queryPeriod is relevant more in the detection queries vs the hunting queries.  You can read a little bit about how the query fields are organized and which one is relevant for hunting queries here:  https://github.com/Azure/Azure-Sentinel/wiki/Query-Style-Guide#queryperiod

Please also include **tactics** and **relevantTechniques** field in the queries. In case if these fields are not relevant to these queries you can leave it blank similar to the query here: https://github.com/Azure/Azure-Sentinel/blob/master/Hunting%20Queries/Microsoft%20365%20Defender/TVM/Microsoft%20Defender%20AV%20mode%20device%20count.yaml

aprakash13 (approved):
tHANKS.

PR #5986, opened 2022-08-24
Adding vim parser for Checkpoint

devikamehra:
Change(s):
   - Adding vim parser for Checkpoint

   Reason for Change(s):
   - Increase coverage for ASIM Network Session Parser

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCheckPointFirewall.yaml
```diff
@@ -0,0 +1,329 @@
...
+        CommonSecurityLog
+        | where not(disabled)
+        | where
+                        (isnull(starttime) or TimeGenerated >= starttime) 
+                        and (isnull(endtime) or TimeGenerated <= endtime)
+        | where 
+                        array_length(hostname_has_any) == 0
+        | where DeviceVendor=="Check Point" and DeviceProduct=="VPN-1 & FireWall-1"
+        | where (isnull(dstportnumber) or (DestinationPort == dstportnumber))
+        | where ((eventresult == "*")
```
oshezaf, line 243: We modified the logic for EventResult to use DeviceAction. What I suggest is to do the action lookup in line 256 and check the condition immediately after, instead of this line and the lines in 302/3

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCheckPointFirewall.yaml
```diff
@@ -0,0 +1,329 @@
...
+                    conn_direction:string,
+                    alert:string,
+                    inspection_category:string,
+                    inspection_item:string
+                ) with (pair_delimiter=';', kv_delimiter='=')
+        | extend
+                ThreatCategory = coalesce(alert, inspection_category),
+                NetworkRuleName = coalesce(DeviceCustomString2, rule_uid, Activity),
+                EventStartTime = TimeGenerated,
+                DvcHostname = tostring(extract_all(@"CN\\=(\S+),",originsicname)[0])
```
oshezaf, line 279: Why extract_all and not extract if you have just one item? also, I think that parse will work here.

oshezaf (changes requested):

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCheckPointFirewall.yaml
devikamehra: Updated with DeviceAction

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCheckPointFirewall.yaml
devikamehra: Replaced with parse

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCheckPointFirewall.yaml
```diff
@@ -253,14 +253,16 @@ ParserQuery: |
                             "No match"
                         )
         | where ASimMatchingIpAddr != "No match"
+        | lookup ActionLookup on DeviceAction
+        | where ((array_length(dvcaction) == 0) or DvcAction has_any (dvcaction))
```
oshezaf, line 257: The test for EventResult also has to come here instead of lines 243-245 given success is not just "accept".

oshezaf (changes requested):

oshezaf (approved):

PR #6005, opened 2022-08-26
Adding ASim Parser for Cisco ASA

devikamehra:
Change(s):
   - Adding new Cisco ASA Parser

   Reason for Change(s):
   - Increasing coverage for ASIM Network Schema

   Version Updated:
   - Yes

   Testing Completed:
   - Yes
   
   - Yes


oshezaf:
@devikamehra : the new parsers file extension is yml and not yaml, which prevented ARM template generation.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
```diff
@@ -0,0 +1,369 @@
...
+                        Dvc = Computer
+                | extend
+                        EventSeverity = iff(isempty(DvcAction) or DvcAction == "Allow", "Informational", "Low"),                             
+                        DstPortNumber = coalesce(DstPortNumber, DestinationPort),
+                        SrcPortNumber = coalesce(SrcPortNumber, SourcePort),
+                        DstIpAddr = coalesce(DstIpAddr, DestinationIP),
+                        SrcIpAddr = coalesce(SrcIpAddr, SourceIP),
+                        NetworkRuleName = coalesce(NetworkRuleName, DeviceCustomString2),
+                        NetworkDirection = case(isnotempty(CommunicationDirection), CommunicationDirection,
+                                    (SrcInterfaceName in (internalInterface) and DstInterfaceName in (externalInterface)) or DstInterfaceName in (externalInterface), "Outbound",
```
oshezaf, line 351: The 4th option, both are external interfaces is also valid.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
```diff
@@ -0,0 +1,369 @@
...
+      let all_106023_unparsed = allLogs
+                | where DeviceEventClassID == "106023" and isempty(SourceIP) and not(Message has "protocol 41")
+                | parse Message with * ": " DeviceAction " " Protocol " src " SrcInterfaceName ":" SrcIpAddrAndPort "(" SrcUsername ") dst " DstInterfaceName ":" DstIpAddrAndPort " " NetworkIcmpInfo 'by access-group "' NetworkRuleName '" [' * "]"
+                | parse NetworkIcmpInfo with "(type " NetworkIcmpType ", code " NetworkIcmpCode:int ") "
+                | extend SrcIpAddr = tostring(split(SrcIpAddrAndPort,"/",0)[0]),
+                        SrcPortNumber = toint(split(SrcIpAddrAndPort,"/",1)[0]),
+                        DstIpAddr = tostring(split(DstIpAddrAndPort,"/",0)[0]),
+                        DstPortNumber = toint(split(DstIpAddrAndPort,"/",1)[0])
+                | extend 
+                        SrcUsernameType = iif(isnotempty(SrcUsername),"Windows",""),
```
oshezaf, line 240: Is the username, here and in other events, actually windows format (i.e. domain\user)?

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
```diff
@@ -0,0 +1,369 @@
...
+                        EventMessage = Message,
+                        Dvc = Computer
+                | extend
+                        EventSeverity = iff(isempty(DvcAction) or DvcAction == "Allow", "Informational", "Low"),                             
+                        DstPortNumber = coalesce(DstPortNumber, DestinationPort),
+                        SrcPortNumber = coalesce(SrcPortNumber, SourcePort),
+                        DstIpAddr = coalesce(DstIpAddr, DestinationIP),
+                        SrcIpAddr = coalesce(SrcIpAddr, SourceIP),
+                        NetworkRuleName = coalesce(NetworkRuleName, DeviceCustomString2),
+                        NetworkDirection = case(isnotempty(CommunicationDirection), CommunicationDirection,
```
oshezaf, line 350: I think that CommunicationDirection is 0 or 1 and needs to be translated to the correct values

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
```diff
@@ -0,0 +1,369 @@
...
+                | parse Message with * " packet type " NetworkIcmpType " " * "list " NetworkRuleName " " *;
+      let all_106023 = allLogs
+                | where DeviceEventClassID == "106023" and not(Message has "protocol 41") and isnotempty(SourceIP)
+                | parse Message with * " src "  SrcInterfaceName ":" * " dst " DstInterfaceName ":" * ' by access-group "' NetworkRuleName '" '  *
+                | parse Message with * "(type " NetworkIcmpType ", code " NetworkIcmpCode:int ")" *;
+      let all_106023_unparsed = allLogs
+                | where DeviceEventClassID == "106023" and isempty(SourceIP) and not(Message has "protocol 41")
+                | parse Message with * ": " DeviceAction " " Protocol " src " SrcInterfaceName ":" SrcIpAddrAndPort "(" SrcUsername ") dst " DstInterfaceName ":" DstIpAddrAndPort " " NetworkIcmpInfo 'by access-group "' NetworkRuleName '" [' * "]"
+                | parse NetworkIcmpInfo with "(type " NetworkIcmpType ", code " NetworkIcmpCode:int ") "
+                | extend SrcIpAddr = tostring(split(SrcIpAddrAndPort,"/",0)[0]),
```
oshezaf, line 235: Not critical, but using a temp dynamic for the split, so don' split the same fields twice is more efficient.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
```diff
@@ -0,0 +1,369 @@
...
+                                            DvcAction == "Allow", "Success",
+                                            "Failure")
+                | extend 
+                        Src = SrcIpAddr,
+                        Dst = DstIpAddr,
+                        Duration = NetworkDuration,
+                        IpAddr = SrcIpAddr,
+                        Rule = NetworkRuleName,
+                        User = DstUsername
+                | project-away CommunicationDirection, Destination*, Device*, LogSeverity, Protocol, SimplifiedDeviceAction, Source*
```
oshezaf, line 367: 1. The list of fields we project away seems to be shorter than the Checkpoint parser. Just making sure we did no forget any.
2. While i does not look "nice", project-away at each union branch everything you can. It significantly reduces the union memory cost. If you need a specific Device field for example, rename it in the branch so you can projec-away Device*

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
```diff
@@ -0,0 +1,369 @@
...
+                        DstUsernameType = iif(isnotempty(DstUsername),"Windows",""),
+                        SimplifiedDeviceAction = DeviceAction,
+                        SessionId = NetworkSessionId;
+      let all_302014_302016 = allLogs
+                | where DeviceEventClassID in ("302014", "302016")
+                | parse kind=regex Message with * ": " DeviceAction " " Protocol " connection " NetworkSessionId " for " SrcInterfaceName ":" SrcIpAddr "/" SrcPortNumber:int @"[(]?" SrcUsername @"[)]? to " DstInterfaceName ":" DstIpAddr "/" DstPortNumber:int @"[(]?" DstUsername1 @"[)]? [Dd]uration " NetworkDuration " bytes " DstBytes:long @"\s?" Remaining 
+                | extend reasonParsed = extract_all(@"$|\((\S+)\)|([\S+\s*]+)\sfrom\s\S+\s(\(\S+\)|\S+)$|([\S+\s*]+)\sfrom\s\S+$|([\S+\s*]+)\s\((\S+)\)$|([\S+\s*]+)$",Remaining)[0]
+                | extend 
+                            SrcUsername = trim(@"[()]", SrcUsername),
+                            DstUsername = trim(@"[()]",coalesce(reasonParsed[0], reasonParsed[2], reasonParsed[5], DstUsername1)),
```
oshezaf, line 283: This sounds really error prone. Can he DstUsername really be in 3 different places in the "Remining"? Same for Reason. Also, the parse will fail if DstUsername1 is missing, want it. And if DstUsername1, what's the importance of the other fields?

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
```diff
@@ -0,0 +1,369 @@
...
+                                                    NetworkDirection == "outbound", "Outbound",
+                                                    "")
+                | extend 
+                        SrcUsernameType = iif(isnotempty(SrcUsername),"Windows",""),
+                        DstUsernameType = iif(isnotempty(DstUsername),"Windows",""),
+                        SimplifiedDeviceAction = DeviceAction,
+                        SessionId = NetworkSessionId;
+      let all_302014_302016 = allLogs
+                | where DeviceEventClassID in ("302014", "302016")
+                | parse kind=regex Message with * ": " DeviceAction " " Protocol " connection " NetworkSessionId " for " SrcInterfaceName ":" SrcIpAddr "/" SrcPortNumber:int @"[(]?" SrcUsername @"[)]? to " DstInterfaceName ":" DstIpAddr "/" DstPortNumber:int @"[(]?" DstUsername1 @"[)]? [Dd]uration " NetworkDuration " bytes " DstBytes:long @"\s?" Remaining
```
oshezaf, line 279: Few regex comments:
- @"[(]?" -> no need for a character group here, you really want \(?.
- Don't change now, but for next time:
  - you can break the parse to the part that does not need kind-regex and do just the rest with a second parse with kind=regex.
  - The parse statement can really be without kind=regex: 
    - trim out the () from user names after the parse. 
    - remove the extra spaces from Remining in the extract_all statement
    - replace the [dD] with a *. I am not sure it would work, if not, the only option is to replace duration with Duration in a previous extend.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
```diff
@@ -0,0 +1,369 @@
...
+                | extend SimplifiedDeviceAction = case(DeviceAction == "denied", "Deny",
+                                                        "Allow");
+      let remainingLogs = allLogs
+                | where DeviceEventClassID in ("106002", "106012", "106013", "106020");
+      let networkaddressWatchlistData = materialize (_ASIM_GetWatchlistRaw("NetworkAddresses"));
+      let internalInterface = networkaddressWatchlistData | where WatchlistItem.Tags has "Internal" | distinct tostring(WatchlistItem["Range Name"]);
+      let externalInterface = networkaddressWatchlistData | where WatchlistItem.Tags has "External" | distinct tostring(WatchlistItem["Range Name"]);
+      union isfuzzy=true all_106001_alike, all_106010_alike, all_106018, all_106023, all_106023_unparsed, all_106023_41,  all_302013_302015_unparsed, all_302013_302015_parsed, all_302014_302016, all_302020_302021, all_7_series, all_106007, all_106017, all_106100_parsed, all_106100_unparsed, remainingLogs
```
oshezaf, line 321: isfuzzy=false. No reason any will fail.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
```diff
@@ -0,0 +1,369 @@
...
+                | extend 
+                        EventStartTime = TimeGenerated,
+                        EventEndTime = TimeGenerated,
+                        EventVendor = "Cisco",
+                        EventProduct = "ASA",
+                        EventCount = coalesce(EventCount,toint(1)),
+                        EventType = "NetworkSession",
+                        EventSchema = "NetworkSession",
+                        EventSchemaVersion = "0.2.4",
+                        DvcAction = case(trim(@"\s",SimplifiedDeviceAction) in ("Teardown","Built"), "",
```
oshezaf, line 331: SimplifiedDeviceAction has_any is much better performance wise

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
```diff
@@ -0,0 +1,369 @@
...
+                | project-rename 
+                        EventProductVersion = DeviceVersion,
+                        EventOriginalType = DeviceEventClassID,
+                        EventOriginalSeverity = OriginalLogSeverity,
+                        DvcOriginalAction = DeviceAction,
+                        EventMessage = Message,
+                        Dvc = Computer
+                | extend
+                        EventSeverity = iff(isempty(DvcAction) or DvcAction == "Allow", "Informational", "Low"),                             
+                        DstPortNumber = coalesce(DstPortNumber, DestinationPort),
```
oshezaf, line 345: For performance reasons, it is better to avoid those and include the choice of fields in the union branches. This prevents the branches in which he original CEF field is fine from becoming calculated fields.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
oshezaf, line 331: Are you sure we normalize the value of DvcAction in all branches? in most cases we take the original value which may not be an allowed. We need mapping as we did just for 302014/6

oshezaf (changes requested):

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
devikamehra: @oshezaf What should be the Network Direction in that case?

oshezaf: Well, it is in the [doc](https://docs.microsoft.com/en-us/azure/sentinel/network-normalization-schema): External

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
devikamehra: Some examples of Username values are LOCAL\U1613@ABC.COM, U1613@ABC.COM, U1613, 512:XYZ_U1632_Group1

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
devikamehra: Updated

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
```diff
@@ -0,0 +1,369 @@
...
+                | lookup ProtocolLookup on Protocol
+                | project-rename 
+                        EventProductVersion = DeviceVersion,
+                        EventOriginalType = DeviceEventClassID,
+                        EventOriginalSeverity = OriginalLogSeverity,
+                        DvcOriginalAction = DeviceAction,
+                        EventMessage = Message,
+                        Dvc = Computer
+                | extend
+                        EventSeverity = iff(isempty(DvcAction) or DvcAction == "Allow", "Informational", "Low"),
```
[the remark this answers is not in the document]
devikamehra: Updated

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
```diff
@@ -0,0 +1,369 @@
+Parser:
```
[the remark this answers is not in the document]
devikamehra: Replaced with correct extension.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
devikamehra: 1. The list is smaller here because I have used project-away in allLogs itself.
2. I have used this technique to project-away at the root query itself. Also, any temp variables created in individual branch are also project-away.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
oshezaf: So, identifying type is harder and some values don't really follow types I know (the 512: one and the one with both @ and \):
I would do with "@" as UPN, with "\" (if there is no "@") as Windows, and Simple otherwise (U1613).

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
devikamehra: Yes, following are the different ways this Remaining string can shape:
1. <empty>
2. (domain\USER001) --> reasonParsed[0]
3. TCP FINs from OUTSIDE (domain\USER001) --> reasonParsed[1] from OUTSIDE reasonParsed[2]
4. TCP FINs from OUTSIDE --> reasonParsed[3] from OUTSIDE
5. TCP FINs (domain\USER001) --> reasonParsed[4] reasonParsed[5]

No, if DstUsername1 is not present, the remaining parsing will work fine as it is recorded as empty string.
image

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
devikamehra: Resolved point 1, noted the remaining.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
devikamehra: Make sense

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
devikamehra: Updated

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
devikamehra: Updated

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
devikamehra: I have mapped the value wherever required. All values of DvcAction are standard.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
devikamehra: This is an error with ASA parsing from data collection team. Data is provided as "inbound"/"outbound" from Cisco ASA raw data which is directly getting picked into CommunicationDirection field.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yaml
```diff
@@ -0,0 +1,375 @@
...
+          , "142","ROHC"
+          , "143","Ethernet"
+          , "253",""
+          , "254",""
+          , "255","Reserved"];
+    let NWParser=(disabled:bool=false){
+      let allLogs = CommonSecurityLog
+      | where not(disabled)
+      | where DeviceVendor == "Cisco" and DeviceProduct == "ASA"
+      | project-rename AliasDeviceEventClassID = DeviceEventClassID,
```
oshezaf, line 217: I hope this is fine. As a rule, nothing should happen before the branch _where_ clause, and this contradicts this rule. When I asked to include in each branch, I did mean each branch. In practice, _project-rename_ may be optimized and it may not make a difference. It is better to make an invoke function from that and call it from each relevant branch after the where clause.

oshezaf (changes requested):
Next round :-)

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
devikamehra: Updated

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yml
devikamehra: Updated

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yaml
```diff
@@ -0,0 +1,375 @@
...
+                                                    "")
+                | extend 
+                        SrcUsernameType = iif(isnotempty(SrcUsername),"Windows",""),
+                        DstUsernameType = iif(isnotempty(DstUsername),"Windows",""),
+                        SimplifiedDeviceAction = DeviceAction,
+                        SessionId = NetworkSessionId;
+      let all_302014_302016 = allLogs
+                | where AliasDeviceEventClassID in ("302014", "302016")
+                | parse kind=regex Message with * ": " DeviceAction " " Protocol " connection " NetworkSessionId " for " SrcInterfaceName ":" SrcIpAddr "/" SrcPortNumber:int @"\(?" SrcUsername @"\)? to " DstInterfaceName ":" DstIpAddr "/" DstPortNumber:int @"\(?" DstUsername1 @"\)? [Dd]uration " NetworkDuration " bytes " DstBytes:long @"\s?" Remaining 
+                | extend reasonParsed = extract_all(@"$|\((\S+)\)|([\S+\s*]+)\sfrom\s\S+\s(\(\S+\)|\S+)$|([\S+\s*]+)\sfrom\s\S+$|([\S+\s*]+)\s\((\S+)\)$|([\S+\s*]+)$",Remaining)[0]
```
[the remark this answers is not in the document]
devikamehra: I have taken a stab at this again. Hope this one helps us improve the efficiency and accuracy.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yaml
devikamehra: I have made the following improvements:
1. Filtered the data required initially itself.
2. Divided the data into parsed and unparsed to improve performance. Here is where all the required fields are projected away or renamed. As per the rule, this is after the filter operation done at step one.
3. In each branch, I used parsed/unparsed respectively. In case any temporary columns are created, the same are projected away in the branch itself.

Do we need the invoke function now?

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yaml
```diff
@@ -0,0 +1,423 @@
...
+                        EventProductVersion = AliasDeviceVersion,
+                        EventOriginalType = AliasDeviceEventClassID,
+                        EventOriginalSeverity = OriginalLogSeverity,
+                        DvcOriginalAction = DeviceAction,
+                        EventMessage = Message,
+                        Dvc = Computer
+                | extend
+                        EventSeverity = iff(isempty(EventResult) or EventResult == "Success", "Informational", "Low"),
+                        NetworkDirection = case(isnotempty(CommunicationDirection), CommunicationDirection,
+                                    (SrcInterfaceName in (internalInterface) and DstInterfaceName in (externalInterface)) or DstInterfaceName in (externalInterface), "Outbound",
```
oshezaf, line 407: I may be missing something, but this logical expression seems to be superfluous. If _DstInterfaceName in (externalInterface)_ is true, _SrcInterfaceName in (internalInterface)_ would not matter. Why do we need the left side of the _or_ operator? Same for the next line.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yaml
oshezaf: It would still be more optimal, but I am not sure how much. Let's leave it at that.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yaml
```diff
@@ -0,0 +1,423 @@
...
+                | parse RemainingString with ReasonString "(" DstUsernameSimple ")"
+                | extend ReasonString = case(RemainingString has "from" and RemainingString !has "(", RemainingString,
+                                             ReasonString)
+                | parse ReasonString with Reason " from " *
+                | extend Reason = case(isempty(Reason), ReasonString,
+                                       Reason)
+                | lookup EventResultMapping on Reason
+                | extend 
+                            SrcUsername = trim(@"\s?\(?\)?\s?", SrcUsername),
+                            DstUsername = case(isempty(DstUsername),DstUsernameSimple,
```
oshezaf, line 303: I think that there is a missing element in this case statement.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yaml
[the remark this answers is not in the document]
oshezaf: Yup. While not simple, It looks less sentivie to changes and easier to maintain. In general (no need to change), it makes sense to add comments to your code.

oshezaf (changes requested):
Two small things, which I think might not be correct.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoASA.yaml
devikamehra: There are two types of events here:
1. "%ASA-6-302014:  Teardown TCP connection 261246338 for OUTSIDE:89.160.20.156/50120(LOCAL\\domain\\USER001) to OUTSIDE:40.0.0.1/443 duration 0:02:05 bytes 9610 TCP FINs (domain\\USER001)"
2. "%ASA-6-302014:  Teardown TCP connection 261246338 for OUTSIDE:89.160.20.156/50120(LOCAL\\domain\\USER001) to OUTSIDE:40.0.0.1/443(LOCAL\\domain\\USER002) duration 0:02:05 bytes 9610 TCP FINs (domain\\USER002)"

In case there is no Destination Username after DstPortNumber (DstUsername in code), we take information from the user details (DstUsernameSimple in code) present at the end. Otherwise, we capture only the earlier one.

oshezaf (approved):

PR #6058, opened 2022-09-01
Adding vim Parser for Cisco ASA

devikamehra:
Change(s):
   - Added vim Parser for Cisco ASA

   Reason for Change(s):
   - Increase the coverage for ASimNetworkSession Parser

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoASA.yaml
```diff
@@ -0,0 +1,576 @@
...
+      let all_106023 = parsedData
+                | where AliasDeviceEventClassID == "106023" and not(Message has "protocol 41")
+                | parse Message with * " src "  SrcInterfaceName ":" * " dst " DstInterfaceName ":" * ' by access-group "' NetworkRuleName '" '  *
+                | parse Message with * "(type " NetworkIcmpType ", code " NetworkIcmpCode:int ")" *;
+      let all_106023_unparsed = unparsedData
+                | where AliasDeviceEventClassID == "106023" and not(Message has "protocol 41")
+                | parse Message with * ":" DeviceAction " " Protocol " src " SrcInterfaceName ":" SrcIpAddrAndPort "(" SrcUsername ") dst " DstInterfaceName ":" DstIpAddrAndPort " " NetworkIcmpInfo 'by access-group "' NetworkRuleName '" [' * "]"
+                | parse NetworkIcmpInfo with "(type " NetworkIcmpType ", code " NetworkIcmpCode:int ") "
+                | extend SrcIpAddrAndPort = split(SrcIpAddrAndPort,"/"), DstIpAddrAndPort = split(DstIpAddrAndPort,"/")
+                | extend SrcIpAddr = tostring(SrcIpAddrAndPort[0]),
```
oshezaf, line 307: Missed asking on the ASIM version: why don't IP addresses and ports work int he "parse" statement itself? Is the port optional?

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoASA.yaml
```diff
@@ -0,0 +1,576 @@
...
+            eventresult:string='*', 
+            disabled:bool=false)
+        { 
+      let src_or_any = set_union(srcipaddr_has_any_prefix, ipaddr_has_any_prefix); 
+      let dst_or_any = set_union(dstipaddr_has_any_prefix, ipaddr_has_any_prefix);
+      let allLogs = CommonSecurityLog
+                | where not(disabled)
+                | where (isnull(starttime) or TimeGenerated >= starttime) and (isnull(endtime) or TimeGenerated <= endtime)
+                | where DeviceVendor == "Cisco" and DeviceProduct == "ASA"
+                | where (isnull(dstportnumber) or (DestinationPort == dstportnumber)) or Message has tostring(dstportnumber)
```
oshezaf, line 258: If you move this line to the parsed/unparsed branches, you can avoid the unncessary compare. So parsed matches only DestinationPort and unparsed only message.

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoASA.yaml
```diff
@@ -0,0 +1,576 @@
...
+                | extend temp_isSrcMatch=has_any_ipv4_prefix(SourceIP,src_or_any),
```
oshezaf, line 259: Since this filtering is relevant only to parsed branches, it would fail (i.e. return no results) on unparsed events. The section should be moved to the "parsed" branch.

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoASA.yaml
```diff
@@ -0,0 +1,576 @@
...
+                                                temp_isSrcMatch, "SrcIpAddr",
+                                                temp_isDstMatch, "DstIpAddr",
+                                                "No match"
+                                            )
+                | where ASimMatchingIpAddr != "No match"
+                | where ((array_length(dvcaction) == 0) or SimplifiedDeviceAction has_any (dvcaction))
+                | where ((eventresult == "*") or 
+                        (eventresult == "Success" and SimplifiedDeviceAction == "Allow") or 
+                        (eventresult == "Failure" and SimplifiedDeviceAction != "Allow"))
+                | where DeviceEventClassID in ("106001","106006","106015","106016","106021","106022","106010","106014","106018","106023","302013","302015","302014","302016","302020","302021","710002","710003","710004","710005","106007","106017","106100","106002","106012","106013","106020");
```
oshezaf, line 273: This should come after line 257

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoASA.yaml
```diff
@@ -0,0 +1,576 @@
...
+                                temp_isDstMatch=has_any_ipv4_prefix(DestinationIP,dst_or_any)
+                | extend ASimMatchingIpAddr = case(
+                                                array_length(src_or_any) == 0 and array_length(dst_or_any) == 0, "-", // match not requested
+                                                (temp_isSrcMatch and temp_isDstMatch), "Both", // has to be checked before the individual 
+                                                temp_isSrcMatch, "SrcIpAddr",
+                                                temp_isDstMatch, "DstIpAddr",
+                                                "No match"
+                                            )
+                | where ASimMatchingIpAddr != "No match"
+                | where ((array_length(dvcaction) == 0) or SimplifiedDeviceAction has_any (dvcaction))
```
oshezaf, line 269: At least for 302014, this condition is incorrect. Is it correct for all others? We need to find a way to do it only when relevant. It is not an easy one and requires some thinking about overall structure.

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoASA.yaml
```diff
@@ -0,0 +1,576 @@
...
+                | where ((eventresult == "*") or 
+                        (eventresult == "Success" and SimplifiedDeviceAction == "Allow") or 
+                        (eventresult == "Failure" and SimplifiedDeviceAction != "Allow"))
+                | project-away SrcIpAddrAndPort, DstIpAddrAndPort, NetworkIcmpInfo;
+      let all_106023_41 = union parsedData, unparsedData
```
oshezaf, line 332: Just noticed the parsed/unparsed union: you might have got it right in the ASIM parser - though it certainly duplicates work - but it is still very hard to maintain this "reverse-fork" in the parser. Given the challenges of the vim parsers, this would just not work.

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoASA.yaml
```diff
@@ -0,0 +1,576 @@
...
+                | project-rename AliasDeviceEventClassID = DeviceEventClassID,
+                                 AliasDeviceVersion = DeviceVersion,
+                                 AliasDeviceAction = DeviceAction,
+                                 NetworkRuleName = DeviceCustomString2,
+                                 SrcIpAddr = SourceIP,
+                                 SrcPortNumber = SourcePort,
+                                 DstIpAddr = DestinationIP, 
+                                 DstPortNumber = DestinationPort
+                | project-away Activity, AdditionalExtensions, ApplicationProtocol, Device*, Destination*, EndTime, ExternalID, File*, Flex*, IndicatorThreatType, Malicious*, Old*, Process*, ReceiptTime, ReceivedBytes, Remote*, Report*, Request*, Source*, SentBytes, StartTime, TenantId, Threat*;
+      let unparsedData = allLogs
```
oshezaf, line 285: You need to add prefiltering on the Message field for all unparsed branches. Currently you filter only after parsing.

My suggestion is to include the parsing of IP/Ports in the unparsed branch and pre and post filter in the unparsed branch. There are two branches in which the parsing may be different (where you add an optional user after). If needed created a 3rd root (parsed/unparsed/special_unparsed or whatever).

oshezaf (changes requested):
Don't forget to update vimNetworkSession

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoASA.yaml
devikamehra: Yes Ofer, you are right. We have port value as optional.

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoASA.yaml
devikamehra: Corrected

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoASA.yaml
devikamehra: Done

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoASA.yaml
devikamehra: Moved

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoASA.yaml
devikamehra: Updated

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoASA.yaml
devikamehra: Only unparsed is required. Updated.

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoASA.yaml
devikamehra: Pre and post filters added.

NikTripathi:
Hi @oshezaf @devikamehra Are we good to go for this?

oshezaf (changes requested):
Sorry for the delay. Because I moved to e-mail I never finished my review here and therefore the PR was still open on my side. For better or worse, the last remaining point is a bug :-( Sorry for the hassle.

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoASA.yaml
```diff
@@ -284,6 +284,13 @@ ParserQuery: |
                 | project-away Activity, AdditionalExtensions, ApplicationProtocol, Device*, Destination*, EndTime, ExternalID, File*, Flex*, IndicatorThreatType, Malicious*, Old*, Process*, ReceiptTime, ReceivedBytes, Remote*, Report*, Request*, Source*, SentBytes, StartTime, TenantId, Threat*;
       let unparsedData = allLogs
                 | where isempty(SourceIP)
+                | where Message has tostring(dstportnumber)
+                        and ((array_length(src_or_any) == 0 or Message has_any(src_or_any))
```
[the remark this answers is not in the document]
devikamehra: Fixed the bug for CiscoASA and would verify for Checkpoint as well.
Kept source and destination separate only because we need it for setting "ASimMatchingIpAddr" parameter.

oshezaf (approved):

PR #6181, opened 2022-09-19
GitHub TemplateSpec Solution

v-atulyadav:
Change(s):

This solution is packaged by making "Is1PConnector: false" as if we make this value false the CCP connector is not visible.

See guidance below
Reason for Change(s):
This PR is part of template spec solution migration for solution GitHub.

See guidance below
Version Updated:
2.0.2

See guidance below
This is an 1P solution.


v-sabiraj (approved):
Looks good

PR #6184, opened 2022-09-20
Updated ASIM Parser for Infoblox NIOS. Replaced usage of Regex function with Parse Operators.

vakohl:
…ons with Parse operators

   Change(s):
   - Updated Parser. Replace use of Regex with Parse operator.

   Reason for Change(s):
   - for parser enhancement.

   Version Updated:
   - N/A

   Testing Completed:
   - Yes

   - Yes

### DIFF
--- a/Parsers/ASimDns/Parsers/ASimDnsInfobloxNIOS.yaml
+++ b/Parsers/ASimDns/Parsers/ASimDnsInfobloxNIOS.yaml
@@ -26,49 +26,47 @@ ParserQuery: |
     Syslog
     | where not(disabled)
     | where ProcessName == "named" and SyslogMessage has_all ("client", "query:", "response:")
-    | parse SyslogMessage with
-        "client "
-        SrcIpAddr:string "#" 
-        SrcPortNumber:int " "
-        NetworkProtocol:string ": query: "
-        DnsQuery:string " "
-        DnsQueryClassName:string " "
-        DnsQueryTypeName:string " "
-        "response: "
-        rest_of_string:string
-    | extend
-        // -- multi-line mode (?m) needed in regex since samples included new lines. Might not exist in real logs
-        extract = extract_all(@"^(\w+) ([\w\+]+?)(?: |$)(?m)(.*?$)?", rest_of_string)[0]
-    | extend 
-        DnsResponseCodeName = tostring(extract[0]),
-        DnsFlags = tostring(extract[1]),
-        DnsResponseName = tostring(extract[2])
-    | extend 
-        EventSubType = "response"
-    | project-away extract, rest_of_string,SyslogMessage, ProcessName, ProcessID, Facility, SeverityLevel, HostName
+    | parse SyslogMessage with *
+        "client " SrcIpAddr: string
+        "#" SrcPortNumber: string
+        " " NetworkProtocol: string
+        ": query: " DnsQuery: string
+        " " DnsQueryClassName: string
+        " " DnsQueryTypeName: string
+        " response: " DnsResponseCodeName: string
+        " " DnsFlags: string
+    | extend DNSResourceRecordIndex= indexof(DnsFlags, " ")
+    | extend DNSResourceRecord =iif(DNSResourceRecordIndex != "-1", substring(DnsFlags, DNSResourceRecordIndex), "")
+    | extend DnsFlags =iif(DNSResourceRecordIndex != "-1", substring(DnsFlags, 0, DNSResourceRecordIndex), DnsFlags)
+    | extend EventSubType = "response"
+    | project-away DNSResourceRecordIndex,SyslogMessage, ProcessName, ProcessID, Facility, SeverityLevel, HostName
   };
   let request = (disabled: boolean=false) {
     Syslog 
     | where not(disabled)
     | where ProcessName == "named" and SyslogMessage has_all ("client", "query:") and SyslogMessage !has "response:"
-    // -- trim_start needed to remove optional @xxxx identifier.
-    | extend SyslogMessage = trim_start (@'^(?:[\w\: ]+)?client (?:@.*? )?', SyslogMessage)
-    // -- trim_end needed to remove optional repetative IP address at the end.
-    | extend SyslogMessage = trim_end (@' \([\d\.]+\)', SyslogMessage)
-    | extend SyslogMessage = replace_string (SyslogMessage, '\\ ', '@@@')
-    | parse SyslogMessage with
-        SrcIpAddr:string "#" 
-        SrcPortNumber:int *
-        "query: "
-        DnsQuery:string " "
-        DnsQueryClassName:string " "
-        DnsQueryTypeName:string " "
-        DnsFlags:string
+    | parse SyslogMessage with *
+        "client " restData: string
+    | extend restData = iif(restData startswith "@", (substring(restData, indexof(restData, " "))), restData)
+    | extend restData = replace_string(restData,"\\ ","@@@")
+    | parse restData with 
+        SrcIpAddr: string
+        "#" SrcPortNumber: string
+        " " *
+        "query: " DnsQuery: string
+        " " DnsQueryClassName: string
+        " " DnsQueryTypeName: string
+        " " DnsFlags: string
     | extend DnsQuery = replace_string (DnsQuery, '@@@', ' ')
+    | extend ServerIPAddressIndex= indexof(DnsFlags, " ")
+    | extend ServerIPAddress = iif(ServerIPAddressIndex != "-1", substring(DnsFlags, ServerIPAddressIndex), "")
+    | extend ServerIPAddress = replace_regex(ServerIPAddress,@"[()]","")
+    | extend DnsFlags =iif(ServerIPAddressIndex != "-1", substring(DnsFlags, 0, ServerIPAddressIndex), DnsFlags)
+    | extend SrcPortNumber = replace_regex(SrcPortNumber,@"[^\d]","")
     | extend 
         EventSubType = "request",
         DnsResponseCodeName = "NA"
-    | project-away SyslogMessage, ProcessName, ProcessID, Facility, SeverityLevel, HostName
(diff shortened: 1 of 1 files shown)


Parsers/ASimDns/Parsers/ASimDnsInfobloxNIOS.yaml
```diff
@@ -26,49 +26,47 @@ ParserQuery: |
...
+        SrcIpAddr: string
+        "#" SrcPortNumber: string
+        " " *
+        "query: " DnsQuery: string
+        " " DnsQueryClassName: string
+        " " DnsQueryTypeName: string
+        " " DnsFlags: string
     | extend DnsQuery = replace_string (DnsQuery, '@@@', ' ')
+    | extend ServerIPAddressIndex= indexof(DnsFlags, " ")
+    | extend ServerIPAddress = iif(ServerIPAddressIndex != "-1", substring(DnsFlags, ServerIPAddressIndex), "")
```
oshezaf, line 62: Few comments on the splitting of DnsFlags:
- The substring expressions keep an unneeded space at the beginning of the second field (as noted above).
- While this can be solved with fixing the indexes, there are two preferable options here:
  - Since you anyway use regex, extract_all is probably more efficient and easier to understand.
  - Use split to get the values, with the current replace_regex.

Parsers/ASimDns/Parsers/ASimDnsInfobloxNIOS.yaml
```diff
@@ -26,49 +26,47 @@ ParserQuery: |
...
+    | extend ServerIPAddress = replace_regex(ServerIPAddress,@"[()]","")
+    | extend DnsFlags =iif(ServerIPAddressIndex != "-1", substring(DnsFlags, 0, ServerIPAddressIndex), DnsFlags)
+    | extend SrcPortNumber = replace_regex(SrcPortNumber,@"[^\d]","")
```
oshezaf, line 65: Not sure I understand this line. First SrcPortNumber has to be an integer, which is how it is extracted in the original. and then, why would you remove the first character?

Parsers/ASimDns/Parsers/ASimDnsInfobloxNIOS.yaml
```diff
@@ -26,49 +26,47 @@ ParserQuery: |
...
+    | parse SyslogMessage with *
+        "client " SrcIpAddr: string
+        "#" SrcPortNumber: string
+        " " NetworkProtocol: string
+        ": query: " DnsQuery: string
+        " " DnsQueryClassName: string
+        " " DnsQueryTypeName: string
+        " response: " DnsResponseCodeName: string
+        " " DnsFlags: string
+    | extend DNSResourceRecordIndex= indexof(DnsFlags, " ")
```
oshezaf, line 38: The substring expressions keep an unneeded space at the beginning of the second field. This can be solved with fixing the indexes.

Parsers/ASimDns/Parsers/ASimDnsInfobloxNIOS.yaml
```diff
@@ -26,49 +26,47 @@ ParserQuery: |
...
+    | extend DNSResourceRecord =iif(DNSResourceRecordIndex != "-1", substring(DnsFlags, DNSResourceRecordIndex), "")
```
oshezaf, line 39: The field name is DnsResponseName. Best to also use it your parser.

oshezaf (changes requested):
1. You need to apply the same changes to the vim parser.
2. When asking to test with all data, I meant using the [ASIM testers](https://learn.microsoft.com/en-us/azure/sentinel/normalization-develop-parsers#test-parsers). You would have identified some of the issues I identified this way (the field names at least). In general, you can benefit from running the parsers on any "semi-normalized" parser like your original.
3. I need to check if you need to update the version number of not. Will get back to you later today on that.
4. Process comment: because you changed the line break in the large parse statements, the diff for reviewers is less useful. It is better to avoid unneeded larger changes to make reviewing easier.
5. More comments for specific lines.

Parsers/ASimDns/Parsers/ASimDnsInfobloxNIOS.yaml
```diff
@@ -26,49 +26,47 @@ ParserQuery: |
...
-        SrcIpAddr:string "#" 
-        SrcPortNumber:int *
-        "query: "
-        DnsQuery:string " "
-        DnsQueryClassName:string " "
-        DnsQueryTypeName:string " "
-        DnsFlags:string
+    | parse SyslogMessage with *
+        "client " restData: string
+    | extend restData = iif(restData startswith "@", (substring(restData, indexof(restData, " "))), restData)
```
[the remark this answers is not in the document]
vakohl: Changed logic to use 'Split' instead of 'parse' here to avoid using 'restData'. Now only using SyslogMessage:

| extend SyslogMessage_s = (split(SyslogMessage_s,"client "))[1]

Parsers/ASimDns/Parsers/ASimDnsInfobloxNIOS.yaml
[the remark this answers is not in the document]
vakohl: Changed column name from 'ServerIPAddress' to 'DvcIpAddr'. Will copy in original parser as well. Now tested schema with ASIM testers.

Parsers/ASimDns/Parsers/ASimDnsInfobloxNIOS.yaml
vakohl: Changed column name from 'DNSResourceRecord' to 'DnsResponseName'.

Parsers/ASimDns/Parsers/ASimDnsInfobloxNIOS.yaml
vakohl: Updated logic to remove extra space using:

substring(DnsFlags, DnsResponseNameIndex+1), "")

Parsers/ASimDns/Parsers/ASimDnsInfobloxNIOS.yaml
vakohl: In few logs, I found 'SrcPortNumber' ending with a ':' This was breaking the 'parse' logic as ':' was not getting saved in an 'int' type. sample logs:

1. query: info: client 1.2.3.4#60010: view standard: query: blip.prefetch.net IN AAAA +
2. client 1.2.3.4#42872 (prefetch.net): query: prefetch.net IN ANY + (1.2.3.4)

Previously I was using regex to remove ':' in 'SrcPortNumber'. Now I changed the logic to use replace_string and then convert to 'int' type. This would take care of intermittent ":" beside port numbers.

[ extend SrcPortNumber = toint(replace_string(SrcPortNumber,":","")) ]

Parsers/ASimDns/Parsers/ASimDnsInfobloxNIOS.yaml
vakohl: Updated logic to remove extra space: iif(DvcIpAddrIndex != "-1", substring(DnsFlags, DvcIpAddrIndex+1), "")
I was using regex to only remove the parenthesis around 'DvcIpAddr' in some logs using replace_regex(ServerIPAddress,@"[()]",""). I changed it and now using trim_start and trim_end to remove the parenthesis. Do you think its better now or we should be using the original extract_all logic?

    | extend DvcIpAddr = trim_start("[(]",DvcIpAddr)
    | extend DvcIpAddr = trim_end("[)]",DvcIpAddr)

Also, I now noticed 'HostIP' value is same as above calculated 'DvcIpAddr' value. So, I guess we don't really need above logic right? We already have below logic in the parser:

DvcIpAddr=iff (HostIP_s == "Unknown IP", "", HostIP_s)

oshezaf: 1. trim functions are regex based, so a single extract_all will be more efficient and, I think, comprehendible. 
2. I think this is the reason I did not parse DvcIpAddr from the message int he first place :-). Yup. Not needed.

Parsers/ASimDns/Parsers/ASimDnsInfobloxNIOS.yaml
oshezaf: wouldn't using 

  **SrcPortNumber:int * " "** 

solve the issue?

int would ensure that the colon is not captures, and the * will capture it.

Parsers/ASimDns/Parsers/ASimDnsInfobloxNIOS.yaml
vakohl: Thanks Ofer. Removed logic where I was calculating 'DvcIpAddr'. Below line should be good for storing DnsFlags and dumping the later texts:

| extend DnsFlags= tostring((split(DnsFlags," "))[0])

for handling both scenarios:
1. client @0x7f1a891bbbb 192.46.117.8#28492 (xxxx.uksouth.cloudapp.azure.com): view Recursion View: query: xxxx.uksouth.cloudapp.azure.com IN A +E(0)DV (192.46.116.155)
2. query: info: client 1.2.3.4#7619: query: 10.10.10.10.in-addr.arpa IN PTR -E

Parsers/ASimDns/Parsers/ASimDnsInfobloxNIOS.yaml
vakohl: It works. Thanks Ofer. Now using:
"#" SrcPortNumber: int *

oshezaf (changes requested):
Thanks. So, just need to update the version to 0.5.1. I assume vim is next?

oshezaf (approved):

v-sabiraj:
Hey @vakohl, the PR is already approved, can you please resolve the conflicts, thanks.

NikTripathi (approved):

PR #6304, opened 2022-10-04
IP to TI map insight updated

shainw:
Change(s):
   - Changing TI map to generic map if IP is passed in, moved old TI map logic to new insight called IP_RemoteConnections.yaml

   Reason for Change(s):
   - Change is to support Entity page search for IP only versus Investigation page mapping to specific data types.
   - Moved existing logic that maps to specific datatypes to a named Insight and updated description for clarity
   - Customer requested feature

   Version Updated:
   - N/A

   Testing Completed:
   - Yes

   - Yes


aprakash13 (approved):
Thanks @shainw for all the work on these.

PR #6388, opened 2022-10-13
COM Event System loading new DLL

shainw:
Change(s):
   - New detection

   Reason for Change(s):
   - Covering existing technique

   Version Updated:
   - N/A

   Testing Completed:
   - Yes

   - Yes


shainw:
Approved via discussion with some EDR folks.

Issue #6397, opened 2022-10-15
Workspace parser functions for Infoblox have major issues.


jordanmills:
The parser functions for infoblox DHCP and DNS syslog messages (at https://github.com/Azure/Azure-Sentinel/tree/master/Solutions/Infoblox%20NIOS/Parser) seem to all have major issues and have not been validated against real world data (or perhaps at least not recently - if the OMS agent has changed behavior in the past few months).  They seem to try to parse data that is outside of the syslog message which leads to blank fields that compete with populated header fields, have poor filtering that leads to unrelated messages being parsed and returned with unpredictable results, and many fields are missing or parsed incorrectly because the parsers do not account for various formats used in the messages (all of which are used and are not configurable).  The end result is that the data from infoblox is useless in sentinel.

To Reproduce:
1. stream syslog messages from infoblox through a forwarding server with an azure agent to a log analytics workspace
2. run the infoblox functions against this data
3. note missing fields and unexpected data

Expected behavior:
Messages parsed limited to message type indicated by function, parsing of messages accounting for formats and fields present in messages in the default (and only) configuration.

I can share a workspace or archive of the messages if necessary to allow anyone to reproduce the issues.  I can also share some of the fixed parsers that I've made.  I'm not sure they return the data that Sentinel expects or can use, but the data they return is consistent.

jordanmills:
One of the main problems seems to be in what's documented as the format in the functions.  It looks like someone wrote parsers based on a couple of raw syslog messages without understanding how syslog, the oms agent, or log analytics work and store the messages, then never tested the parsers against real data.

From https://github.com/Azure/Azure-Sentinel/blob/master/Solutions/Infoblox%20NIOS/Parser/Infoblox_dnszone.txt
// LOG SAMPLES:
// This parser assumes the raw log are formatted as follows:
// 
//      May 13 12:05:52 10.0.0.0 dhcpd[30174]: DHCPDISCOVER from 0a:0b:0c:0d::0f via eth2 TransID 5daf9374: network 10.0.0.0/24: no free leases
//      May 13 12:05:52 10.1.1.1 named[11325]: zone voip.abc.com/IN: ZRQ applied transaction 0101010 with SOA serial 9191919. Zone version is now 0202020

The parser may be written to assume that (assuming there's a missing "messages" that should be there), but the raw log messages are not in LA and not accessible to the parser.  In addition, the assumption is incorrect.  The format of the message in the second example (note this is the message and not the raw message) is:
[...]

PR #6398, opened 2022-10-15
Update Infoblox_dhcpdiscover.txt

jordanmills:
Added optional router field.  Fixed parser to account for more message formats in this type not previously considered.  Simplified parser logic.  Fixed parsing in edge cases with unexpected colons.

   Change(s):
   Updated infoblox parser functions with similar fixes: Updated to only parse the MSG (RFC3164) part of the Syslog message. Excluded the Header. (actually in code and not just saying it was done in the comments). Added optional router field (excluded with final parse-away so output format is identical to previous version).  Fixed parser to account for message formats in this type not previously considered.

   Reason for Change(s):
   To resolve issues in #6397

   Version Updated:
   Yes.

   Testing Completed:
   Yes, tested against real world data in an LA workspace.  Which does not seem to have previously been done.

   Yes

### DIFF
--- a/Solutions/Infoblox NIOS/Parser/Infoblox_dhcpack.txt
+++ b/Solutions/Infoblox NIOS/Parser/Infoblox_dhcpack.txt
@@ -1,8 +1,8 @@
 // Title:           Infoblox parser for type - DHCPACK
 // Author:          Microsoft
-// Version:         1.1
-// Last Updated:    20/05/2022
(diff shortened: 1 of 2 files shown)


jordanmills:

jordanmills:

jordanmills:
@microsoft-github-policy-service agree company="default"

v-mchatla (approved):
Parsers are working with out any errors.

PR #6595, opened 2022-11-10
asim/kustoking/add-registryevent-microsoftwindows-events

KustoKing:
Change(s):
   - Add Registry ASIM Security Event 

   Reason for Change(s):
   - New parser

   Testing Completed:
   - Yes
   
   - Yes


KustoKing:
Closed thise one because I will update the current parser

PR #6596, opened 2022-11-10
Secured params and Added additional TVM API call.

aSquared01:
   
   Change(s):
   - Secured the Logic Apps parameters so the secret is not exposed in clear text after deployment.
   - Removed "UserName" ARM parameter and hardcoded generic display name for Log Analytics connector to remove confusion.
   - Added additional TVM API call to get additional information into Log Analytics. 

   Reason for Change(s):
   - Secrets in clear text is not a good idea.
   - Critical TVM information was missing from the solution.

   Version Updated:
   - N/A

   Testing Completed:
   - Yes, deployed in test environment and confirmed data was getting ingested into Log Analytics.

   - No KQL or YAML.

### DIFF
--- a/Playbooks/M365-Security-Posture/azuredeploy.json
+++ b/Playbooks/M365-Security-Posture/azuredeploy.json
@@ -31,10 +31,6 @@
             "metadata": {
                 "description": "The primary or secondary key of the Azure Log Analytics workspace."
             }
-        },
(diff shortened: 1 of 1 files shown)


aSquared01:

aSquared01:
Additional requirements being worked on. Will submit when complete.

PR #6682, opened 2022-11-18
Create ASIM_LookupHTTPStatusCode.yaml

goosvorbook:
Change(s):
   - I created a lookup table for HTTP Status Codes

   Reason for Change(s):
   - Can be a useful addition for Web SErver logs

   Version Updated:
   - n/a -> v0.1


ASIM/lib/functions/ASIM_LookupHTTPStatusCode.yaml
```diff
@@ -0,0 +1,128 @@
...
+      "460":"AWS ELB Client closed the connection with the load balancer before the idle timeout period elapsed",
+      "463":"AWS ELB The load balancer received an X-Forwarded-For request header with more than 30 IP addresses"
+    });
+    let HTTPStatusCodeLookup = HTTPStatusCodesTable[tostring(StatusCode)];
+    case (
+      HTTPStatusCodeLookup != "", HTTPStatusCodeLookup,
+      'Unassigned'
+    )
+  };
+print function(400)
```
oshezaf, line 127: Well, I assume this is only for testing and not the final :-)

ASIM/lib/functions/ASIM_LookupHTTPStatusCode.yaml
```diff
@@ -0,0 +1,128 @@
...
+Description: |
+  This ASIM function returns the HTTP Status code name associated with the response code numerical value provided as a parameter. For example, for the Status code 200, the function returns OK.  
+FunctionName: ASIM_LookupHTTPStatusCode
+EquivalentBuiltInFunction: _ASIM_LookupHTTPStatusCode
+FunctionParams:
+  - Name: StatusCode
+    Type: int
+
+FunctionQuery: |
+let function = (StatusCode:int) {
```
oshezaf, line 19: For YAML validation, the value of the "FunctionQuery" field from line 19 and on must be indented.

oshezaf, line 19: Since "EventRestulDetails" and the alias "HttpStatusCode" are strings, it makes sense to except string as a parameter. If the user passes int, it would be automatically converted.

oshezaf (changes requested):

oshezaf (approved):

PR #6813, opened 2022-12-05
Hyperlink validation fix

v-amolpatil:
   
   Change(s):
   - Fix changes when there are files related to azuredeploy, host, .py, txt, function.py etc files should be ignored as it might have half links i.e links used in variables as partial

   Reason for Change(s):
   - See guidance below

   Version Updated:
   - See guidance below

   Testing Completed:
   - See guidance below

   - See guidance below


mkchiliveri (approved):
looks good, approved

v-dvedak (approved):

PR #6846, opened 2022-12-07
Updated IOC matching queries to fix issues with entity mapping

petebryan:
   
   Change(s):
   - Updated IoC matching queries to force output of entities to be strings in cases where they could be strings or dynamic. This was causing issues with entity mapping when deploying tempaltes.

   Reason for Change(s):
   - Bug fix

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - See guidance below


aprakash13 (approved):
Thanks @petebryan

PR #6897, opened 2022-12-13
Linux sysmon3 update logic

goosvorbook:
   
   Change(s):
Changed the Initiated logic, changed the parsing.

   Reason for Change(s):
This logic is not relying on 'probability' but on something as reported by the event log (namely the Initiated value).

### DIFF
--- a/Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionMicrosoftLinuxSysmon.yaml
+++ b/Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionMicrosoftLinuxSysmon.yaml
@@ -1,12 +1,12 @@
 Parser:
   Title:  Network Session ASIM parser for Sysmon for Linux
-  Version: '0.3'
+  Version: '0.4'
   LastUpdated: Dec 13, 2022
 Product:
   Name: Sysmon for Linux
 Normalization:
   Schema: NetworkSession
-  Version: '0.2.0'
+  Version: '0.2.5'
 References:
 - Title: ASIM Network Session Schema
   Link: https://aka.ms/ASimNetworkSessionDoc
@@ -102,7 +102,7 @@ ParserQuery: |
             EventStartTime = EventEndTime,
             EventCount = int(1),
             EventVendor = 'Microsoft',
-            EventSchemaVersion = '0.2.0',
+            EventSchemaVersion = '0.2.5',
             EventSchema = 'NetworkSession', 
             EventProduct = 'Sysmon for Linux',
             EventResult = 'Success',
--- a/Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionMicrosoftLinuxSysmon.yaml
+++ b/Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionMicrosoftLinuxSysmon.yaml
@@ -1,7 +1,7 @@
 Parser:
   Title:  Network Session ASIM parser for Sysmon for Linux
   Version: '0.3'
-  LastUpdated: Jun 16, 2022
+  LastUpdated: Dec 13, 2022
 Product:
   Name: Sysmon for Linux
 Normalization:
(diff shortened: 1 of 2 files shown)


Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionMicrosoftLinuxSysmon.yaml
```diff
@@ -25,40 +25,52 @@ ParserQuery: |
     Syslog | where not(disabled)
     | where SyslogMessage has_all ('<Provider Name="Linux-Sysmon"', '<EventID>3</EventID>')
     | project-away ProcessName, ProcessID
-    | parse SyslogMessage with * '<Data Name="SourceIp">' SrcIpAddr:string '</Data>' *
-    | extend outbound = (SrcIpAddr == HostIP or SrcIpAddr in ('127.0.0.1', '0.0.0.0'))
+    | parse-kv SyslogMessage as (
```
oshezaf, line 28: No need to use parse-kv for a single parameter. Parse works just as well, and is much more efficient. It also does not need the regex mode in this case.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionMicrosoftLinuxSysmon.yaml
```diff
@@ -25,40 +25,52 @@ ParserQuery: |
...
-          '<Data Name="SourceHostname">' SrcHostname:string '</Data>'
-          '<Data Name="SourcePort">' SrcPortNumber:int '</Data>'
-          '<Data Name="SourcePortName">' SrcPortName:string '</Data>'
-          '<Data Name="DestinationIsIpv6">' DestinationIsIpv6:bool '</Data>'
-          '<Data Name="DestinationIp">' DstIpAddr:string '</Data>'
-          '<Data Name="DestinationHostname">' DstHostname:string '</Data>'
-          '<Data Name="DestinationPort">' DstPortNumber:int '</Data>'
-          '<Data Name="DestinationPortName">' DstPortName:string '</Data>'
-          *
+      | parse-kv SyslogMessage as (
```
oshezaf, line 34: The challenge with such changes is that I prefer not to do them without testing

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionMicrosoftLinuxSysmon.yaml
```diff
@@ -25,40 +25,52 @@ ParserQuery: |
...
+                 DestinationHostname:string ,
+                 DestinationPort:int,
+                 DestinationPortName:string
+                ) with (regex=@'<Data Name="(\w+)">{?([^>]*)}?</Data>')
+      | project-rename
+                EventEndTime = UtcTime,
+                SysmonComputer = Computer,
+                Process = Image,
+                EventOriginalUid = EventRecordID,
+                Computer = SysmonComputer,
```
oshezaf, line 61: This would probably not work.

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionMicrosoftLinuxSysmon.yaml
```diff
@@ -73,7 +73,6 @@ ParserQuery: |
         | where (array_length(srcipaddr_has_any_prefix)==0 
                   or has_any_ipv4_prefix(SrcIpAddr,srcipaddr_has_any_prefix)
                 ) 
-        | extend outbound = (SrcIpAddr == HostIP or SrcIpAddr in ('127.0.0.1', '0.0.0.0'))
```
oshezaf, line 76: No replacement

oshezaf (changes requested):
Seems to be premature. Let's not continue with that. I do want to go over the comments together.

PR #6959, opened 2022-12-23
Updated schedule for Duo security functionApp

v-sabiraj:
image
image


DataConnectors/Duo Security/AzureFunctionDuoSecurity/function.json
```diff
@@ -4,7 +4,7 @@
       "name": "Timer",
       "type": "timerTrigger",
       "direction": "in",
-      "schedule": "%Schedule%"
```
v-rucdu, line 7: @v-sabiraj instead of making this change, let it be configurable only and just add the schedule with cron value in the properties section of the azure deploy file. Please refer to the [**AWS CloudTrail Azure Function**](https://github.com/Azure/Azure-Sentinel/blob/master/DataConnectors/AWS-CloudTrail-AzureFunction/azuredeploy_awscloudtrail.json)

v-rucdu (approved):

PR #7017, opened 2023-01-03
Hyperlink Validations with Error

v-amolpatil:
   
   Change(s):
   - Updated hyperlinkValidator.ts file

   Reason for Change(s):
   - When a new image is added in PR which is not in master then we should validate the image is present in the PR. If the image is from the github repo then check if the image is in the PR changes list and still if link is breaking then build should be failed.

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes

**On Success:**
image

**On Failure:**
image


mkchiliveri:
Looks good, Approved

v-amolpatil:
Waiting for response from Pratik

v-dvedak (approved):

PR #7105, opened 2023-01-16
Enhancement:Identify targetProduct change With Azure Sentinel

v-amolpatil:
   
   Change(s):
   - When a Playbook file or mainTemplate.json file has "targetProduct": "Azure Sentinel" then we should skip the tag key value pair only. If there are Azure Sentinel in the same file at other places then we should fail the build.

   Reason for Change(s):
   - We are using "targetProduct": "Azure Sentinel" in Playbooks and we need to keep it intact so we should support skip for such cases of tag not complete file.

   Version Updated:
   - No

   Testing Completed:
   - Yes

   - Yes

**Scenarios I tested are as below and are passing:**
targetproduct = yes and value AS, azure sentinel in file yes ==> fail the build
targetproduct = yes and value AS, no azure sentinel in file ==> pass
targetproduct = yes and value Not AS, has azure sentinel in file ==> fail 
targetproduct = not present, no azure sentinel in file ==> pass
targetproduct = not present, azure sentinel in file ==> fail

**Screenshot while testing PR:**
image

### DIFF
--- a/.script/contentValidator.ts
+++ b/.script/contentValidator.ts
@@ -9,12 +9,22 @@ export async function ValidateFileContent(filePath: string): Promise<ExitCode>
     {
         const fileContent = fs.readFileSync(filePath, "utf8");
         const searchText = "Azure Sentinel"
-        const hasAzureSentinelText = fileContent.toLowerCase().includes(searchText.toLowerCase());
-        const hasTargetProductAzureSentinel = fileContent.includes('"targetProduct": "Azure Sentinel"')
+        const replaceText = '"targetProduct": "Azure Sentinel"';
 
-        if (hasAzureSentinelText && !hasTargetProductAzureSentinel)
+        const hasTargetProductAzureSentinel = fileContent.includes(replaceText);
+        const replacedFileContent = fileContent.replace(replaceText, "");
+        const hasAzureSentinelText = replacedFileContent.toLowerCase().includes(searchText.toLowerCase());
+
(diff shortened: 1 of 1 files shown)


rahul0216 (approved):
Changes look good.

v-rucdu (approved):

.script/contentValidator.ts
```diff
@@ -9,11 +9,22 @@ export async function ValidateFileContent(filePath: string): Promise<ExitCode>
...
+        const hasTargetProductAzureSentinel = fileContent.includes(replaceText);
+        const replacedFileContent = fileContent.replace(replaceText, "");
+        const hasAzureSentinelText = replacedFileContent.toLowerCase().includes(searchText.toLowerCase());
+
         if (hasAzureSentinelText)
         {
-            throw new Error(`Please update text from 'Azure Sentinel' to 'Microsoft Sentinel' in file '${filePath}'`);
+            if (hasTargetProductAzureSentinel)
+            {
+                throw new Error(`Please update text from 'Azure Sentinel' to 'Microsoft Sentinel' except 'targetProduct' key-value pair in file '${filePath}'`);
```
keyoke, line 22: Should this line really throw an error? I would think it should only emit log message i.e. console.log(`Please update text from 'Azure Sentinel' to 'Microsoft Sentinel' except 'targetProduct' key-value pair in file '${filePath}'`)

PR #7113, opened 2023-01-17
Fix to handle multiple replacement of targetProduct tag in Content Validation

v-amolpatil:
   
   Change(s):
   - When having multiple tags of "targetProduct" with value "Azure Sentinel" then it fails in current case. Javascript ".replace" function only replaces first occurance of the text not all matching text. So this new changes will handle all matching text. Here "/g" is for global and "i" is for ignore case insensitivity.

   Reason for Change(s):
   - Fix issue of single targetProduct as there can be multiple targetProduct tags.

   Version Updated:
   - No

   Testing Completed:
   - Yes

   - Yes


rahul0216 (approved):
Changes look good.

v-dvedak (approved):

PR #7117, opened 2023-01-19
Optimizing bag_unpack to be more performant

aprakash13:
   
   Change(s):
   - Adding schema to bag_unpack in the existing query

   Reason for Change(s):
   - Optimizing the current query so that bag_unpack is more performant


petebryan (approved):
Great work Ajeet!

PR #7149, opened 2023-01-25
Update the value of Dvc Alias in Network Session Schema

devikamehra:
Change(s):
   - Modified parameters contributing to Dvc Alias in Native Network Session Table

   Reason for Change(s):
   - To increase coverage and ensure no duplicate data is ingested in Native tables

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


oshezaf (changes requested):
I don't think that in NetworkSession the destination system is a preferred choice for Dvc. I would stick to two other options making the "NetworkSession" variant the default one.

oshezaf (approved):

PR #7187, opened 2023-01-31
Update vimAuthenticationPostgreSQL.yaml

Steeeeeef:
Change(s):
   - Update vimAuthenticationPostgreSQL.yaml

Reason for Change(s):
   - Parser would not return records without the endtime being set to endtime instead of starttime.
   - I used this parser as template for building a different parser and the parser that I built contained the mistakes from the current version of this parser. Changing the starttime to endtime allows my parser to work and other parsers on the github also use endtime instead of starttime.

Version Updated:
   - No

Testing Completed:
   - No

   - No


oshezaf (changes requested):
Thanks @Steeeeeef for this important fix! We do need you to update the version to 0.1.2 as well as the date in the YAML file.

PR #7203, opened 2023-02-01
Skip Azure Sentinel Community Github text from validations

v-amolpatil:
   
   Change(s):
   - We should skip any text "Azure Sentinel Community Github" from file and should not fail in content validation build.

   Reason for Change(s):
   - In some files "Azure Sentinel Community Github" text is used as a "DetectionService" where this data is coming from a url and this looks for the text from the url response. If we make this as "Microsoft Sentinel Community Github" then it will not work as the url response from client data also needs to be updated which is not possible so we are skipping this text.
   Below is the screenshot page in which url has the data and query filters data by "Azure Sentinel Community Github" text:
image

   Version Updated:
   - No

   Testing Completed:
   - Yes

   - Yes

**Testing Result screenshots:**
**1. Failed image:**
image

**2. Success image:**
image

### DIFF
--- a/.script/contentValidator.ts
+++ b/.script/contentValidator.ts
@@ -9,38 +9,44 @@ export async function ValidateFileContent(filePath: string): Promise<ExitCode>
     {
         const fileContent = fs.readFileSync(filePath, "utf8");
         const searchText = "Azure Sentinel";
+        const expectedText = "Microsoft Sentinel";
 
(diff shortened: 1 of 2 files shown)


mkchiliveri (approved):
Looks good. Approved.

v-dvedak (approved):

PR #7260, opened 2023-02-02
ActivityStatusValue appears to be changed/updated

treyperrone:
   
Change(s):
   - Update search from `Succeeded` list with both `Succeeded` and `Success`. 
   - This proposed solution matches on both values and is case-insensitive.

 Reason for Change(s):
   - Having issues testing this rule in an environment; found that the `ActivityStatusValue` from table AzureActivity regarding workspaces/computes/delete appears to be formatted as `Success` instead of `Succeeded`
   - screenshot example:
   Screenshot 2023-02-02 at 2 57 42 PM

Version Updated:
   - Yes

Testing Completed:
   - Yes

### DIFF
--- a/Detections/SecurityAlert/Suspicious_WorkSpaceDeletion_Attempt.yaml
+++ b/Detections/SecurityAlert/Suspicious_WorkSpaceDeletion_Attempt.yaml
@@ -51,7 +51,7 @@ query: |
     (
     AzureActivity
     | where OperationNameValue hassuffix ("/workspaces/computes/delete")
-    | where ActivityStatusValue =~ "Succeeded"
+    | where ActivityStatusValue has_any ("Succeeded", "Success")
     | project TimeGenerated, _ResourceId, SubscriptionId, UserAccount=Caller, IpAddress=CallerIpAddress, CorrelationId, OperationId, ResourceGroup, TenantId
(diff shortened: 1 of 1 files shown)


v-mchatla:
Hi @devikamehra,
Can you please review and share your feedback.
Thanks

treyperrone:
Does MS have a published list of the tables? Reason I ask is that in this specific example I was testing I was deleting ML Machine Learning compute and noticed it did not trigger the alert.

Deleting other Azure services that write the appended value `workspaces/computes/delete` would also have the success/succeeded messages. I think maybe Azure Virtual Desktops also uses "workspaces" and should trigger for example.

Is there an easy way in docs to know which Azure services I should be testing against?

v-vdixit:
Hi @treyperrone we will provide you an update on this ASAP. Thanks!

v-vdixit:
Hi @treyperrone We are working on this, will update you shortly, Thanks!

treyperrone:
@v-vdixit  no worries. Also, i realize my prior screenshot wasnt completely clear. Here's a better version I just generated:
Screenshot 2023-02-15 at 11 07 37 AM

aprakash13 (approved):
Thanks for the PR and the suggested changes.

PR #7311, opened 2023-02-09
Remove trailing spaces in SEABORGIUMIOC.csv

ep3p:
Change(s):
   - Remove trailing spaces of domain IOCs

   Reason for Change(s):
   - They will not match events with the trailing space [See images]

image
image
image

   - Yes


v-sabiraj (approved):
looks good

v-dvedak (approved):

PR #7319, opened 2023-02-10
Rename Auth0 custom table to match documentation

rene-ammerlaan:
   
   Change(s):
   - Updated log type for Auth0Connector/main.py

   Reason for Change(s):
   - The Function App creates an invalid table named `Auth0_CL`, which doesn't match with the settings and documentation.

   Testing Completed:
   - ..

   -..

### DIFF
--- a/Solutions/Auth0/Data Connectors/Auth0Connector/main.py
+++ b/Solutions/Auth0/Data Connectors/Auth0Connector/main.py
@@ -17,7 +17,7 @@ WORKSPACE_ID = os.environ['WorkspaceID']
 SHARED_KEY = os.environ['WorkspaceKey']
 
 FILE_SHARE_CONNECTION_STRING = os.environ['AzureWebJobsStorage']
-LOG_TYPE = 'Auth0'
+LOG_TYPE = 'Auth0AM'
 
 MAX_SCRIPT_EXEC_TIME_MINUTES = 5
 logging.getLogger('azure.core.pipeline.policies.http_logging_policy').setLevel(logging.ERROR)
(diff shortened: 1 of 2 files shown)


v-vdixit:
@rene-ammerlaan We are looking into this, will update you shortly, Thanks!

v-mchatla:
Hi @rene-ammerlaan,
Sorry for the delay. We will review and provide you the update by 17th Feb EOD.
Thanks

v-mchatla:
Hi @rene-ammerlaan,
Can you please share the impact on existing customers and also requesting you to update the same file in the zip as well.
Thanks

rene-ammerlaan:
@v-mchatla existing customers would need to change their use-cases to the new table name `Auth0AM_CL`. The alternative is to change all settings and references from `Auth0AM_CL` to `Auth0_CL`.

v-mchatla:
Hi @rene-ammerlaan,
Can you please updated the zip as well. 
Thanks

rene-ammerlaan:
@microsoft-github-policy-service agree company="Nederlandse Loterij BV"

rene-ammerlaan:
@v-mchatla I've updated the ZIP file

v-mchatla:
Hi @rene-ammerlaan 
Do we have any update on the zip update?
Thanks

rene-ammerlaan:
@v-mchatla as mentioned the update zip file is in the PR.

v-mchatla:
Hi @rene-ammerlaan,
Thanks for accommodating the requested changes, will check and get back to you.
Thanks

v-mchatla:
Hi @rene-ammerlaan,
Changes look good. Can you please add union in the parse for both of these tables Auth0_CL and Auth0AM_CL instead of running on Auth0AM_CL, so that it will return accurate results for existing customers as they are going to have the data in 2 different tables at least for next couple of months - currently they are getting the logs in Auth0_CL table and new logs will get injected into Auth0AM_CL.
Let me know if you need any other details.
Thanks

rene-ammerlaan:
@v-mchatla I've added the union to the parser file, but with this change we add a temporary union that should be removed later. The query was previously broken due to the mismatch in table name, so we could also ignore that legacy table name entirely. This avoids the need to cleanup this union.

v-mchatla:
Hi @rene-ammerlaan,
Agree, just to ensure that there won't be any impact to the existing customer.
Thanks

v-mchatla (approved):
Changes look good

PR #7334, opened 2023-02-14
Add more cases to RareRunCommandPowerShellScript.yaml

ep3p:
Change(s):
   - Add case where OperationNameValue is not uppercase, and ActivityStatusValue can have other value for successful operation.

   Reason for Change(s):
   - AzureActivity events with the same OperationNameValues can have different parsings, so the rules must contemplate both possible parsings.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


devikamehra (approved):
The change looks good

v-mchatla (approved):

v-dvedak (approved):

PR #7336, opened 2023-02-14
Solution Domain and Vertical validation.

mkchiliveri:
   
   Change(s):
   - Added new validation for validating the solution domain and vertical.

   Reason for Change(s):
   - Added new validation for validating the solution domain and vertical.

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes


.azure-pipelines/solutionValidator.yaml
```diff
@@ -0,0 +1,13 @@
+jobs:
+- job: "SolutionValidations"
```
anki-narravula, line 2: Idea is to expand / use the same pipelines for other solution validation scenarios ? Otherwise just for this test, looks too generic/

.script/solutionValidator.ts
```diff
@@ -0,0 +1,128 @@
...
+            console.warn(`No "resources" field found in the file. Skipping file path: ${filePath}`);
+            return ExitCode.SUCCESS;
+        }
+
+        // get the resources from the file
+        let resources = jsonFile.resources;
+
+        // filter resources that have type "Microsoft.OperationalInsights/workspaces/providers/metadata"
+        const filteredResource = resources.filter(function (resource: { type: string; }) {
+            return resource.type === "Microsoft.OperationalInsights/workspaces/providers/metadata";
```
anki-narravula, line 39: === works ?

.script/solutionValidator.ts
```diff
@@ -0,0 +1,128 @@
...
+                if (element.hasOwnProperty("properties") === true) {
+                    // check if the "properties" field has a "categories" field
+                    if (element.properties.hasOwnProperty("categories") === true) {
+                        const categories = element.properties.categories;
+
+                        let invalidDomains = [];
+                        let invalidVerticals = [];
+
+                        // check if the categories have a "domains" field
+                        if (categories.hasOwnProperty("domains")) {
```
anki-narravula, line 53: We need to fail in case either "domains" property itself is not available or don't have at least 1 value.

.script/solutionValidator.ts
```diff
@@ -0,0 +1,128 @@
...
+                            for (const domain of domains) {
+                                // check if the domain is valid
+                                if (!validDomains.includes(domain)) {
+                                    invalidDomains.push(domain);
+                                }
+                            }
+                        }
+
+                        // check if the categories have a "verticals" field
+                        if (categories.hasOwnProperty("verticals")) {
```
anki-narravula, line 64: We need to fail in case either "verticals" property itself is not available or don't have at least 1 value ?

.script/solutionValidator.ts
```diff
@@ -0,0 +1,128 @@
+import { runCheckOverChangedFiles } from "./utils/changedFilesValidator";
```
anki-narravula, line 1: solutionValidator.ts - name is too generic unless you are planning to use for all the solution validation scenarios

.script/ValidDomainsVerticals.json
```diff
@@ -0,0 +1,35 @@
+{
```
anki-narravula, line 1: This config file live in a same folder as script or can we keep at a separate place. I am not sure how it was done for other testcases

anki-narravula (commented):
Added some comments.

.script/solutionValidator.ts
mkchiliveri: yes, it works.

.script/ValidDomainsVerticals.json
mkchiliveri: placed it in separate folder.

.script/solutionValidator.ts
mkchiliveri: Added logic for at least 1 domain

.azure-pipelines/solutionValidator.yaml
mkchiliveri: we are targeting to keep all solution related GitHub Validations here.

.script/solutionValidator.ts
mkchiliveri: Got a confirmation that verticals can be optional.

anki-narravula (approved):
Looks good. Approved

v-mchatla (approved):

PR #7372, opened 2023-02-19
Update TimeSeriesAnomaly-ProcessExecutions.yaml

princemathew:
Changing rundll.exe to rundll32.exe, which is the actual process name of the dll loader.

   Change(s):
   - Changing "rundll.exe" to "rundll32.exe"

   Reason for Change(s):
   -  rundll32.exe  is the actual name of the process used to load dll.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


v-prasadboke:
Hello @princemathew Please repackage the solution and update the version of the solution

princemathew:
> Hello @princemathew Please repackage the solution and update the version of the solution

PR #7411, opened 2023-02-22
Content Validation to check only Specific tag in Data,DataConnectors and CreateUiDefinition folder file

v-amolpatil:
   
   Change(s):
   - We will check only specific tag specified i.e. description and descriptionMarkdown tag text value which are visible to endusers. We will eventually grow this once we are sure on which tags we have to target.

   Reason for Change(s):
   - Due to dependency of Azure Sentinel in many cases we cannot change the text from Azure Sentinel to Microsoft Sentinel in lot of cases so we have kept the scope limited.

   Version Updated:
   - No

   Testing Completed:
   - Yes

   - Yes

**Testing Screenshots**
**1. Failure Result:**
MicrosoftTeams-image

**2. Success Result:**
image

### DIFF
--- a/.script/contentValidator.ts
+++ b/.script/contentValidator.ts
@@ -5,51 +5,50 @@ import * as logger from "./utils/logger";
 
 export async function ValidateFileContent(filePath: string): Promise<ExitCode> 
 {
-    if (!filePath.includes("azure-pipelines"))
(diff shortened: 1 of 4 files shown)


v-amolpatil:
It is in review

mkchiliveri (approved):
Looks good.

v-dvedak (approved):

PR #7461, opened 2023-03-01
Update gte_6_FailedLogons_10m.yaml

shainw:
Change(s):
   - Bin to 10m to make sure it is limited even if the rule default timespan is changed.
   - Argmax done on overall results as there is no reason to fire multiple rows for the same computer and user account combo
   - Updated entity mapping for user account and removed legacy entity mapping fields

   Reason for Change(s):
   - Reduce the number of results.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


ashwin-patil (approved):
Query changes validated. looks good.

PR #7509, opened 2023-03-07
Update SharePoint_Downloads_byNewUserAgent.yaml

shainw:
Change(s):
   - Adjusted to `8d` due to perf considerations and doing 8d as query should be comparing the last day to the previous 7 days to the last day, otherwise it will miss like actions for the same day one week ago
   - Change order of operation for checking time
   - Do not need to summarize the min/max time in the `RecentActivity` portion of the query as it is timebound to the last day
   - Do not need to check the time frame of the results as again it is already time bound to last day after the join back to get full details of identified events
   - Removing old entity mapping rows
   - Bringing through proper account entity fields
   - Adjusted order to make more table view of the results more useful

   Reason for Change(s):
   - Primarily performance and high alert count returns

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes

### DIFF
--- a/Solutions/Microsoft 365/Analytic Rules/SharePoint_Downloads_byNewUserAgent.yaml
+++ b/Solutions/Microsoft 365/Analytic Rules/SharePoint_Downloads_byNewUserAgent.yaml
@@ -10,7 +10,7 @@ requiredDataConnectors:
     dataTypes:
       - OfficeActivity
 queryFrequency: 1d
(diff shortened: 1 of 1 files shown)


ashwin-patil (dismissed):
changes looks good. not entirely sure in this case but do we want to use `make_set` instead of `make_list` in the output.

ashwin-patil (approved):

PR #7519, opened 2023-03-08
Fixingworkspacedetection

treyperrone:
Change(s):
  1. Updated method of finding IP in dataset with a `mv-expand on where` mechanism
  2. Added a new entity for the resource(s) deleted  

 Reason for Change(s):
   - From our testing across multiple tenants, some change around February 6/7th caused the data structure to shift slightly
   - existing method was a little more brittle as it expected the IP to exist in array/index [2] and the change shifted IP address to index [3]
   - **this is fairly critical, as the alert does not fire with the current [2] index and gives a false sense of security if enabled**
   - Using the `mv-expand` and `where` logic we can concisely select the the correct index based on key name, which should work even if more indexes are added later

 Version updated:
   - Yes

 Testing Completed:
   - Yes

### DIFF
--- a/Detections/SecurityAlert/Suspicious_WorkSpaceDeletion_Attempt.yaml
+++ b/Detections/SecurityAlert/Suspicious_WorkSpaceDeletion_Attempt.yaml
@@ -29,13 +29,16 @@ query: |
     | where TimeGenerated > ago(1d)
     | where ProductName == "Azure Active Directory Identity Protection"
     | where AlertName == "Sign-in from an infected device"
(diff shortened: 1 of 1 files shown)


v-atulyadav:
Hi @devikamehra @vakohl, could you please provide your feedback on this. Thanks.

devikamehra (approved):
Changes are looking good

v-atulyadav (approved):

v-dvedak (approved):

PR #7537, opened 2023-03-10
Add modern Recycle Bin path in Analytic rule malware_in_recyclebin.yaml

ep3p:
Change(s):
   - Add another path for the recycle bin files to be checked.

   Reason for Change(s):
   - It seems ```RECYCLER``` path was for XP and older systems and ```$RECYCLE.BIN``` is being used by newer Windows systems.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes

Maybe this rule should not look for the process to be a LOLBAS and the command to have a Recycle Bin path... maybe it should strictly look for a path that contains the recycle bin strings and the lolbas strings, but I prefer not to question that part in this pull request.


v-atulyadav:
Hi @devikamehra, could you please provide your feedback on this. Thanks

devikamehra (approved):

v-atulyadav (approved):

PR #7548, opened 2023-03-13
Update Dev-0530AVHits.yaml for performance

kfriede:
Change(s):
   - Update Dev-0530AVHits.yaml for performance

   Reason for Change(s):
   - The DeviceInfo table is significantly larger than SecurityAlert in most organizations, and a large join using DeviceInfo as $left was causing Log Analytics lookup failure leading to Auto Disablement of the analytic query. Per https://learn.microsoft.com/en-us/azure/data-explorer/kusto/query/lookupoperator, `join` assumes that $left is smaller than $right, which was not true in the previous commit. Alternatively, lookup assumes $left is the larger table.

   Version Updated:
   - Yes, 1.0.0 -> 1.0.1

   Testing Completed:
   - Need Help: KQL functions fine but can't verify with live data as I don't have means to trigger MDE to alert to Sienna* IOC's.  

   - Yes

### DIFF
--- a/Detections/SecurityAlert/Dev-0530AVHits.yaml
+++ b/Detections/SecurityAlert/Dev-0530AVHits.yaml
@@ -20,15 +20,15 @@ tags:
   -  Dev-0530 actors
 query: |
   let Dev0530_threats = dynamic(["Trojan:Win32/SiennaPurple.A", "Ransom:Win32/SiennaBlue.A", "Ransom:Win32/SiennaBlue.B"]);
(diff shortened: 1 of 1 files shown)


v-atulyadav:
Hi @devikamehra @vakohl, 
could you please provide your feedback on this. 
Thanks

devikamehra (approved):

v-atulyadav (approved):

v-dvedak (approved):

PR #7604, opened 2023-03-20
Minor updates on entity mapping - AAD

devikamehra:
Change(s):
   - Improved Entity mapping

   Reason for Change(s):
   - AAD queries required performance improvements
  
   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes

### DIFF
--- a/Solutions/Azure Active Directory/Analytic Rules/Sign-in Burst from Multiple Locations.yaml
+++ b/Solutions/Azure Active Directory/Analytic Rules/Sign-in Burst from Multiple Locations.yaml
@@ -1,7 +1,7 @@
 id: d3980830-dd9d-40a5-911f-76b44dfdce16
 name: GitHub Signin Burst from Multiple Locations
 description: |
-  'This alerts when there Signin burst from multiple locations in GitHub (AAD SSO).'
+  'This detection triggers when there is a Signin burst from multiple locations in GitHub (AAD SSO).'
 severity: Medium
 requiredDataConnectors:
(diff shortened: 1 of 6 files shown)


Solutions/Azure Active Directory/Analytic Rules/Sign-in Burst from Multiple Locations.yaml
```diff
@@ -22,19 +22,22 @@ relevantTechniques:
 query: |
   let aadFunc = (tableName:string){
   table(tableName)
-  | where AppDisplayName == "GitHub.com"
+  | where AppDisplayName =~ "GitHub.com"
   | where ResultType == 0
-  | summarize CountOfLocations = dcount(Location), Locations = make_set(Location), BurstStartTime = min(TimeGenerated), BurstEndTime = max(TimeGenerated) by UserPrincipalName, Type
+  | summarize CountOfLocations = dcount(Location), Locations = make_set(Location,100), BurstStartTime = min(TimeGenerated), BurstEndTime = max(TimeGenerated) by UserPrincipalName, Type
   | where CountOfLocations > 1
```
petebryan, line 28: This is quite simplistic, I think we would get better results from using series_decompose_anomalies. Often a user has more than 1 location legitimately especially when locations are close together. series_decompose_anomalies would show us when its out of normal activity for that user.
I think there was a similar query to this where you were going to use series_decompose_anomalies - so this might be a duplicate.

anki-narravula (approved):

PR #7610, opened 2023-03-21
Update Infoblox_dnsclient.txt

twi252:
Parser fix to match source Syslog.
   
   Change(s):
   - Added ':' to match ':' at the end of SrcPortNumber from the source Syslog data to parse correctly.

   Reason for Change(s):
   - Out of the box parser does not correctly parse and is missing fields because it is missing the ':'

   Version Updated:
   - No version update

   Testing Completed:
   - Making this change corrected the parsing in a local environment.

### DIFF
--- a/Solutions/Infoblox NIOS/Parser/Infoblox_dnsclient.txt
+++ b/Solutions/Infoblox NIOS/Parser/Infoblox_dnsclient.txt
@@ -26,7 +26,7 @@ let response =
     | parse SyslogMessage with *
         "client " SrcIpAddr: string
         "#" SrcPortNumber: int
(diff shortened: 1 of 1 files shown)


vakohl:
@twi252 

Can you please share few Infoblox Logs which you are parsing?

The logs against which we tested parser is in following format:

client 192.168.1.1#57398 UDP: query: a4.foo.com IN AAAA response: NOERROR +AED a4.foo.com. 28800 IN AAAA ab::a

Above logs are being parsed correctly with existing parser.

oshezaf:
The update addresses the new variant, but not both existing and new variants. Closing.

PR #7643, opened 2023-03-24
Windows Security Events - entity mapping and version property updates

rahul0216:
   
   Change(s):
   - Added Entity mapping wherever missing
   - Added version property

   Reason for Change(s):
   - Adding missing entity mapping and version properties

   Version Updated:
   - 1.0.1

   Testing Completed:
   - Functional testing

### DIFF
--- a/Solutions/Windows Security Events/Hunting Queries/UserAccountAddedToPrivlegeGroup.yaml
+++ b/Solutions/Windows Security Events/Hunting Queries/UserAccountAddedToPrivlegeGroup.yaml
@@ -17,7 +17,6 @@ relevantTechniques:
   - T1098
   - T1078
 query: |
-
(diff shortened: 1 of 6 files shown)


anki-narravula (changes requested):
@rahul0216 - Entity mapping for hunting queries is little tricky as these will be saved as "Saved Searches" and fetch back to hunting blade.
1. We need to make sure we extend each attribute in a separate extend statement
image
You can refer this for example - https://github.com/Azure/Azure-Sentinel/blob/master/Solutions/Network%20Session%20Essentials/Hunting%20Queries/MismatchBetweenDestinationAppNameAndDestinationPort.yaml
It doesn't consider the mapping (as entityMaopping) we have in yaml file. We depend only on queries. So please assign and change the query accordingly

rahul0216:
@anki-narravula , updated with suggested changes.

anki-narravula (approved):

PR #7671, opened 2023-03-28
Azure SQL - Hunting query improvements

rahul0216:
Change(s):
   - Updated entity mappings
   - Added strong entity for Account, Name and UPNSuffix
   - Added version property
   - Removed entity mapping for some queries, since columns were not projected

   Reason for Change(s):
   - Verifying hunting queries

   Version Updated:
   - Yes
   
   Testing Completed:
   - Functional testing


aprakash13 (commented):
A few small suggestions:

++ For Hunting queries version field is not needed. 

++ In the last hunting query (HuntingQuery-VolumeResponseRowsStatefulAnomalyOnDatabase.yaml) in the entity mapping for Account entity do we also want to add UPNSuffix as an identifier

++ The requiredDataConnectors field for these queries seems to be empty. Do we want to populate the requiredDataConnectors field for the hunting queries with something like below:

requiredDataConnectors:
  - connectorId: AzureSql
    dataTypes:
      - AzureDiagnostics

rahul0216:
Thanks @aprakash13 for comments.
- Corrected entity mapping for HuntingQuery-VolumeResponseRowsStatefulAnomalyOnDatabase.yaml
- Populated missing requiredDataConnectors field
- we have decided to add version filed, so that like analytical query it can be used to prompt user to update hunting queries.
Please let me know if any more suggestions.

aprakash13 (approved):

PR #7724, opened 2023-04-01
Added support for the case when adding a user takes place outside/wit…

WojciechLesicki:
…hout PIM.
Currently this detection works only in cases when the user is added to the role via PIM, more precisely if PIM added via "eligible". If PIM performs user activations, but especially when the user is added to a role in an environment where there is no PIM, this detection will not detect this case. I added support for the situation when adding a role takes place without PIM or when PIM activated the user.
To maintain the consistency of the fields that we provide in the detection result (Role and AddedUser), I added the iff() functions.

   Change(s):
   - Adding "Add member to role" as another OperationName
   - Adding iff() to correctly populate the Role and AddedUser fields

   Reason for Change(s):
   - No support for adding a user to a role with the word "admin" when PIM is not used or when PIM activated the user

   Version Updated:
   - No, will be in next commit

   Testing Completed:
   - I tested the rule in a test environment.

   - Change successfully tested via KqlvalidationsTests...still wating for DetectionTemplateSchemaValidation results.

### DIFF
--- a/Detections/AuditLogs/UserAddedtoAdminRole.yaml
+++ b/Detections/AuditLogs/UserAddedtoAdminRole.yaml
@@ -21,16 +21,27 @@ tags:
 query: |
   AuditLogs
     | where OperationName in ("Add eligible member (permanent)", "Add eligible member (eligible)", "Add member to role")
-    | extend Role = iff(OperationName=="Add member to role", tostring(parse_json(tostring(parse_json(tostring(TargetResources[0].modifiedProperties))[1].newValue))),tostring(TargetResources[0].displayName))
-    | where Role contains "admin"
-    | extend AddedBy = iff(tostring(parse_json(tostring(InitiatedBy.user)).userPrincipalName)=="", tostring(parse_json(tostring(InitiatedBy.app)).displayName), tostring(parse_json(tostring(InitiatedBy.user)).userPrincipalName))
+    | mv-apply TargetResource = TargetResources on 
+    (
+        where TargetResource.type =~ "User"
+        | extend Target = tostring(TargetResource.userPrincipalName),
+                 props = TargetResource.modifiedProperties
+    )
+    | mv-apply Property = props on 
+    (
+        where Property.displayName =~ "Role.DisplayName"
+        | extend RoleName = trim('"',tostring(Property.newValue))
+    )
+    | where RoleName contains "admin"
+    | extend InitiatingApp = tostring(InitiatedBy.app.displayName)
+    | extend Initiator = iif(isnotempty(InitiatingApp), InitiatingApp, tostring(InitiatedBy.user.userPrincipalName))
     | extend AddedUser = iff(OperationName=="Add member to role",tostring(TargetResources[0].userPrincipalName),tostring(TargetResources[2].userPrincipalName))
-    | project-reorder TimeGenerated, AddedUser, Role, AddedBy
+    | project-reorder TimeGenerated, AddedUser, RoleName, Initiator
 entityMappings:
   - entityType: Account
     fieldMappings:
       - identifier: FullName
-        columnName: AddedBy
+        columnName: Initiator
   - entityType: Account
     fieldMappings:
       - identifier: FullName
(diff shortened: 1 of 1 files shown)


Detections/AuditLogs/UserAddedtoAdminRole.yaml
```diff
@@ -20,11 +20,11 @@ tags:
...
 query: |
   AuditLogs
-    | where OperationName in ("Add eligible member (permanent)", "Add eligible member (eligible)")
-    | extend Role = tostring(TargetResources[0].displayName)
+    | where OperationName in ("Add eligible member (permanent)", "Add eligible member (eligible)", "Add member to role")
+    | extend Role = iff(OperationName=="Add member to role", tostring(parse_json(tostring(parse_json(tostring(TargetResources[0].modifiedProperties))[1].newValue))),tostring(TargetResources[0].displayName))
     | where Role contains "admin"
-    | extend AddedBy = tostring(parse_json(tostring(InitiatedBy.user)).userPrincipalName)
-    | extend AddedUser = tostring(TargetResources[2].userPrincipalName)
+    | extend AddedBy = iff(tostring(parse_json(tostring(InitiatedBy.user)).userPrincipalName)=="", tostring(parse_json(tostring(InitiatedBy.app)).displayName), tostring(parse_json(tostring(InitiatedBy.user)).userPrincipalName))
```
devikamehra, line 26: InitiatedBy.user.userPrincipalName and InitiatedBy.app.displayName will get you the required data so no need to parse_json for it.

devikamehra (changes requested):
A very similar query resides in AAD solution already and thus this query may get decommissioned based on Pete's advise.
https://github.com/Azure/Azure-Sentinel/blob/master/Solutions/Azure%20Active%20Directory/Analytic%20Rules/UserAssignedPrivilegedRole.yaml

Query improvements can be taken from this query as well.

v-atulyadav:
Hi @WojciechLesicki, please check comments above from @devikamehra. Thanks,

Detections/AuditLogs/UserAddedtoAdminRole.yaml
```diff
@@ -20,11 +20,11 @@ tags:
   - AADSecOpsGuide
 query: |
   AuditLogs
-    | where OperationName in ("Add eligible member (permanent)", "Add eligible member (eligible)")
-    | extend Role = tostring(TargetResources[0].displayName)
+    | where OperationName in ("Add eligible member (permanent)", "Add eligible member (eligible)", "Add member to role")
+    | extend Role = iff(OperationName=="Add member to role", tostring(parse_json(tostring(parse_json(tostring(TargetResources[0].modifiedProperties))[1].newValue))),tostring(TargetResources[0].displayName))
```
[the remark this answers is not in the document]
WojciechLesicki: I made this change, request for review

Detections/AuditLogs/UserAddedtoAdminRole.yaml
WojciechLesicki: Great suggestion, changed :)

WojciechLesicki:
@devikamehra @v-atulyadav - thanks for the review! I've made changes as suggested, a request to check especially in the context of when OperationName is "Add eligible member (permanent)" or "Add eligible member (eligible)".

WojciechLesicki:

v-atulyadav:
Hi @devikamehra, @WojciechLesicki has responded on your comments please check. Thanks,

WojciechLesicki:
I agree the rules are very similar. BulkChangestoPrivilegedAccountPermissions, on the other hand, applies to the bulk situation, through the condition: 
`where dcount_Target > 9`
Considering the environment where role assignment action (especially to the roles containing "admin") are not very often a situation, this rule will perform better.

Another question is if this rule duplicate the UserAssignedPrivilegedRole rule that you mentioned some time ago.

devikamehra:
> I agree the rules are very similar. BulkChangestoPrivilegedAccountPermissions, on the other hand, applies to the bulk situation, through the condition: `where dcount_Target > 9` Considering the environment where role assignment action (especially to the roles containing "admin") are not very often a situation, this rule will perform better.
> 
> Another question is if this rule duplicate the UserAssignedPrivilegedRole rule that you mentioned some time ago.

I agree @WojciechLesicki, UserAssignedPrivilegedRole is a duplicate and thus would check with Pete on this.

devikamehra (approved):

v-atulyadav (approved):

v-dvedak (approved):

PR #7750, opened 2023-04-04
Updates for first cred added, performance, cleanup, entity addition

treyperrone:
   
   Change(s):
   - Removed OperationName search for `add service principal` as brings more results in to filter on and not needed, `where old_value_set == "[]"` looks for empty prior value and restricts to FIRST password or certificate added only.
   - Removed comment for code removed in [prior commit ](https://github.com/Azure/Azure-Sentinel/commit/634c4abe7d42332ad44ce22aa5038518a514bd90?diff=unified#diff-157d168077ce95b5777c0525ed4ea258ef8d3410e1fd9c9e534dd665b2442597L46) | [PR](https://github.com/Azure/Azure-Sentinel/pull/7589)
   - Add entity for SPN/App Registration name

   Reason for Change(s):
   - OperationName change results in less results to filter on, increases performance slightly
   - Remove comment since code was removed prior, prevent confusion
   - Entity added for Analyst easier to see, also populate the investigation graph/gui

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes

### DIFF
--- a/Solutions/Azure Active Directory/Analytic Rules/FirstAppOrServicePrincipalCredential.yaml
+++ b/Solutions/Azure Active Directory/Analytic Rules/FirstAppOrServicePrincipalCredential.yaml
@@ -24,7 +24,7 @@ tags:
   - NOBELIUM
 query: |
   AuditLogs
-  | where OperationName has_any ("Add service principal", "Certificates and secrets management") // captures "Add service principal", "Add service principal credentials", and "Update application - Certificates and secrets management" events
+  | where OperationName has ("Certificates and secrets management")
   | where Result =~ "success"
   | where tostring(InitiatedBy.user.userPrincipalName) has "@" or tostring(InitiatedBy.app.displayName) has "@"
   | mv-apply TargetResource = TargetResources on 
@@ -52,8 +52,7 @@ query: |
     )
   | extend InitiatingUserOrApp = iff(isnotempty(InitiatedBy.user.userPrincipalName),tostring(InitiatedBy.user.userPrincipalName), tostring(InitiatedBy.app.displayName))
   | extend InitiatingIpAddress = iff(isnotempty(InitiatedBy.user.ipAddress), tostring(InitiatedBy.user.ipAddress), tostring(InitiatedBy.app.ipAddress))
-  // The below line is currently commented out but Microsoft Sentinel users can modify this query to show only Application or only Service Principal events in their environment
-    | project-away new_value_set, old_value_set
+  | project-away new_value_set, old_value_set
   | project-reorder TimeGenerated, OperationName, InitiatingUserOrApp, InitiatingIpAddress, UserAgent, targetDisplayName, targetId, targetType, keyDisplayName, keyType, keyUsage, keyIdentifier, CorrelationId, TenantId
   | extend timestamp = TimeGenerated, Name = tostring(split(InitiatingUserOrApp,'@',0)[0]), UPNSuffix = tostring(split(InitiatingUserOrApp,'@',1)[0])
 entityMappings:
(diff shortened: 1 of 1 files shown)


v-rbajaj:
Looking into this

v-atulyadav:
Hi @v-rbajaj, please provide your feedback on this. Thanks

v-atulyadav:
Hi @v-rbajaj, please provide your feedback on this. Thanks

v-rbajaj:
Looking into this

Solutions/Azure Active Directory/Analytic Rules/FirstAppOrServicePrincipalCredential.yaml
```diff
@@ -24,7 +24,7 @@ tags:
   - NOBELIUM
 query: |
   AuditLogs
-  | where OperationName has_any ("Add service principal", "Certificates and secrets management") // captures "Add service principal", "Add service principal credentials", and "Update application - Certificates and secrets management" events
+  | where OperationName has ("Certificates and secrets management")
```
devikamehra, line 27: The author of the query has seen instances of SP getting used for thus activity. If this brings in false positives in your environment, making changes in active analytical rule would be a better choice.

v-rbajaj:
Hi @treyperrone, can you please look into Devika's comment?

v-atulyadav:
Hi @treyperrone, please act on @devikamehra's comment. Thanks

v-rbajaj:
Hi @treyperrone, request you to please look into @devikamehra's comment.

treyperrone:
'where old_value_set == "[]"' looks for an empty value being populated with a non-empty. 

From our LAWs datasets, this doesn't appear to be possible  in the Azure Portal at the initial SPN creation time. 

Looking at operation 'Add Service Principal' is not applicable; an update to that value empty or even a rotation appears to occur **only** under the operation 'Certificates and secrets management'

Is there some testing scenario from your side I could replicate here?

treyperrone:
> > 'where old_value_set == "[]"' looks for an empty value being populated with a non-empty.
> > From our LAWs datasets, this doesn't appear to be possible in the Azure Portal at the initial SPN creation time.
> > Looking at operation 'Add Service Principal' is not applicable; an update to that value empty or even a rotation appears to occur **only** under the operation 'Certificates and secrets management'
> > Is there some testing scenario from your side I could replicate here?
> 
> @treyperrone For non-empty old_value_set, there is another detection already in place. Does that solve your problem?

I dont understand what you mean at all. I dont have a problem to solve here.

The current rule is pulling more data in at the first couple steps as I described. This data is unnecessary which makes the query less efficient based on the testing we did since it will be discarded when it looks for the non-empty value. 

On other tickets Microsoft has asked us to provide data/logs/export. Can you please provide the data showing where this is needed? Maybe our testing is not extensive enough.

devikamehra:
> > > 'where old_value_set == "[]"' looks for an empty value being populated with a non-empty.
> > > From our LAWs datasets, this doesn't appear to be possible in the Azure Portal at the initial SPN creation time.
> > > Looking at operation 'Add Service Principal' is not applicable; an update to that value empty or even a rotation appears to occur **only** under the operation 'Certificates and secrets management'
> > > Is there some testing scenario from your side I could replicate here?
> > 
> > 
> > @treyperrone For non-empty old_value_set, there is another detection already in place. Does that solve your problem?
> 
> I dont understand what you mean at all. I dont have a problem to solve here.
> 
> The current rule is pulling more data in at the first couple steps as I described. This data is unnecessary which makes the query less efficient based on the testing we did since it will be discarded when it looks for the non-empty value.
> 
> On other tickets Microsoft has asked us to provide data/logs/export. Can you please provide the data showing where this is needed? Maybe our testing is not extensive enough.

@treyperrone I internally checked with the team and there are no recent examples of data supporting "Add service principal". I will get this PR go through for now and can revisit in future.

devikamehra (approved):

PR #7768, opened 2023-04-05
add the rulename, transactionID, Resourcetype and resourceID for WAF

treyperrone:
Change(s):
   - Add more details when alert fires for WAF item(s): ruleName_s, transactionId_g, ResourceType, ResourceId

   Reason for Change(s):
   - We had this alert fire off for an application gateway and the existing items pulled into the alert do not state the actual resourcetype (from 4 in rule), or any details for the Resource or WAF Rule. The offending IP, URL, statuscode, useragent are not enough to easily figure out "what" WAF protected (or perhaps unprotected) resource is in question.
   - Adding the rulename is helpful if the data is passed and not blocked - we see what rule allows attempt
   - Adding Resourcetype + ResourceID helps the analyst determine the target quickly
   - Adding the transactionID is a unique search item to dig in deeper for details from WAF?azurediagnostics log directly

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


v-prasadboke:
Looking into this

PR #7796, opened 2023-04-12
modify guid for uniqueness

austinmccollum:
If customers use the template as is with an ID that is a GUID only based on subscriptionID, any new CCP will overwrite previous CCP made with this template, including the GitHub data connector. By revising the GUID to be based on resourcegroup and deployment name, additional use of this template for new CCP won't overwrite previous.

Tested and found this revised Id GUID won't overwrite other CCP.

### DIFF
--- a/DataConnectors/Templates/Connector_API_CCP_template.json
+++ b/DataConnectors/Templates/Connector_API_CCP_template.json
@@ -9,8 +9,8 @@
     },
     "resources": [
         {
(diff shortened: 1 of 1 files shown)


v-vdixit:
Hi @austinmccollum we are looking into this, thanks!

v-vdixit:
Hi @austinmccollum we are still working on this, thanks!

v-vdixit (approved):
Looks good!

v-dvedak (approved):

PR #7942, opened 2023-04-28
EndpointThreatEssetials_AnalyticsRules

vakohl:
   
   Change(s):
   - Case-sensitivity, Entity mapping

   Reason for Change(s):
   - rules re-validation

   Version Updated:
   - various

   Testing Completed:
   - yes

   - yes

### DIFF
--- a/Solutions/Endpoint Threat Protection Essentials/Analytic Rules/DumpingLSASSProcessIntoaFile.yaml
+++ b/Solutions/Endpoint Threat Protection Essentials/Analytic Rules/DumpingLSASSProcessIntoaFile.yaml
@@ -1,9 +1,9 @@
 id: a7b9df32-1367-402d-b385-882daf6e3020
 name: Dumping LSASS Process Into a File
 description: | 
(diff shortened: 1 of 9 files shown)


rahul0216 (commented):
Rule PotentialRemoteDesktopTunneling.yaml has Name and UPNSuffix as Identifier for Account Entity but for many other rules Account Entity has only Name identifier. Any specific reason?

vakohl:
> Rule PotentialRemoteDesktopTunneling.yaml has Name and UPNSuffix as Identifier for Account Entity but for many other rules Account Entity has only Name identifier. Any specific reason?

vakohl:
> Rule PotentialRemoteDesktopTunneling.yaml has Name and UPNSuffix as Identifier for Account Entity but for many other rules Account Entity has only Name identifier. Any specific reason?

thanks for checking @rahul0216 Mostly other rules are based on 'Event' table. I found 'User' value in that table is always coming as windows username unlike email or UPN format. That's the reason I didn't split User.

rahul0216 (approved):
Changes look good.

Issue #7611, opened 2023-03-21
Infoblox parser (Solution & ASIM) typo


twi252:
Missing ':' after SrcPortNumber both in Infoblox NIOS solution parser & related built-in ASIM parsers.

**To Reproduce**
1. Ingest Infoblox data
2. The following KQL returns no data in the DnsQuery field
3. Infoblox | summarize count() by DnsQuery

**Expected behavior**
Results in the DnsQuery field from client DNS lookups.

**Additional context**
Created the following PR for the solution https://github.com/Azure/Azure-Sentinel/pull/7610 but the same problem exists in the built-in Infoblox parsers (_Im_Dns_InfobloxNIOSVxx & _Asim_Dns_InfobloxNIOSVxx) also preventing correct results there as well.

oshezaf:
hi @twi252 : my sample data does not have a colon between the port number and the protocol. Actually I believe your fix does not work with our sample events. We would need you to provide samples data to test against to improve the parser. 

Tagging @vakohl

twi252:
No problem that's fair, here are some raw query logs (from Syslog) as i see them.

22-Mar-2023 08:59:34.790 client 1.1.1.1#50164: UDP: query: 2.2.2.2.in-addr.arpa IN PTR response: NOERROR +A 3.3.3.3.in-addr.arpa. 3600 IN PTR myhost.mydomain.local.

22-Mar-2023 08:59:34.791 client 1.1.1.1#51443: UDP: query: myhost.mydomain.local IN A response: NOERROR +A thehost.mydomain.local. 3600 IN A 4.4.4.4

22-Mar-2023 08:59:34.804 client 1.1.1.1#50499: UDP: query: ahost.mydomain.local IN AAAA response: NOERROR +A

Perhaps it's ever so slightly changed based on the version of Infoblox, but i'm not certain the version of this one (not in my control) unfortunately.

twi252:
This is not concrete, but i saw a similar issue posted over at ArcSight.

Guessing this might have changed around v8.x

https://community.microfocus.com/cyberres/arcsight/f/arcsight-discussions/352069/infoblox-unparsed-events

Might have to tweak the parsers to cater for both cases. I can verify that i got a correct parse on my Infoblox DNS query logs with that change in the PR for the solution (not tested the change on the ASIM parsers), the rest of the fields seem to line up ok after the change.

Hope it helps.

twi252:
[...]

PR #7949, opened 2023-05-01
Asim dns infoblox parser update

vakohl:
   
   Change(s):
   - Updating the parse function logic to handle different SrcPortNumber format

   Reason for Change(s):
   - An issue was reported by a user. Old parser was not parsing DNS query logs correctly in some cases when the log format was slightly different. Probably due to a different version of Infoblox NIOS. Newer log was having an extra ':' just after SrcPortNumber in 'response' type, this wasn't the case with the old logs. Hence, causing parse function to fail. The updated code can handle both log formats.

   Version Updated:
   - 0.5.1 (VIM) and 0.1.3 (ASIM)

   Testing Completed:
   - Yes, Did test this updated parser with old as well as newer logs format 

   - Yes


oshezaf (commented):
@vakohl : I think the changes are good, but we need to make sure we have the new sample added somewhere. Maybe add them [here](https://github.com/Azure/Azure-Sentinel/blob/master/Sample%20Data/Syslog/InfobloxNIOS.json)

vakohl:
> @vakohl : I think the changes are good, but we need to make sure we have the new sample added somewhere. Maybe add them [here](https://github.com/Azure/Azure-Sentinel/blob/master/Sample%20Data/Syslog/InfobloxNIOS.json)

sure @oshezaf . I updated the new sample logs both our private repo as well as github public link you shared above using a different PR https://github.com/Azure/Azure-Sentinel/pull/7954

vakohl:
hi @oshezaf , can you please help approving this PR?

vakohl:
Github Issue link related to this PR: https://github.com/Azure/Azure-Sentinel/issues/7611

oshezaf:
> hi @oshezaf , can you please help approving this PR?

I left a comment. General one. Can you see it?

vakohl:
> > @vakohl : I think the changes are good, but we need to make sure we have the new sample added somewhere. Maybe add them [here](https://github.com/Azure/Azure-Sentinel/blob/master/Sample%20Data/Syslog/InfobloxNIOS.json)
> 
> sure @oshezaf . I updated the new sample logs both our private repo as well as github public link you shared above using a different PR #7954

@oshezaf are you referring to this comment? This has been taken care.

oshezaf (approved):

PR #7961, opened 2023-05-02
ASimTester: Relax NetworkSession/EventSeverity to Recommended

deggis:
In current ASIM common fields and NetworkSession schemas EventSeverity is listed as a Recommended field.

Change its class from Mandatory to Recommended.

https://learn.microsoft.com/en-us/azure/sentinel/normalization-common-fields
https://learn.microsoft.com/en-us/azure/sentinel/normalization-schema-network

   Change(s):
   - ASimTester: Change EventSeverity field's class from Mandatory to Recommended

   Reason for Change(s):
   - To have ASIM tester function match current documentation

   Version Updated:
   - N/A

   Testing Completed:
   - N/A
   - I tried modifying re-saving ASimSchemaTester with a URL to changed CSV version, but looks like the function cannot be saved using GUI.
   - ASimSchemaTester has parameters "T:(ColumnName:string,ColumnType:string),selected_schema:string", and the function dialog does not support that yet.

   - N/A


oshezaf (approved):

oshezaf:
Thanks.

PR #7967, opened 2023-05-04
Azure Active Directory - Analytic Rule Validation Fix

### LINKED ISSUE #7948: Three bugs in AzurePortalSigninfromanotherAzureTenant.yaml
The latest update for the analytic rule "Azure Portal sign in from another Azure Tenant" contains three bugs that will prevent us from updating the rule.  

The first bug: On lines 26 and 27 the line "with(format='multijson')" exists 2 times. The solution is to remove "with(format='multijson')" from line 27 as I did locally.
The second bug: the function "ipv6_is_in_any_range" is not available for Log Analytics but for Azure Data Explorer. No solution found, I've commented these out.
The third bug: The error in "Alert Description Format" in the section "Alert details" is "Too many column names. You can define up to 3 column names". The solution is to delete the latest sentence with a column "with the User Agent of {{UserAgent}}."

I did the bugfixes locally and the validation passed

v-mchatla:
**Azure Portal sign in from another Azure Tenant** Analytic Rule Validation Fix

   Change(s):
   - Modified the alert description to have only 3 parameterized column names

   Reason for Change(s):
   - To fix the Validation error

   Version Updated:
   - Yes

   Testing Completed:
   - Yes


v-mchatla:
Hi @petebryan,
Hope you are doing good.
This Analytic Rule 'Azure Portal sign in from another Azure Tenant' validation is failing due to the parameterized column details are more than 3 in Alert Description.
To fix the validation we have removed ipaddress value from the description as its already part of the alert title.
Can you please review and suggest if any further changes required.
Thanks

v-atulyadav:
Hi @petebryan, 
Please respond to the above comments from @v-mchatla. Thanks

petebryan:
Can I suggest we drop the UserAgent and keep the IP address in the description.

Also is 3 the limit for number of parameters? Our documentation in the wiki says there are limits but doesn't actually specify what they are, so if its 3 I will get that updated to reflect this.

v-mchatla:
> Can I suggest we drop the UserAgent and keep the IP address in the description.
> 
> Also is 3 the limit for number of parameters? Our documentation in the wiki says there are limits but doesn't actually specify what they are, so if its 3 I will get that updated to reflect this.

Hi @petebryan 
Thanks for reviewing the changes, I have updated the description as suggested. 
Also, to your second point - 3 is the limit for number of parameters we can use in Alert description. can you please update the document accordingly.
Thanks

petebryan (approved):

petebryan:
Thanks, I've approved and updated https://github.com/Azure/Azure-Sentinel/wiki/Query-Style-Guide#alertdetailsoverride to reflect this 3 paramters limit.

PR #8008, opened 2023-05-09
Add more operations in Detection/Analytics Rule SuspiciousLoginfromDeletedExternalIdentities.yaml

ep3p:
Change(s):
   1. Check deletion operations every hour, and check signin events for the last 24 hours.
   2. Check "Delete user" operation, instead of "Delete external user" operation.
   3. Join by a translated version of UserPrincipalName, instead of the original UserPrincipalName.

   Reason for Change(s):
   1. If query frequency and query period are the same, there could be a sign in event at 23:59, a deletion event at 00:01 the next day, and this rule might not be triggered by these two events.
   2. Not every guest user deletion has a "Delete external user" event, but I believe they all have a "Delete user" event where the string "#EXT#" will appear.
   3. "Delete user" events do not have the complete UserPrincipalName of a guest user, only the #EXT# version, where the "@" character has been replaced by "\_". We can't infer the original UserPrincipalName replacing all "\_" characters by "@".

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


v-atulyadav:
Hi @devikamehra, please provide your feedback on this. Thanks

devikamehra (changes requested):
Minor changes are required.

v-dvedak (approved):

PR #8055, opened 2023-05-16
Add IdentityInfo as a data connector

jannieli:
   
   Change(s):
   - Added IdentityInfo as a valid data connector

   Reason for Change(s):
   - Added a few UEBA detections featuring this table, and there are also published detections using it. Confirmed that it was valid.

   Version Updated:
   - N/A

   Testing Completed:
   -Yes

   - Yes


aprakash13 (approved):
Thanks @jannieli

PR #8085, opened 2023-05-22
SAP version 2.0.66

oferInbar:
please note that this version is already on marketplace
   
   Change(s):
   many (:))

   Reason for Change(s):
   catching up with post GA version

   Testing Completed:
   yes

   Version updated:
   - Yes


oferInbar:
have these to fix
1. missing SAP - (Preview) HANA DB - User Admin actions- need to add 
2. add parsers to https://github.com/Azure/Azure-Sentinel/tree/master/.script/tests/KqlvalidationsTests/CustomTables
3. non-ascii chars in unauthorized remote execution
4. fallbackResourceIDs should be removed from the 2 workbooks
5. remove /en-us/ from localle

PR #8208, opened 2023-06-07
CiscoDuoSecurity function app timeout issue

v-mchatla:
   
   Change(s):
   - Added condition to limit the max value to 1hr from mintime

   Reason for Change(s):
   - To avoid Function Timeout issues

   Version Updated:
   - No

   Testing Completed:
   - Yes


v-sabiraj (approved):

mnolan-ipsec:
Hi,

We recently deployed this data connector and have found the following issue in the merged code.

https://github.com/Azure/Azure-Sentinel/blob/48acc925ba5aef08d45a3dd1a1b06a34d3c922cf/Solutions/CiscoDuoSecurity/Data%20Connectors/AzureFunctionCiscoDuo/main.py#L112-L124

Line 119 creates a new mintime variable and multiplies it by 1000 because mintime is not present in the state.

Later on in line 124, the maxtime is created by referencing the mintime variable (that has already been multiplied by 1000) adds 3600 and then multiplies it by 1000 again, making it an invalid value.

Possible solution on line 124:
```python
maxtime = (int(mintime/1000) + 3600) * 1000
```

PR #8228, opened 2023-06-09
Removal of standalone content

kingwil:
   
   Change(s):
   - Removal of standalone content

   Reason for Change(s):
   - Gallery content is duplication of content in the Dynamics 365 solution. This change is to remove duplicated content so the Solution is the only sourc.

   Version Updated:
   - N/A

   Testing Completed:
   - N/A

   - N/A


devikamehra:
@kingwil Do we want to move this content in SAP solution or we already have corresponding queries in the solution folder?

I ask because the exercise of content migration does not delete the queries from the detection folder rather provides a link to the analytical rule in solutions folder.

kingwil:
> @kingwil Do we want to move this content in SAP solution or we already have corresponding queries in the solution folder?
> 
> I ask because the exercise of content migration does not delete the queries from the detection folder rather provides a link to the analytical rule in solutions folder.

@devikamehra This content relates to Dynamics 365, and there is a separate solution on Content Hub (not published on GitHub). The plan is to consolidate all content into the solution.

In this case the content will not be migrated since the solution will not be published on GitHub.

@yohasson  - anything to add to this ?

yohasson:
As the Dynamics 365 solutions will be monetized, we want to remove them completely from GitHub. Both the stand-alone content elements and the full solution folder. Thanks.

devikamehra:
@kingwil @yohasson Is this content referred in any of the blog posts? If so, the links would start breaking.

@prtanej How should we proceed in this case?

yohasson:
@devikamehra this content is not referred in any blog post.

prtanej:
If the content is planned to be added to the solution, we should go ahead and remove the standalone tags. Removing the standalone tags would effectively make it disappear from the Standalone content items in Content Hub and then @kingwil can have these added to the Dynamics solution so they remain available to customers via a packaged solution.

kingwil:
@prtanej  Standalone content in scope of this PR is already part of the Dynamics CE Apps solution. However, the solution is only available via Azure Marketplace and not published on GitHub. This is intentional as outlined by @yohasson earlier in the thread.

prtanej:
@kingwil Sounds good to me.

v-atulyadav (approved):

PR #8261, opened 2023-06-14
Secure Practice MailRisk data connector

ingin97:
   
   Change(s):
   - Update solution_id/offer_id and support address

   Reason for Change(s):
   - Correct mistakes

   Testing Completed:
   - Yes

   - Yes


Solutions/MailRisk/Package/2.0.2.zip
v-vdixit, line 1: Hi @ingin97 Please ensure that the package created only has the latest CreateUI and maintemplate files and not the older packages, currently I can see this zip file contains older packages too, please check on this, thanks!
image

v-vdixit: Hi @ingin97 thanks for updating package, your approach is correct the recent version of packaging tool includes older packages by default, we will work on updating this in the next version.

Solutions/MailRisk/Package/mainTemplate.json
```diff
@@ -30,9 +30,9 @@
     }
   },
   "variables": {
-    "solutionId": "securepractice.azure-sentinel-solution-mailrisk",
+    "solutionId": "securepractice.microsoft-sentinel-solution-mailrisk",
```
v-vdixit, line 33: Hi @ingin97 we are trying to deploy the solution but it is not reflecting post deployment, can you please confirm why this offer id has been updated, thanks!

ingin97: Hi @v-vdixit, the solution is not yet published, and the offer in the Partner for publishing somehow got mismatched with the publisher ID, we were adviced from a Partner Program Manager that the simplest fix was to change the offer name in the solution id for it to match properly.

v-vdixit (approved):

PR #8282, opened 2023-06-16
CiscoDuoSecurity Function App code update

mnolan-ipsec:
Fixed error : 400 Invalid request parameters ('maxtime' must be a timestamp in milliseconds)

   Change(s):
   - Updated maxtime variable definition when the difference is greater than 3600000 ms

   Reason for Change(s):
   - When the difference between the mintime & maxtime is greater than 3600000, the new maxtime value becomes invalid. This is because the value is referencing the existing mintime value (which is already multiplied by 1000) which is then multiplied by another 1000.
   - Resolves issue identified here: https://github.com/Azure/Azure-Sentinel/pull/8208#issuecomment-1594091137

   Version Updated:
   - N/A

   Testing Completed:
   - Yes. Testing results:
   image
   Changes made at: ~2023-06-16 05:15:00.000

   - Need help


mnolan-ipsec:
@microsoft-github-policy-service agree company="IPSec Pty Ltd"

v-vdixit (approved):

PR #8338, opened 2023-06-22
Cisco ISE ASIM AuditEvent Parser

spsocprime:
   
   Change(s):
   - Added Cisco ISE parser for ASIM AuditEvent Schema
   - Added tests files
   - Added Sample Data to "Sample Data/ASIM" folder

   Reason for Change(s):
   - Initial commit

   Testing Completed:
   - Yes

### DIFF
--- a/ASIM/dev/ASimTester/filteringTest/filteringTest.py
+++ b/ASIM/dev/ASimTester/filteringTest/filteringTest.py
@@ -1,0 +1,277 @@
+__unittest = True #prevents stacktrace during most assertions
+
+import unittest
+import os
+import sys
+import yaml
+import contextlib
+import argparse
+from datetime import datetime, timedelta, timezone
+from azure.monitor.query import LogsQueryClient, LogsQueryStatus
+from azure.identity import DefaultAzureCredential
+from azure.core.exceptions import HttpResponseError
+from azure.identity import DefaultAzureCredential
+from schemasParameters import all_schemas_parameters
+
+
+DUMMY_VALUE = "\'!not_REAL_vAlUe\'"
+
+
+argparse_parser = argparse.ArgumentParser()
+argparse_parser.add_argument("ws_id", help = "workspace ID")
+argparse_parser.add_argument("parser_file_relative_path", help = "relative path to the parser file")
+argparse_parser.add_argument("days_delta", default = 1, type = int, help = "number of days to query over ,default is set to  1", nargs='?')
+args = argparse_parser.parse_args()
+ws_id = args.ws_id
+parser_file_path = args.parser_file_relative_path
+days_delta = args.days_delta
+
+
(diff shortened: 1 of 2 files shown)


Parsers/ASimAuditEvent/Parsers/ASimAuditEventCiscoISE.yaml
```diff
@@ -0,0 +1,76 @@
...
+          , username contains "\\", "Windows"
+          , isempty(username), ""
+          , "Simple"
+      )
+    };
+    let CiscoISEAuditParser=(disabled:bool=false){
+      Syslog
+      | where not(disabled)
+      | where ProcessName has_any ("CISE", "CSCO")
+      | extend EventOriginalType = extract(@".*\d{10}\s(\d{4,5})\s.*", 1, SyslogMessage)
```
devikamehra, line 39: Value of EventOriginalType and EventOriginalSeverity can be extracted using parse statement as the initial format of the event is the same. This would avoid the use of inefficient regex experessions.

Parsers/ASimAuditEvent/Parsers/ASimAuditEventCiscoISE.yaml
```diff
@@ -0,0 +1,76 @@
...
..."60060","60061","60062","60063","60064","60068","60069","60070","60083","60084","60085","60086","60087","60088","60089","60090","60091","60092","60094","60095","60097","60098","60100","60101","60102","60103","60105","60106","60107","60111","60112","60113","60114","60118","60119","60120","60121","60122","60123","60126","60128","60129","60130","60131","60132","60133","60153","60166","60167","60172","60173","60187","60189","60193","60194","60197","60198","60199","60200","60201","60202","60207","60208","60209","60210","60213","60214","60215","60216","60451","60452","61002","61003","61004","61005","61006","61007","61008","61009","61016","61017","61018","61020","61022","61023","61024","61030","61033","61034","61051","61052","61100","61101","61102","61103","61104","61105","61106","61107","61108","61109","61110","61111","61114","61115","61116","61117","61118","61120","61121","61122","61123","61124","61125","61126","61127","61128","61129","61130","61158","61160","61161","61162","61163","61164","61165","61166","61167","61168","61169","61234","62000","62001","62002","62003","61300","61301","61302","60467","60468","62008","62009","62010","62011","62012","62013","62014","62015","62016","62017")
```
devikamehra, line 40: We don't need to repeat the 200+ Ids here. We can use the EventName datatable with project EventOriginalType and filter in the vim parser can also be applied in this table itself.
For example, for eventresult == "Success", filter this table itself and the rest of the parser will only get the records which are successful. Same is applicable to other filters too.

devikamehra (changes requested):

PR #8345, opened 2023-06-23
fixing handling of entity mappings in rules

javiersoriano:


javiersoriano:
the checks that fail have nothing to do with the code

javiersoriano:
@v-atulyadav can you check why the npm script is failing? it doesn't seem to be related with the code changes proposed here...

Thanks

sreedharande (approved):

PR #8399, opened 2023-06-29
Move all resources ApiVersion to a single place

v-amolpatil:
   
   Change(s):
   - Moved apiVersion for resources to common location

   Reason for Change(s):
   - Moved apiVersion for resources to common location. Instead of replacing at all places kept it at 1 place

   Version Updated:
   - No

   Testing Completed:
   - Yes

   - Yes


Tichandr (approved):
Looks good with the api versions for all sentinel resources

PR #8408, opened 2023-06-29
ASIM parsers filtering validations - dynamic tests - part1

TomerAligaev:
- Add utility functions for the dynamic parameters tests as part of the ASim parsers filtering validations.
- Rename the tests script file.
- Add Readme file.

### DIFF
--- a/ASIM/dev/ASimTester/filteringTest/ASimFilteringTest.py
+++ b/ASIM/dev/ASimTester/filteringTest/ASimFilteringTest.py
@@ -101,12 +101,30 @@ def create_execution_strings_with_one_parameter(parameter, value, column_name):
     return f"query({parameter}={value}) | summarize count() by {column_name}\n"
 
 
+def get_substring_or_default(default, substring, rows, current_list):
+    '''
+    Returning the substring of a string if this substring is not in the list ("current_list") and if this substring is not contained in all of the values in rows.
+    If the substring is not in the list and not contained in all of the values in rows, the substring is returned.
+    If the substring is in the list or contained in all of the values in rows, the default value is returned.
+    '''
+    if substring in current_list:
+        return default
+    # This prefix is the only possible prefix and thus, it is returned
+    if len(rows) == 1:
+        return substring
+    # Checking if there is at least one value in rows that the substring is not contained in.
+    for row in rows:
+        value = row[0]
+        if substring not in value:
+            return substring
+    return default
 
-def get_prefix(str, query_response, current_list):
+
+def get_prefix(str, rows, current_list):
     '''
     Returns the prefix of a string until its last dot, under certain conditions:
     - The prefix is not present in the 'current_list'.
-    - The prefix is not contained in all of the values in the 'query_response' logs query results.
+    - The prefix is not contained in all of the values in rows.
     If the string does not contain a dot or fails to meet the conditions above, the original string is returned.
     Example:
         str = "example.com.subdomain"
@@ -116,25 +134,16 @@ def get_prefix(str, query_response, current_list):
     # If there is no dot in the string, return the string
     if last_dot_index == -1:
         return str
+    
     substring = str[:last_dot_index]
-    # This prefix is the only possible prefix and thus, it is returned
-    if len(query_response.tables[0].rows) == 1 and len(current_list) == 0:
-        return substring
-    if substring in current_list:
-        return str
-    # Checking if there is at least one value in the query response that the prefix is not contained in.
-    for row in query_response.tables[0].rows:
-        value = row[0]
-        if substring not in value:
-            return substring
-    return str
+    return get_substring_or_default(str, substring, rows, current_list)
 
 
-def get_postfix(str, query_response, current_list):
+def get_postfix(str, rows, current_list):
     '''
     Returns the postfix of a string following its first dot, under certain conditions:
     - The postfix is not present in the 'current_list'.
-    - The postfix is not contained in all of the values in the 'query_response' logs query results.
+    - The postfix is not contained in all of the values in rows.
     If the string does not contain a dot or fails to meet the conditions above, the original string is returned.
     Example:
         str = "example.com.subdomain"
@@ -144,18 +153,9 @@ def get_postfix(str, query_response, current_list):
     # If there is no dot in the string, return the string
     if first_dot_index == -1:
         return str
+    
     substring = str[first_dot_index + 1:]
-    # This postfix is the only possible postfix and thus, it is returned
-    if len(query_response.tables[0].rows) == 1 and len(current_list) == 0:
-        return substring
-    if substring in current_list:
-        return str
-    # Checking if there is at least one value in the query response that the postfix is not contained in.
-    for row in query_response.tables[0].rows:
-        value = row[0]
-        if substring not in value:
-            return substring
-    return str
+    return get_substring_or_default(str, substring, rows, current_list)
 
 
 class FilteringTest(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


ASIM/dev/ASimTester/filteringTest/ASimFilteringTest.py
```diff
@@ -77,6 +77,12 @@ def create_parameters_string(parser_file):
     return ','.join(paramsList)
 
 
+# Creating a string of the values in the list with colons between them
```
ori-licht, line 80: >colons [](http://example.com/codeflow?start=51&length=6)

worth adding an example, and is that colons? or commas?

ASIM/dev/ASimTester/filteringTest/ASimFilteringTest.py
```diff
@@ -94,6 +100,55 @@ def create_execution_strings_with_one_parameter(parameter, value, column_name):
...
+        return str
+    # Checking if there is at least one value in the query response that the prefix is not contained in.
+    for row in query_response.tables[0].rows:
+        value = row[0]
+        if substring not in value:
+            return substring
+    return str
+
+
+def get_postfix(str, query_response, current_list):
```
ori-licht, line 128: >postfix [](http://example.com/codeflow?start=8&length=7)

?
maybe suffix?

ASIM/dev/ASimTester/filteringTest/ASimFilteringTest.py
```diff
@@ -218,6 +273,105 @@ def scalar_test_check_fictive_value(self, param_name, query_definition, column_n
             self.assertEqual(0, len(no_results_response.tables[0].rows), f"Parameter: {param_name} - Returned results for non existing filter value. Filtered by value: {DUMMY_VALUE}")
 
         
+    # Returning an array of at most two values from the query response, each string in the returned array is not contained in the other
```
ori-licht, line 276: this comment is not clear, what is the meaning of "each string in the returned array is not contained in the other"

ASIM/dev/ASimTester/filteringTest/ASimFilteringTest.py
```diff
@@ -218,6 +273,105 @@ def scalar_test_check_fictive_value(self, param_name, query_definition, column_n
...
+            if value == "":
+                continue
+            for row2 in query_response.tables[0].rows:
+                value2 = row2[0]
+                if value not in value2:
+                    values.append(value)
+                    break
+            if len(values) == 2:
+                break
+        return values
```
ori-licht, line 302: let's talk about this, it is not understood why you are dealing with the case where the length is 1, 2. 
why the general approach is not good enough

TomerAligaev: For the case of length 2 you are correct. I will remove it.
For the case of length 1 lets assume we have a response with only one value. The if condition in line 297 (if value not in value2) will not be true as both "value" and "value2" will have the value of the only element in the response.
As a result, this one value won't be appended to the returned array even though it should be appended.
By the way, following our teams meeting, this is the same reason why the case of length 1 is checked in both get_prefix and get_postfix.

ASIM/dev/ASimTester/filteringTest/ASimFilteringTest.py
```diff
@@ -218,6 +273,105 @@ def scalar_test_check_fictive_value(self, param_name, query_definition, column_n
...
+
+    def add_substring_to_list(self, query_response, substrings_list, num_of_substrings):
+        '''
+        The function adds at most "num_of_substrings" substrings of values from "query_response" to "substrings_list".
+        A substring of a value will be either its postfix from after the first dot in the value or its prefix until the first dot in the value.
+        '''
+
+        # Looking for values with substrings that can be appended to the list
+        for row in query_response.tables[0].rows:
+            value = row[0]
```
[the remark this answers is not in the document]
TomerAligaev: The rows in query_response are a result of a KQL query summarized and counted by a certain row.
For example:
_Im_NetworkSession_Microsoft365DefenderV04() | summarize count() by IpAddr 

image

So the value relevant to each row will be in row[0].

ori-licht:
if the tests are for query_response.tables[0].rows[0] in all cases, why not just pass that to all functions?
you are constantly using the "query_response" and inside the function you always have:
"query_response.tables[0].rows[0]"

ASIM/dev/ASimTester/filteringTest/ASimFilteringTest.py
ori-licht: Let's talk

ASIM/dev/ASimTester/filteringTest/ASimFilteringTest.py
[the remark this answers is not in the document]
ori-licht: not sure I understand the last sentence:
"the value relevant to each row will be in row[0]."

what do you mean by "each row" if you are only looking at row[0]?

v-atulyadav:
Hi @TomerAligaev, please act on @ori-licht's comments. Thanks

v-atulyadav:
Hi @TomerAligaev, waiting for your response on @ori-licht's comments. Thanks

ori-licht:
> Hi @TomerAligaev, waiting for your response on @ori-licht's comments. Thanks

Hi @v-atulyadav , Tomer is OOF for the upcoming weeks.

v-atulyadav:
sure @ori-licht thank you for this update. We look forward to Tomer's response.

ASIM/dev/ASimTester/filteringTest/ASimFilteringTest.py
ori-licht: def get_values_for_dynamic_tests(self, rows):
	if len(rows) == 1:
		return [] if rows[0][0] == "" else return [rows[0][0]]
		
	values = []
	# Searching values in rows which are not contained in at least one other value
	for row in rows:
		value = row[0]
		# if the value in an empty string we want to skip it
		if value == "":
			continue
			
		# if we already found 2 values that sutisfy the condifitons we can return them 
		if len(values) == 2:
			break
			
		for rowToCompareWith in rows:
			valueToCompareWith = rowToCompareWith[0]
			if value not in valueToCompareWith:
				values.append(value)
				# if we found one value that sutisfy the condition, we can continue to the next one
				break
	return values

TomerAligaev:
Passing now only rows instead of query_response.

---
In reply to: [1618014737](https://github.com/Azure/Azure-Sentinel/pull/8408#issuecomment-1618014737) [](http://example.com/codeflow?ancestors=1618014737)

ASIM/dev/ASimTester/filteringTest/ASimFilteringTest.py
TomerAligaev: Indeed it is commas, my mistake.
And added example.

ASIM/dev/ASimTester/filteringTest/ASimFilteringTest.py
TomerAligaev: Changed the comment.

ASIM/dev/ASimTester/filteringTest/ASimFilteringTest.py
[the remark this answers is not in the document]
TomerAligaev: Changed the function after talking face to face.

ASIM/dev/ASimTester/filteringTest/ASimFilteringTest.py
TomerAligaev: Changed the function after talking face to face.

ori-licht (approved):

v-atulyadav (approved):

PR #8419, opened 2023-07-03
Skip contentId and id error in arm-ttk github actions

v-amolpatil:
   
   Change(s):
   - Currently when we package a solution for 3.0.0 we get arm-ttk error for contentProductId and id. So we should skip this errors. This error will go away when a PR which is created against One arm-ttk repo is merged in that repo.

   Reason for Change(s):
   - To skip errors on arm-ttk for contentProductId and id

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes

**Testing output:**
**Output from pipeline:**
image

**When we run it on local we see below error for same solution package as below:**  
image

**If we add below properties in "One" arm-ttk repo on local to get rid of this error we need to add properties like below:**
image


mkchiliveri (approved):
Looks good

v-atulyadav (approved):

PR #8446, opened 2023-07-05
Data Connectors Kql validations tables update

v-vdixit:
   
   Change(s):

   Reason for Change(s):

   Version Updated:
   - See guidance below

   Testing Completed:
   - See guidance below

   - See guidance below


devikamehra (approved):
There is no direct way to validate the schema of these tables. Approval is provided with the trust that you have tested the changes and now the associated KQLs will not fail.

v-atulyadav (approved):

PR #8447, opened 2023-07-05
Custom tables added for Data connector validations

v-vdixit:
   
   Change(s):

   Reason for Change(s):

   Version Updated:
   - See guidance below

   Testing Completed:
   - See guidance below

   - See guidance below


devikamehra (approved):
There is no way to validate if the schemas are correct. Trusting that all the required KQLs will work now.

v-atulyadav (approved):

PR #8448, opened 2023-07-05
Adding custom tables for Data Connectors KQL validations

v-vdixit:
   
   Change(s):

   Reason for Change(s):

   Version Updated:
   - See guidance below

   Testing Completed:
   - See guidance below

   - See guidance below


devikamehra:
@v-vdixit Resolve the conflicts.

devikamehra (approved):
There is no way to validate the schemas. Thus approving with the trust that associated KQLs will not break now.

v-dvedak (approved):

PR #8486, opened 2023-07-10
Shift left package automation

v-amolpatil:
   
   Change(s):
   - When PR is open then anyone should be able to create package using /slash command. This will not work for fork prs as there will be security issues of sharing token.

   Reason for Change(s):
   - We should be able to create package using slash command.

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes


Tichandr (approved):
Looks Good

PR #8549, opened 2023-07-17
Added Readme, optional v3 execution

v-amolpatil:
   
   Change(s):
   - Added readme file for github pipeline workflow run which create package automatically when azure build validations get passed,
   - Make createSolutionV3.ps1 file for local run as optional i.e. when we run this file it ask for path to data file. So we have made this as an optional i.e if we dont provide path and execute it will run based on "input" folder file inside of "V3" folder. 
   - Kept 1 data file inside of data input folder as git doesnt recognize input folder alone and it requires atleast 1 file.

   Reason for Change(s):
   - This change from Will King and readme we added so that everyone is aware.

   Version Updated:
   - No

   Testing Completed:
   - Yes

   - Yes


Tichandr (approved):
Looks Good

PR #8556, opened 2023-07-17
slash command dispatch workflow change

v-amolpatil:
   
   Change(s):
   - Update token for slash command dispatch. We are not able to trigger github workflow when slash command due to permission restriction.

   Reason for Change(s):
   - slash command is not able to trigger packaging workflow so we have used read:org permission on token

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes


v-sabiraj (approved):

Tichandr (approved):
Looks Good

PR #8601, opened 2023-07-21
Remove concat concat for all resources

v-amolpatil:
   
   Change(s):
   - As per discussion with Sarath we need to remove `concat(concat` and version from the same string for all resources from packaging tool.

   Reason for Change(s):
   - Remove `concat(concat` and version for all resources.

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes

**After custom deployment for cisco umbrella**
image


Tichandr (approved):
Changes looks good for watchlist param issue and version issue change

Issue #8376, opened 2023-06-27
DNS workbook not working when the DnsEvents table gets large.


holi1-ops:
**Describe the bug**
When IP address is set to All, all the sub tabs would show error “couldn’t connect to server” even if we filter time range to a few hours. 
 
![image](https://github.com/Azure/Azure-Sentinel/assets/62872235/b5a64b88-c886-4f91-aa95-005de2a5eec4)

**To Reproduce**
1. Go to 'Sentinel Workbooks'
2. Click on 'DNS'
3. Select "All" (should be default) for the "IpAddress" filter with a relatively large DnsEvents table in the workspace

**Expected behavior**
We should be able to see the charts showing up in the sub panels

**Screenshots**
From the captured browser trace we can see the server returned payload too large –
![image](https://github.com/Azure/Azure-Sentinel/assets/62872235/30ab6524-99db-4d0b-a624-b7ec99c1ea86)
And we can see an incredibly huge query(about 1,690,000 characters) being sent in the post body. 
![image](https://github.com/Azure/Azure-Sentinel/assets/62872235/f71622a4-9708-49e6-ae1d-6dc1a931a350)

The IP Address filter gets it’s value sets by below query –
DnsEvents
| extend IpAddresses = iif(IPAddresses=="","empty" ,IPAddresses)
| summarize by IpAddresses
| sort by IpAddresses desc

And in the sub tabs, below query are being used –
DnsEvents
| extend IPAddresses = iif(IPAddresses=="", "empty", IPAddresses)
**| where "*" in ({IpAddresses}) or IPAddresses in ({IpAddresses})**
| summarize Events= count() by bin_at(TimeGenerated, 1h, now()), SubType

When “All” is selected, the **bold line** would inject all the IP addresses from the DnsEvents table into the query body and make the query unacceptable large.

v-rbajaj:
Hi @holi1-ops, can you please specify the name of the workbook or maybe link that workbook here, this will help us understand which specific workbook you are facing this issue. Thanks

holi1-ops:
> Hi @holi1-ops, can you please specify the name of the workbook or maybe link that workbook here, this will help us understand which specific workbook you are facing this issue. Thanks

This is the link to the workbook https://github.com/Azure/Azure-Sentinel/blob/master/Workbooks/Dns.json

v-rbajaj:
Hi @holi1-ops, we have made some modifications to the query please try this attached template and let us know if that solves the issue. 

[...]

PR #8647, opened 2023-07-27
Update Workbook Dns.json

rahul0216:
Change(s):
   - Updated "selectAllValue" value from empty string to "*".

   Reason for Change(s):
   - https://github.com/Azure/Azure-Sentinel/issues/8376
   - In case of empty string, Ip list was forming a large string and when used in query, it was exceeding the max query length.

   Version Updated:
   - No

   Testing Completed:
   - Yes, functional testing and deployment testing

### DIFF
--- a/Workbooks/Dns.json
+++ b/Workbooks/Dns.json
@@ -88,7 +88,7 @@
               "additionalResourceOptions": [
                 "value::all"
               ],
-              "selectAllValue": ""
+              "selectAllValue": "*"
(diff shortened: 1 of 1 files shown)


v-prasadboke:
Hello @rahul0216 Thank you for raising the PR, will be investigating this. You can expect an update till 31 Jul 2023.

vakohl (approved):

PR #8688, opened 2023-08-01
Added condition so that repo checkout doesn't happen

v-amolpatil:
   
   Change(s):
   - Currently checkout is ran which is not request so added condition so that it gets skipped.

   Reason for Change(s):
   - Currently checkout is ran which is not request so added condition so that it gets skipped.

   Version Updated:
   - NA

   Testing Completed:
   - Yes
   
   - Yes

Below is the output after testing on Sarath's private repo:
image


mkchiliveri (approved):
looks good

v-dvedak (approved):

PR #8729, opened 2023-08-07
Fix: on addition of package earlier commit were reset

v-amolpatil:
   
   Change(s):
   - Earlier we were using github action from marketplace of peter-evans/create-pull-request action to create and update the package with artifacts but we see that if the pr is existing it clears out all the earlier commits and just added the new package as a commit. It should not move all earlier commit but should add a new commit with package. In new changes we are using git command to add the package to existing pr.  

   Reason for Change(s):
   - Fix issue on commits cleared out for a pr when package is added using slash command.

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes


Tichandr (approved):
Looks Good

PR #8754, opened 2023-08-09
Update TargetResource logic for UserAssignedPrivilegedRole

eroll-chorus:
Change(s):
   - Condition to check if `TargetResource.type` is a User or Service Principal for `UserAssignedPrivilegedRole` (`User Assigned Privileged Role`)
   - Assignment of the `Target` variable based on whether the `TargetResource.type` is a User or Service Principal
        - update to a case-insensitive equality on the Service Principal to be consistent with the initial `type` filter

   Reason for Change(s):
   - The current logic of `where TargetResource.type =~ "User"` already filters the `type` by User, so the subsequent `extend Target = iff(TargetResources.type == "ServicePrincipal"` logic will never return `true`
   - `iff(TargetResources.type == "ServicePrincipal", tostring(TargetResources.displayName)` references `TargetResources` instead of `TargetResource`

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes

### DIFF
--- a/Solutions/Azure Active Directory/Analytic Rules/UserAssignedPrivilegedRole.yaml
+++ b/Solutions/Azure Active Directory/Analytic Rules/UserAssignedPrivilegedRole.yaml
@@ -26,9 +26,8 @@ query: |
   | where ActivityDisplayName has_any ("Add eligible member to role", "Add member to role")
   | mv-apply TargetResource = TargetResources on 
     (
-        where TargetResource.type =~ "User"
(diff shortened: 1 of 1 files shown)


v-atulyadav:
Hi @eroll-chorus,
Thank you for raising Pull Request with us! We will review the Pull Request and get back to you by 17 Aug 2023. Thanks

v-prasadboke:
Hello @eroll-chorus, Thank you for raising this PR. This PR will be investigated and you can expect an update till 18 Aug 2023,.

PR #8759, opened 2023-08-10
Make Parser FunctionParams as optional

v-amolpatil:
   
   Change(s):
   - When FunctionParams attribute is not present in yaml based parser then package creation tool fails with an error. So we have make this functionparams check optional i.e if not specified then we dont calculate it and keep it as blank.

   Reason for Change(s):
   - Make FunctionParams as optional check

   Version Updated:
   - NA

   Testing Completed:
   - Yes, did testing by deploying the create ui using custom deploy. Also verified if maintemplate is getting deployed successfully using custom deploy

   - Yes


mkchiliveri (approved):
Looks good

v-atulyadav (approved):

PR #8764, opened 2023-08-10
Fix Source IP aggregation for Firewall

ricardotolentino:
Change(s):
   - Fix source IP aggregation for Azure Firewall analytics rules "Multiple Sources Affected by the Same TI Destination"

   Reason for Change(s):
   - Undo a change from a recent commit (a7132b0) and restore the "Multiple Sources Affected by the Same TI Destination" Firewall analytics rule source IP aggregation. Currently the rule summarizes by the source IP, which results in the query returning zero results unless the "MinAffectedThreshold" is set to 0, thereby making the parameter useless.
   - Fix a typo for the "MinAffectedThreshold" parameter

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes

### DIFF
--- a/Solutions/Azure Firewall/Analytic Rules/Azure Firewall - Multiple Sources Affected by the Same TI Destination.yaml
+++ b/Solutions/Azure Firewall/Analytic Rules/Azure Firewall - Multiple Sources Affected by the Same TI Destination.yaml
@@ -26,7 +26,7 @@ query: |
   let RunTime = 1d; 
   let StartRunTime = 1d; 
   let EndRunTime = StartRunTime - RunTime; 
-  let MinAffectedThershold = 5; 
+  let MinAffectedThreshold = 5;
   union isfuzzy=true
   (AzureDiagnostics 
   | where TimeGenerated  between (ago(StartRunTime) .. ago(EndRunTime))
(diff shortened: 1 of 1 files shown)


v-prasadboke:
Hello @ricardotolentino, Thanks for raising this PR. I'll be investigating this PR and will update you before 16 Aug 2023.

v-prasadboke (dismissed):

Solutions/Azure Firewall/Analytic Rules/Azure Firewall - Multiple Sources Affected by the Same TI Destination.yaml
```diff
@@ -26,16 +26,16 @@ query: |
...
-  let MinAffectedThershold = 5; 
+  let MinAffectedThreshold = 5;
   union isfuzzy=true
   (AzureDiagnostics 
   | where TimeGenerated  between (ago(StartRunTime) .. ago(EndRunTime))
   | where OperationName == "AzureFirewallThreatIntelLog"
   | parse msg_s with * "from " SourceIp ":" SourcePort:int " to " Fqdn ":" DestinationPort:int  "." * "Action: Deny. " ThreatDescription),
   (AZFWThreatIntel
   | where TimeGenerated between (ago(StartRunTime) .. ago(EndRunTime)))
-  | summarize TiTrafficCount = count(), dCountSourceIps = dcount(SourceIp), AffectedIps = make_set(SourceIp, 10000) by Fqdn, ThreatDescription,SourceIp
```
v-prasadboke, line 37: Hello @ricardotolentino, here SourceIp is removed. any specific reason for this.

ricardotolentino: Hi @v-prasadboke.  Thanks for the feedback, you're right I missed that by removing the `SourceIp` aggregation then that affects its usage as an entity mapping.

How about instead of removing the entity mapping, I `mv-expand()` the `AffectedIps` array into a separate `SourceIp` column to restore the entity mapping?  The `MinAffectedThreshold` parameter would still work to filter the results.

v-prasadboke:
Hello @ricardotolentino, Thank you for the updates. Continuing to work on this.

v-prasadboke:
Hello @ricardotolentino, query looks good but due to lack of sample data actual results are missing. 
Can you please add sample data for query results.
Please add it to this path: https://github.com/Azure/Azure-Sentinel/tree/master/Sample%20Data/Custom

ricardotolentino:

v-prasadboke:
Thank you @ricardotolentino, for the sample data. I'll test the analytic for results. Thanks!

v-prasadboke (approved):

PR #8874, opened 2023-08-24
Skip Detection comment on fork PR for security reasons

v-amolpatil:
   
   Change(s):
   - In 'addCommentToRemindUpdatingTemplateVersion.yml' file we have added a condition to skip action run if the pr is from fork repo.
   
   Reason for Change(s):
   - When a fork pr is runs this action it fails as the token secret cannot be shared with the fork pr and can be a security risk so we are avoiding any job run for fork pr.

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes


mkchiliveri (approved):
Looks good

v-dvedak (approved):

PR #8888, opened 2023-08-25
Fix FQDN aggregation for Port Sweep Firewall rule

ricardotolentino:
Change(s):
   - Fix FQDN aggregation for Azure Firewall analytics rules "Port Sweep"

   Reason for Change(s):
   - Undo a change from a recent commit (a7132b0) and restore the "Port Sweep" Firewall analytics rule FQDN aggregation. Currently the rule summarizes by the FQDN, which results in the query returning zero results unless the "MinimumDifferentHostsThreshold" is set to 0, thereby making the parameter useless.  (This is similar to PR 8764)

   Version Updated:
   - Yes

   Testing Completed:
   - Yes.  Sample data is included within the file "AZFWPortSweepCustom_CL.csv"

   - Yes


v-rbajaj (approved):
Looks good.

v-dvedak (approved):

PR #8892, opened 2023-08-28
Compress only files instead of complete folder in package

v-amolpatil:
   
   Change(s):
   - Made change in "commonFunctions.ps1" file

   Reason for Change(s):
   - When a package i.e. zip is created then instead of zipping only createUiDefinition and mainTemplate it also adds old zip files in the package so this PR solves this issue.

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes

**Below is the zip package content of Azure Active Directory:**
image

image


mkchiliveri (approved):
Looks good

v-atulyadav (approved):

PR #8954, opened 2023-09-06
Description containing double quotes via image tag

v-amolpatil:
   
   Change(s):
   - When a image tag is added then it has double quotes and in our workflow it was failing with assigning values. So we have made it single quote. This happens when we use slash command

   Reason for Change(s):
   - See guidance below

   Version Updated:
   - See guidance below

   Testing Completed:
   - See guidance below

   - See guidance below


Tichandr (approved):
Looks Good

PR #9003, opened 2023-09-13
HttpHost LogicalType Fix ASimTester.csv

Detectioneer:
   
   Change(s):
   - Removed the hostname LogicalType at httpHost (websession)

   Reason for Change(s):
   - Removed the hostname LogicalType at httpHost (websession) as it was incorrect according to the documentation https://learn.microsoft.com/en-us/azure/sentinel/normalization-schema-web#:~:text=applicable.%0A%0AExample%3A%20800-,HttpHost,-Optional

   Version Updated:
   - No

   Testing Completed:
   - No

   - NA

### DIFF
--- a/ASIM/dev/ASimTester/ASimTester.csv
+++ b/ASIM/dev/ASimTester/ASimTester.csv
@@ -701,7 +701,7 @@ Hostname,string,Alias,WebSession,Hostname,,DstHostname
 HttpContentFormat,string,Optional,WebSession,,,
 HttpContentType,string,Optional,WebSession,,,
 HttpCookie,string,Optional,WebSession,,,
-HttpHost,string,Optional,WebSession,Hostname,,
+HttpHost,string,Optional,WebSession,,,
 HttpIsProxied,bool,Optional,WebSession,,,
 HttpReferrer,string,Optional,WebSession,,,
 HttpRequestBodyBytes,long,Optional,WebSession,,,


vakohl:
@Detectioneer the logical type for httpHost is Hostname, so in ASIMtester.csv its correct. I'll have it corrected in Docs

v-atulyadav:
Thanks @vakohl for clarifying, @Detectioner I hope the change you have proposed is clarified so we can close this case.

Detectioneer:
Hi @vakohl, Thanks for the clarification!

I understood the HttpHost field is typically based on the HTTP Host Header
According to the mozilla documentation linked this field can contain a FQDN for the server, example: developer.mozilla.org
(https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host)

However, developer.mozilla.org is not a valid hostname according to the Asim Datatester. 
It tests against the regex: `let Hostname_regex = @'^[a-zA-Z0-9-]{1,61}$';\r\n`  
Which doesn't seem to allow for periods in the hostname.  (https://github.com/Azure/Azure-Sentinel/blob/master/ASIM/dev/ASimTester/ASimDataTester.json)

How should I proceed with this issue ?

vakohl:
> Hi @vakohl, Thanks for the clarification!
> 
> I understood the HttpHost field is typically based on the HTTP Host Header According to the mozilla documentation linked this field can contain a FQDN for the server, example: developer.mozilla.org (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host)
> 
> However, developer.mozilla.org is not a valid hostname according to the Asim Datatester. It tests against the regex: `let Hostname_regex = @'^[a-zA-Z0-9-]{1,61}$';\r\n` Which doesn't seem to allow for periods in the hostname. (https://github.com/Azure/Azure-Sentinel/blob/master/ASIM/dev/ASimTester/ASimDataTester.json)
> 
> How should I proceed with this issue ?

Got it. Make sense. Thanks for sharing this example. @Detectioneer

vakohl (approved):

v-atulyadav (approved):

PR #9013, opened 2023-09-15
MainTemplate Generation should use ParserId instead of ParserName

v-amolpatil:
   
   Change(s):
   - Changes in CommonFunctions and V2 powershell script files

   Reason for Change(s):
   - In Content and Metadata DependOn is referring to parserName instead it should be parserId.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


Tichandr (approved):
looks good

PR #9019, opened 2023-09-15
Added python package with latest version of Cryptography.

jayeshprajapaticrest:
Change(s):
   - Added python packages in the data connector zip with the latest version of Cryptography package
   - Fixed session closed issue in data connector

   Reason for Change(s):
   - Python package folder is not available in the data connector
   - When exception occur in the mapping then it closes session and due to that indicators can't be created

   Version Updated:
   - 3.0.0

   Testing Completed:
   - Yes

   - Yes


v-prasadboke:
MicrosoftTeams-image (24)
MicrosoftTeams-image (23)

v-prasadboke (approved):

PR #9059, opened 2023-09-21
Arm-ttk slash command and customization

v-amolpatil:
   
   Change(s):
   - Made change to existing arm-ttk-validations file and moved common code to separate file. Also added new file to do slash command using arm-ttk. Slash command will run when anyone add comment like below :'/arm-ttk', '/ARM-TTK', '/Arm-Ttk', '/armttk' or '/ARMTTK'. Also did fix deprecated command from github.

   Reason for Change(s):
   - We need to support slash command using arm-ttk.

   Version Updated:
   - NA

   Testing Completed:
   - Yes on private repo

   - Yes

**Screenshot from private repo:**
image

### DIFF
--- a/.github/workflows/slash-command-armttk.yaml
+++ b/.github/workflows/slash-command-armttk.yaml
@@ -1,10 +1,14 @@
-name: Arm-ttk Validations
+# THIS WORKFLOW WILL RUN WHEN WE ADD SLASH COMMAND LIKE '/arm-ttk', '/ARM-TTK', '/Arm-Ttk', 'armttk' or 'ARMTTK'
+name: Slash Command ARM-TTK Tests
 
-on: [pull_request]
+on:
+  issue_comment:
+    types: [created, edited]
 
 jobs:
   run-arm-ttk:
     runs-on: ubuntu-latest
+    if: ${{ github.actor != 'dependabot[bot]' && !github.event.pull_request.head.repo.fork && github.event.issue.pull_request && (contains(github.event.comment.body, '/armttk') || contains(github.event.comment.body, '/Armttk') || contains(github.event.comment.body, '/ARM-TTK') || contains(github.event.comment.body, '/ARMTTK') || contains(github.event.comment.body, '/arm-ttk') || contains(github.event.comment.body, '/ArmTtk')) }}
     outputs:
       solutionName: ${{ steps.step1.outputs.solutionName }}
       mainTemplateChanged: ${{ steps.step1.outputs.mainTemplateChanged }}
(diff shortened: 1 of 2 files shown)


.github/workflows/slash-command-armttk.yaml
```diff
@@ -0,0 +1,44 @@
...
+name: Slash Command ARM-TTK Tests
+
+on:
+  issue_comment:
+    types: [created, edited]
+
+jobs:
+  run-arm-ttk:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor != 'dependabot[bot]' && !github.event.pull_request.head.repo.fork && github.event.issue.pull_request && (contains(github.event.comment.body, '/armttk') || contains(github.event.comment.body, '/Armttk') || contains(github.event.comment.body, '/ARM-TTK') || contains(github.event.comment.body, '/ARMTTK') || contains(github.event.comment.body, '/arm-ttk') || contains(github.event.comment.body, '/ArmTtk')) }}
```
Tichandr, line 11: can't we make this array of string items and check with contains

.script/package-automation/arm-ttk-tests.ps1
```diff
@@ -0,0 +1,57 @@
+
+try {
+    $diff = git diff --diff-filter=d --name-only HEAD^ HEAD
+    Write-Host "List of files in PR: $diff"
+
+    $hasmainTemplateChanged = $false
+    $hasCreateUiDefinitionTemplateChanged = $false
+
+    $isChangeInSolutionsFolder = [bool]($diff | Where-Object {$_ -like 'Solutions/*'})
```
Tichandr, line 9: What is this check for solutions folder?

v-amolpatil: It checks if there are changes in Solutions folder only. If yes then it goes and check if there are changes in maintemplate and/or createUiDefinition file.

.github/workflows/slash-command-armttk.yaml
v-amolpatil: updated code

.script/package-automation/arm-ttk-tests.ps1
Tichandr: I got that. why don't we directly check inside solutions/package etc..

v-amolpatil: Inside of Solutions we have different solution folders like Alibab Cloud, Cisco Umbrella etc and inside of those folders we have Package folder. We cannot directly get Package folder for a solution. Also we need Solution Name as well.

Tichandr (approved):
Looks good

PR #9085, opened 2023-09-25
Filter out exclusion list so that package is not created

v-amolpatil:
   
   Change(s):
   - Filter out exclusion list so that package is not generated unnecessarily.

   Reason for Change(s):
   - Filter out exclusion list so that package is not generated unnecessarily.

   Version Updated:
   - NA

   Testing Completed:
   - 

   - Yes

**Changes in PR:**
image

**It is not creating package when changes are from exclusion list**
image

**When PR contains only ReleaseNotes.md change then it will skip as shown in below screenshot:**
image


Tichandr (dismissed):
Looks good

Tichandr (approved):
looks good

PR #9093, opened 2023-09-26
Updating entity mappings to remove legacy support and map additional entities - Set 1

shainw:
Change(s):
   - Removed *CustomEntity
   - Mapped fields directly for entities
   - Parsed some fields to allow for proper entity mapping to occur

   Reason for Change(s):
   - Improving correlation capabilities across incidents, alerts and raw events.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes

### DIFF
--- a/Detections/LAQueryLogs/UserSearchingForVIPUserActivity.yaml
+++ b/Detections/LAQueryLogs/UserSearchingForVIPUserActivity.yaml
@@ -26,18 +26,19 @@ query: |
   | where QueryText has_any (vips) or QueryText has_any ('_GetWatchlist("VIPUsers")', "_GetWatchlist('VIPUsers')")
   | where AADEmail !in (allowed_users)
   | project TimeGenerated, AADEmail, RequestClientApp, QueryText, ResponseRowCount, RequestTarget
-  | extend timestamp = TimeGenerated, AccountCustomEntity = AADEmail
-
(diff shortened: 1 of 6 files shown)


Detections/LAQueryLogs/UserSearchingForVIPUserActivity.yaml
```diff
@@ -26,24 +26,25 @@ query: |
...
-      - identifier: FullName
-        columnName: AccountCustomEntity
+      - identifier: Name
+        columnName: AccountName
+      - identifier: UPNSuffix
+        columnName: AccountUPNSuffix
   - entityType: AzureResource
     fieldMappings:
       - identifier: ResourceId
         columnName: RequestTarget
```
ashwin-patil, line 40: Existing change looks good. I noticed mapping for resourceId and was wondering should it be mapped to _ResourceId column ? we may have to project it in that case.

ashwin-patil (approved):
1 Unrelated suggestion to mapping in last query. other than changes looks good.

Detections/LAQueryLogs/UserSearchingForVIPUserActivity.yaml
shainw: We will leave as is to make sure the intent of the query is maintained and no assumptions are made by fixing the mapping.

https://learn.microsoft.com/en-us/azure/azure-monitor/reference/tables/laquerylogs

<html>
<body>


RequestTarget | string | ResourceId of the request URL.
-- | -- | --
_ResourceId | string | A unique identifier for the resource that the record is associated with


</body>
</html>

PR #9206, opened 2023-10-12
MimecastSEG: Updated publisherId

nipun-crestdatasystem:
Change(s):
   - Updated publisherId

   Reason for Change(s):
   - Customer has published an offer with different publisherId

   Version Updated:
   - N/A

   Testing Completed:
   - Yes

   - Yes


v-dvedak (approved):

v-dvedak:
Approved considering the publisher ID for this solution is same as Mimecast Cloud Gateway Audit.

PR #9227, opened 2023-10-17
Fix empty array issue and json valid or not after package

v-amolpatil:
   
   Change(s):
   - Fix empty error issue on playbook
   - added json valid or not check in tool
   - minor change incorporated for Rahul in V3 tool line 70

   Reason for Change(s):
   - Fix empty error issue on playbook

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes

**Testing results:**
**1.Source playbook file having null value:**
image

**After generating maintemplate:**
image

**After deployment of playbook to Azure Portal:**
image

**Check the code for actual replaced value:**
image

**2.When playbook file has empty object as value. After deployment we see below: **
image

**Check the code for actual replaced value:**
image

### DIFF
--- a/Tools/Create-Azure-Sentinel-Solution/V3/createSolutionV3.ps1
+++ b/Tools/Create-Azure-Sentinel-Solution/V3/createSolutionV3.ps1
@@ -67,8 +67,11 @@ Write-Host "SolutionBasePath is $solutionBasePath, Solution Name $solutionName"
 
 $isPipelineRun = $false
 
-. "$repositoryBasePath/Tools/Create-Azure-Sentinel-Solution/common/commonFunctions.ps1" # load common functions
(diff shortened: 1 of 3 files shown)


manishkumar1991:
Reviewing

manishkumar1991:
Observation 

Playbook before template packaging 
image

Playbook deployed through template spec post packaging :
image

Both the screenshots are matching hence script has generated the desired and expected output.

@rahul0216  : Do you think of any other use case of null value or empty assignment of variable, which we need to test

rahul0216:
@manishkumar1991 Looks good to me. Both Empty and null assignment test cases look good.

manishkumar1991 (approved):
approved

rahul0216 (approved):
Changes look good.

v-atulyadav (approved):

PR #9283, opened 2023-10-27
Fix categories in Packaging tool

v-amolpatil:
   
   Change(s):
   - In SolutionMetadata.json file we have categories attribute array which contains domains and verticals. Current code was in V2 as well and it fails when we have both domains and verticals specified in this file. 
   - Currently, when we try to create a package it fails with an error as the property categories already exist in psobject. So I have moved code out in this PR so that we don't get error specified earlier.

   Reason for Change(s):
   - Identified issue for categories which is present in V2 and V3 packaging tool.

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes


rahul0216 (approved):
Changes look good.

v-atulyadav (approved):

PR #9289, opened 2023-10-28
Validate Workbook SerializedData in MainTemplate

v-amolpatil:
   
   Change(s):
   - In mainTemplate.json if there are any workbook which has attribute "serializedData" then workbook content is coming as a string in this field. It is observed that this "serializedData" attribute field value i.e json is invalid and fails when we open workbook view template. So this PR will validate if the "serializedData" attribute value is a valid json or not.

   Reason for Change(s):
   - To Validate if workbook, "serializedData" attribute value is a valid json format or not.

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes

**Testing Screenshot:**
**When mainTemplate has invalid json content in serializedData attribute as shown below:**
image
image

**When mainTemplate has valid json content in serializedData attribute as shown below it will be success:**
image

### DIFF
--- a/.script/jsonFileValidator.ts
+++ b/.script/jsonFileValidator.ts
@@ -12,7 +12,7 @@ export async function IsValidJsonFile(filePath: string): Promise<ExitCode> {
     if (fileContent.includes('serializedData')) {
       let mainTemplateJsonObj = JSON.parse(fileContent);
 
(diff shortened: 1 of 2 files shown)


v-amolpatil:
After discussion with Sanmit this PR is not need as we have taken care in PR https://github.com/Azure/Azure-Sentinel/pull/9135

PR #9299, opened 2023-10-30
Singleton orchestration pattern

wolyslager:
Required items, please complete
   
   Change(s):
   - Implemented singleton orchestration pattern as per https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-singletons?tabs=python. 
   - Added more logs

   Reason for Change(s):
   - We are seeing a bug where concurrent instances of the orchestration modify the SoarDatetimeEntity's state simultaneously, and results in undesired behavior. We rely on the state of this entity to determine the `startTime` filter parameter for API requests that are made to our REST API. Upon investigating our internal logs we can see that the `startTime` is not being set properly. A scenario where this becomes evident is the following: SentinelTimerTrigger triggers at T0, SentinelFunctionsOrchestrator (A) is kicked off at T1 (request takes 5 minutes due to timing out), SentinelTimerTrigger triggers at T5 (its on a 5 minute chron schedule), SentinelFunctionsOrchestrator (B) is kicked off at T6  ... here we can see the problem. SentinelFunctionsOrchestrator (A) has not finished executing and as a result has not updated the SoarDatetimeEntity with the appropriate date, so SentinelFunctionsOrchestrator (B) will be reading from a stale state. This change aims to ensure that only 1 instance of the orchestrator can run from start to finish to avoid multiple orchestrators modifying the SoardatetimeEntity's state. 

   Version Updated:
   -No

   Testing Completed:
   - Deployed functions to Azure environment and inspected logs to verify that only 1 instance of the orchestration was running at any given time, and verified that the SentinelTimerTrigger skipped the creation of new orchestrations when there was one already running.  

   Checked that the validations are passing and have addressed any issues that are present:
   - See guidance below

# Guidance <- remove section before submitting
-----------------------------------------------------------------------------------------------------------
## Before submitting this PR please ensure that you have read the following sections and filled out the changes, reason for change and testing complete sections:

Thank you for your contribution to the Microsoft Sentinel Github repo.

> Details of the code changes in your submitted PR.  Providing descriptions for pull requests ensures there is context to changes being made and greatly enhances the code review process.  Providing associated Issues that this resolves also easily connects the reason.
   
   Change(s):
   - Updated syntax for XYZ.yaml

   Reason for Change(s):
   - New schema used for XYZ.yaml
   - Resolves ISSUE #1234

   Version updated:
   - Yes
   - Detections/Analytic Rule templates are required to have the version updated

[...]

PR #9378, opened 2023-11-13
Function update

wolyslager:
NOTE: This pr is a[ copy of this one](https://github.com/Azure/Azure-Sentinel/pull/9299), which went bad for an unknown reason. 

Change(s):

Implemented singleton orchestration pattern as per https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-singletons?tabs=python.
Added more logs
Parametrized the API host value
Added auto-recovery mechanism when API request encounters too many results. In order to avoid hitting the 5 minute timeout, the query filter is automatically adjusted to fetch less items.
Changed the durable entity to hold a dictionary as state rather than a string.
Reason for Change(s):

We are seeing a bug where concurrent instances of the orchestration modify the SoarDatetimeEntity's state simultaneously, and results in undesired behavior. We rely on the state of this entity to determine the startTime filter parameter for API requests that are made to our REST API. Upon investigating our internal logs we can see that the startTime is not being set properly. A scenario where this becomes evident is the following: SentinelTimerTrigger triggers at T0, SentinelFunctionsOrchestrator (A) is kicked off at T1 (request takes 5 minutes due to timing out), SentinelTimerTrigger triggers at T5 (its on a 5 minute chron schedule), SentinelFunctionsOrchestrator (B) is kicked off at T6 ... here we can see the problem. SentinelFunctionsOrchestrator (A) has not finished executing and as a result has not updated the SoarDatetimeEntity with the appropriate date, so SentinelFunctionsOrchestrator (B) will be reading from a stale state. This change aims to ensure that only 1 instance of the orchestrator can run from start to finish to avoid multiple orchestrators modifying the SoardatetimeEntity's state.
Additionally, we want to avoid timeouts. When we make a top level API request to fetch threats, we get a response with threat ids. For each threat_id, we make a new API request. If the initial request results in too many ids (we've determined this heuristic based on latency tests) then the function will inevitably time out, and the following execution of it will only contain more ids. For this reason, we've implemented an "auto-recovery" mechanism by which the function always aims to stay under a certain number of API calls, this value is parametrized and can be set in the function config variables.
Version Updated:
-No

Testing Completed:

Deployed functions to Azure environment and inspected logs to verify that only 1 instance of the orchestration was running at any given time, and verified that the SentinelTimerTrigger skipped the creation of new orchestrations when there was one already running.
Tested that the auto-recovery mechanism works as expected when the function identifies that it will time out based ion the heuristic MAX_NUMBER_OF_THREATS.

See guidance below


wolyslager:
image

v-prasadboke:
Hello @wolyslager , Thanks for raising this PR.
This PR will be investigated and will provide you some updates before 23 November, 2023.

v-prasadboke (approved):

v-atulyadav (approved):

PR #9388, opened 2023-11-14
Disable GitHub Workflow Call from Azure pipeline which creates Auto Package

v-amolpatil:
   
   Change(s):
   - From azurepipeline.yaml file removed callGithubWorkflow file, so that it will not run auto packaging.

   Reason for Change(s):
   - We are disabling auto packaging for now and we will be rethinking on whether to move it after merge of the pr. As when package is added and if there reviewer approves the pr but auto package added commit to the same pr then it dismisses the pr approval by saying stale review discussed. 

   Version Updated:
   - See guidance below

   Testing Completed:
   - Yes
   
   - Yes


Tichandr (approved):
Looks good. Just disabling the server-side package creation for now. because after PR merge we have the package creation enabled and using the slash command from client side also enabled

PR #9401, opened 2023-11-15
Multiple to single object variables object

v-amolpatil:
   
   Change(s):
   - Multiple variables to single variables inside of object. Also removed circular dependency issue. This is applicable for parsers, analytic rules and hunting queries

   Reason for Change(s):
   - Multiple variables to single variables inside of object. Also removed circular dependency issue. This is applicable for parsers, analytic rules and hunting queries

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes

Below is for corelight deployment as a custom deployment:
image


Tichandr (approved):
looks fine

PR #9409, opened 2023-11-16
Corelight package updated to 3.0.1

v-amolpatil:
   
   Change(s):
   - We have updated tool to create corelight package with string-object mapping in variables and added references where ever it is used. Earlier package was rejected from content hub team so based on the tool change and removed circular dependency and created a new package for corelight.

   Reason for Change(s):
   - Update package using new way of string-object mapping as content hub team rejected previous package and they have made change at their end for string-object mapping.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


Solutions/Corelight/Hunting Queries/CorelightRarePOST.yaml
```diff
@@ -17,7 +17,7 @@ query: |
   corelight_http
   | where TimeGenerated > ago(24h)
   | where method in~ ('POST', 'PUT')
-  | where status_code >= 400 and status_code < 500
+  | where toint(status_code) >= 400 and toint(status_code) < 500
```
Tichandr, line 20: why did you changed this? may be the data is coming for status code as int only?

Tichandr (approved):
Looks Good

Solutions/Corelight/Hunting Queries/CorelightRarePOST.yaml
v-amolpatil: as per discussion we will keep this.

PR #9414, opened 2023-11-17
Possible error in copy paste action in detection AWS_CreatedCURDLambdaPolicytoPrivilegEscalation.yaml

ep3p:
I don't know if "lambda:Delete" is a valid action, please, could you check with the original author?   

   Change(s):
   - Change ```kms``` by ```lambda```

   Reason for Change(s):
   - I believe KMS and Lambda might not be related for this detection, and there was a mistake copypasting a previous detection.

   Version Updated:
   - Yes

   Testing Completed:
   - I don't know if "lambda:Delete" is a valid action, please, could you check with the original author?

   - Yes


ep3p:
It seems to be Delete actions in lambda, but I still think the original author should be consulted.

image

ep3p:
> Hi @ep3p, have you tested the rule after making these changes?

@v-rbajaj The query is valid and the changes makes sense. But this type of error is a detection logic error. The only way to solve this is asking the original author what was the original intention of this detection, if maybe they commited a mistake copypasting.

v-rbajaj (approved):

PR #9431, opened 2023-11-21
AnalyticRules Entity Mapping Updates

ddamenova:
Change(s):
   - Primarily breaking up Account entity mappings into AccountName and 
     UPNSuffix
- Getting rid of extraneous variable names for entities columns such as IP

   Reason for Change(s):
   - Updating entity mapping for increased alert correlation

   Version Updated:
   - yes

   Testing Completed:
   - yes

   - no


shainw (approved):
Looks good!!

PR #9510, opened 2023-11-29
GithubWebhookConnector now passes "event" from GitHub webhook to Sentinel

PCNZ:
   
   Change(s):
   - Updated python script in GithubWebhookConnector to capture the event name which is provided in the webhook x-github-event header and pass the event name through to Sentinel.

   Reason for Change(s):
   - The event name is required when sending multiple event types to Sentinel using the webhook data connector. Event name combined with action are key to identifying the type of event which was received.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes new "event" value is now parsed to Sentinel from Github webhook

   - Need help


PCNZ:

v-prasadboke (approved):

PR #9518, opened 2023-11-30
Fix Cloudflare Performance Issue and Bug Causing Logjams

aaronbeuhring:
Change(s):
   - updated function.json, main.py, and azuredeploy_Cloudflare_API_FunctionApp.json

   Reason for Change(s):
   - The cloudflare function app does not perform well with production logs. Cloudflare logpush sends log files that are often larger than 5MB. The defaults in the function app only download the first 2MB. When it fails to process the full log file it then reprocesses it over and over again. Eventually, the max number of files to be processed concurrently will only be failed log files and no new logs flow in. The changes in these files add configurable properties for max chunk size and run time to the function app. The main.py has been modified to utilize more sane defaults and also to avoid attempting to process more than one file at a time as the function app does not handle parallel processing of large files well. Additional functionality has been added to tag log files when they are processed and check for the tag on subsequent runs so as to skip log files that previously failed to be processed and avoid a log jam.  The cron job has been set to run every 3 minutes. If the function app runs longer than 3 minutes it will just continue to run. The max run time has been set to the default for consumption function app runs at 10 minutes. The function will check if it has been running for 7.5 minutes and stop picking up new files to process so it can complete before 10 minutes and forced termination. In my experience no large log files have taken more than a minute to process and frequently far less. More frequent runs should also avoid backlogs.

   Version Updated:
   - No

   Testing Completed:
   - Yes

   - Yes


v-prasadboke:
Hello @aaronbeuhring, Thanks for raising this PR. This PR will be investigated and will update you about the same before 06 December, 2023.

aaronbeuhring:
image
The error count in the following screenshot is from before the changes were made.
image

Solutions/Cloudflare/Data Connectors/azuredeploy_Cloudflare_API_FunctionApp.json
```diff
@@ -168,7 +168,9 @@
                         "AzureWebJobsStorage": "[concat('DefaultEndpointsProtocol=https;AccountName=', toLower(variables('FunctionName')),';AccountKey=',listKeys(resourceId('Microsoft.Storage/storageAccounts', toLower(variables('FunctionName'))), '2019-06-01').keys[0].value, ';EndpointSuffix=',toLower(variables('StorageSuffix')))]",                      
                         "WORKSPACE_ID": "[parameters('AzureSentinelWorkspaceId')]",
                         "SHARED_KEY": "[parameters('AzureSentinelSharedKey')]",
+                        "MAX_SCRIPT_EXEC_TIME_MINUTES": "[parameters('MaxRunTime')]",
```
v-prasadboke, line 171: Hello @aaronbeuhring, you have not declared the parameters in the global parameters of the function app
Please declare **MaxRunTime** && **MaximumStorageBlobSize**

aaronbeuhring: I'm not sure what you are asking for. The parameters I added to this JSON were not originally in this at all. I followed the same format for the other entries in the properties section. The defaults are set in line 20 and 42 of main.py. ex: MAX_SCRIPT_EXEC_TIME_MINUTES = int(os.environ.get('MAX_SCRIPT_EXEC_TIME_MINUTES', 10)) and MAX_CHUNK_SIZE_MB = int(os.environ.get('MAX_CHUNK_SIZE_MB', 500))

v-prasadboke (approved):

v-atulyadav (approved):

v-sudkharat:
Hi @aaronbeuhring, Hope you are doing well. we got a GitHub regarding this PR. In which customer facing issue while deploying the function app - https://github.com/Azure/Azure-Sentinel/issues/9601
It may, because of recent changes in function app. Could you please check on this? Thanks!

aaronbeuhring:
> Hi @aaronbeuhring, Hope you are doing well. we got a GitHub regarding this PR. In which customer facing issue while deploying the function app - #9601 It may, because of recent changes in function app. Could you please check on this? Thanks!

Check https://github.com/Azure/Azure-Sentinel/pull/9614 There may have been additional parameters that needed to be added to azuredeploy_Cloudflare_API_FunctionApp.json to account for the additional properties added. I could only test on a production function app where I added application settings for these variables to the existing properties. If that doesn't resolve the issue then I suggest working with Cloudflare as they are the provider who should be supporting this solution.

PR #9553, opened 2023-12-06
Entity Work #4

mmelndezlujn:
Change(s):
   - Updated mapping for two detections

   Reason for Change(s):
   - To get better correlations with other detections

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - No


shainw:
Added these to #9594, so closing this.

PR #9598, opened 2023-12-12
Added permission: none for Slash command dispatch

v-amolpatil:
   
   Change(s):
   - Added permission: none for slash command dispatch.

   Reason for Change(s):
   - Added permission: none for slash command dispatch.
   
   Version Updated:
   - NA

   Testing Completed:
   - NA

   - Yes

As we are using github apps there will be no user pat token and due to this we are using permission none so that the slash command will get triggered. 
image


Tichandr (approved):
Looks good

PR #9648, opened 2023-12-21
Images for backwards compatibility with old versions of playbooks.

RecordedFutureOskbo:
   
   Change(s):
   - Keep images in old location for backwards compatibility
   
   Reason for Change(s):
   - Found in content hub pipeline

   Testing Completed:
   - Yes

   - Yes


RecordedFutureOskbo:
@v-prasadboke, @v-atulyadav this is a small fix, where we had references to images in the playbook description. We will keep the images in old position so we don't break old template descriptions.
 
image

v-atulyadav (approved):

PR #9707, opened 2024-01-03
Prisma Cloud CWPP solution type change for password

v-amolpatil:
   
   Change(s):
   - For password type we need to change its type from "string" to "securestring".

   Reason for Change(s):
   - Currently password text is a plain text instead of securestring, so we updated the package and its version changed to 3.0.2.

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes

### **Checklist for this is not required as we are changing its type.**

Testing document attached below.

[Prisma Cloud CWPP.docx](https://github.com/Azure/Azure-Sentinel/files/13828946/Prisma.Cloud.CWPP.docx)


v-mchatla (approved):
changes look good

v-atulyadav (approved):

PR #9717, opened 2024-01-04
Adding new fields to ASIM Schema

vakohl:
   
   Change(s):
   - Adding new ASIM Fields
   - TargetOriginalAppType
   - ActingOriginalAppType
   - OriginalObjectType

   Reason for Change(s):
   - Adding new fields

   Version Updated:
   - NA

   Testing Completed:
   - NA

   - Yes

### DIFF
--- a/ASIM/schemas/entities/ASimApp.yaml
+++ b/ASIM/schemas/entities/ASimApp.yaml
@@ -18,6 +18,12 @@ Fields:
   Follows: <<Role>>AppName
   Description: The type of the application.
 
+- Name: <<Role>>OriginalAppType
+  Type: string
+  Class: Optional
+  Description: The application type as reported by the reporting device.
+  For roles: [ Target, Acting ]
+
 - Name: <<Role>>Url
   Type: string
   Class: Optional
--- a/ASIM/schemas/ASimAuditEvent.yaml
+++ b/ASIM/schemas/ASimAuditEvent.yaml
@@ -116,6 +116,11 @@ Fields:
   Description:
   Related to: Object
 
+- Name: OriginalObjectType
+  Type: string
+  Class: Optional
+  Description: The object type as reported by the reporting device.
+
 - Name: OldValue
   Type: string
   Class: Optional


ASIM/dev/ASimTester/ASimTester.csv
```diff
@@ -10,6 +10,9 @@ ActingAppType,string,Optional,AuditEvent,Enumerated,Process|Service|Resource|URL
 ActingAppType,string,Optional,Authentication,Enumerated,Process|Service|Resource|URL|SaaS application|CSP|Other,
 ActingAppType,string,Optional,FileEvent,Enumerated,Process|Service|Resource|URL|SaaS application|CSP|Other,
 ActingAppType,string,Optional,UserManagement,Enumerated,Process|Service|Resource|URL|SaaS application|CSP|Other,
+ActingOriginalAppType,string,Optional,AuditEvent,,,
+ActingOriginalAppType,string,Optional,UserManagement,,,
```
anki-narravula, line 14: @vakohl - As discussed, please generate AsimTester.csv through script by reading the original schemas and dependencies.

ASIM/schemas/entities/ASimApp.yaml
```diff
@@ -18,6 +18,12 @@ Fields:
   Follows: <<Role>>AppName
   Description: The type of the application.
 
+- Name: <<Role>>OriginalAppType
+  Type: string
+  Class: Optional
```
anki-narravula, line 23: I don't see we have Role defined for other fields  such as AppName, AppType in the same Entity, I feel OriginalAppType tied with AppName and App Type then why << Role >> is relevant only for this. Can you please clarify?

ASIM/dev/ASimTester/ASimTester.csv
vakohl: @anki-narravula done. Shared results with you.

ASIM/schemas/entities/ASimApp.yaml
vakohl: @anki-narravula 
yes, removed it as other related fields are not having it.

PR #9719, opened 2024-01-04
Entity Updates Jan 4 Diana Half

ddamenova:
Change(s):
   - Adding full name entity mappings back to detections 
   - Splitting account full names mappings into UPN Suffix and name (but also keeping full name mapping as well) 

   Reason for Change(s):
   - Increasing entity mappings helps w alert correlations

   Version Updated:
   - yes

   Testing Completed:
   - yes

   - not yet


v-prasadboke:
Hello @ddamenova, Please try to resolve Validation errors

shainw (approved):

PR #9729, opened 2024-01-06
Updating validation for entityMapping max from 5 to 10

shainw:
   
   Change(s):
   - Updated entityMappingsMaxLength: 10 from 5

   Reason for Change(s):
   - 10 entity mappings is supported now, so the validation is failing when it should not - https://learn.microsoft.com/en-us/azure/sentinel/sentinel-service-limits


v-prasadboke:
Hello @shainw, Please let us know if there's someone who would be reviewing this PR so that we act according to the same. 
Thanks.

v-prasadboke (approved):

PR #9789, opened 2024-01-19
Analytic Rule "requiredDataConnectors" property should be [] in maintemplate and handle ARM-TTK Error

v-amolpatil:
   
   Change(s):
   - For analytic rules we are changing "requiredDataConnectors" value if [] in source then in maintemplate also it should come as [] without making use of json[].
   - Also, added skip validation for armttk in Github pipeline to skip if we get "Template should not contain blanks" error for "requiredDataConnectors" property. -- reverted code for this as pr for arm-ttk is approved and is merged so we dont need this.

   Reason for Change(s):
   - Specified above

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes

Maintemplate for Apache Log4j Vulnerability Detection solution has source file "requiredDataConnectors" value as [] which is the expected output as shown below:
image

### DIFF
--- a/.github/actions/entrypoint.ps1
+++ b/.github/actions/entrypoint.ps1
@@ -13,6 +13,12 @@ if ($mainTemplateChanged -eq $true)
     # SKIP ANY ERRORS ON contentProductId AND id 
     $filterTestResults = New-Object System.Collections.ArrayList
     $hasContentProductIdError = $false
+    $hasRequiredDataConnectorsError = $false
+
+    $mainTemplateFilePath = ($PackageFolderPath + "/mainTemplate.json").Replace("//", "/")
+    $mainTemplateContent = Get-Content "$mainTemplateFilePath"
+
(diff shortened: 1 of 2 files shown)


.github/actions/entrypoint.ps1
```diff
@@ -30,13 +36,38 @@ if ($mainTemplateChanged -eq $true)
                 }
             }
         }
+        elseif($testInfo.Name -eq 'Template Should Not Contain Blanks')
```
Tichandr, line 39: we need to remove this once arm-ttk pr gets merged https://github.com/Azure/arm-ttk/pull/760

Tichandr (dismissed):

.github/actions/entrypoint.ps1
v-amolpatil: Removed code as our one arm-ttk repo pr is now merged.

daxesht (approved):

PR #9893, opened 2024-02-05
Improvements as per Github new change

v-amolpatil:
   
   Change(s):
   - As per new Github change in actions run, all actions with run in read mode. So instead of github token we are using github apps for actions which are trying to add comment or add label in PR, issue.
   - Updated version for github actions on github-script and github-checkout to use new version i.e commitid. Currently, we are getting warning for Node16 is deprecated and so we are updating this change.

   Reason for Change(s):
   - Stated above

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes

**Testing screentshot from private repo:**
For issue comment:
image
For label:
image

### DIFF
--- a/.github/workflows/AddLabel.yaml
+++ b/.github/workflows/AddLabel.yaml
@@ -8,6 +8,10 @@ on:
     paths:
       - Solutions/**
 
+env:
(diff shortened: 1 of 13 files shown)


Tichandr (approved):
Looks Good

PR #9935, opened 2024-02-09
Fix for analytic pipeline comment

v-amolpatil:
   
   Change(s):
   - For analytic rule comment is not adding because worflow secrets should be inherited. secrets inherit allows inner workflows to use those values.

   Reason for Change(s):
   - There are minor change

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes

**Testing:**
image


Tichandr (approved):
Ok

PR #9947, opened 2024-02-13
Update azuredeploy.json for RecordedFuture-ThreatIntelligenceImport to conform with request body required

jusso-dev:
The current API spec expects an array of "indicators" and string of "sourcesystem" params, currently this is passing the value of STIX indicators as a named array "indicators" but needs to be "value" as per the spec - https://learn.microsoft.com/en-us/azure/sentinel/upload-indicators-api#sample-request-body

   Change(s):
   - Updated RecordedFuture-ThreatIntelligenceImport azuredeploy.json to meet API spec
   
   Reason for Change(s):
   - In the current state this azuredeploy.json for the RecordedFuture-ThreatIntelligenceImport Azure Logic App does not work and returns a 400 bad request as the request JSON schema is incorrect as per the API spec - https://learn.microsoft.com/en-us/azure/sentinel/upload-indicators-api#sample-request-body

   Version Updated:
   - See guidance below

   Testing Completed:
   - Deploy azuredeploy.json and validate change
   
   - I have successfully deployed with this change and it resolved the 400 bad request errors


v-prasadboke:
Thanks for raising the PR, We will investigate this PR and update you about the same before 15 February, 2024.

v-prasadboke:
Hello @jusso-dev, Please accept license agreement

jusso-dev:
> @microsoft-github-policy-service agree [company="{your company}"]

@microsoft-github-policy-service agree [company="N/A"]

jusso-dev:

v-atulyadav (approved):

jusso-dev:
@RecordedFutureOskbo is it possible the connector versions are different targeting a different API version under the covers? I have this deployed in my clients tenant and I'm getting the opposite of that error when I try and use "indicators" as the named array.

Secondary, the current api spec suggest "value" is the current accepted named array parameter, which makes me further believe there is an API version mismatch happening somewhere.

RecordedFutureOskbo:
> @RecordedFutureOskbo is it possible the connector versions are different targeting a different API version under the covers? I have this deployed in my clients tenant and I'm getting the opposite of that error when I try and use "indicators" as the named array.
> 
> Secondly, the current api spec suggest "value" is the current accepted named array parameter, which makes me further believe there is an API version mismatch happening somewhere.

We have seen differences using different datacenters? I'm testing this in west-europe and east-us.

RecordedFutureOskbo:
@jusso-dev are you using the deprecated version? 
image

Also, the documentation on the logic app does not specify the data format at all 
https://learn.microsoft.com/en-us/connectors/azuresentinel/#threat-intelligence---upload-indicators-of-compromise-(v2)-(preview) (╯°□°）╯︵ ┻━┻

jusso-dev:
@RecordedFutureOskbo no I was using the latest version available at - Solutions/Recorded Future/Playbooks/IndicatorImport/RecordedFuture-ThreatIntelligenceImport/azuredeploy.json

Here's the current working, deployed Logic App:
image

The API doco I was referring to specifies that the "value" named array parameter is the required named array parameter to be passed - https://learn.microsoft.com/en-us/azure/sentinel/upload-indicators-api#sample-request-body

jusso-dev:
> > @RecordedFutureOskbo is it possible the connector versions are different targeting a different API version under the covers? I have this deployed in my clients tenant and I'm getting the opposite of that error when I try and use "indicators" as the named array.
> > Secondly, the current api spec suggest "value" is the current accepted named array parameter, which makes me further believe there is an API version mismatch happening somewhere.
> 
> We have seen differences using different datacenters? I'm testing this in west-europe and east-us.

We are using Australia East and it's working with "value" as a named array parameter.

RecordedFutureOskbo:
This is all very strange, I reached out to Microsoft and this is the response I got back.

> When you use the V2 action, it requires a "sourcesystem" and an "indicators" array, whereas when you use the older "Deprecated" action, it requires "sourcesystem" and a "value" array in the request body. Please let me know if this change is still causing you errors.

 I also created a logic app in Australia East and it requested indicators
 
image


Are you using standard or consumption logic apps? I run consumption when testing.

jusso-dev:
That is strange, I'm definitely using the current connector action and it's a consumption Logic App in Australia East, I get the above error if I change it to "indicators" so I'm at a loss what's causing this. The target LAW is in Aus Central I don't know if that is related to this issue?

RecordedFutureOskbo:
Message from Microsoft Sentinel team:


I found the bug, it was an api endpoint mismatch only in the Australia regions. The fix is checked in, and it should be deployed on Monday! You should be able to use the logic app actions as expected in all other regions.

 Still working on updating the api spec.

jusso-dev:
@RecordedFutureOskbo excellent! Thank you for following up on this and finding a fix 😊

RecordedFutureOskbo:
Great, thanks @jusso-dev I'm afk until Tuesday.

jusso-dev:
Hi @v-prasadboke and @v-atulyadav, any ETA on getting the API spec updated? This is causing a lot of confusion.

RecordedFutureOskbo:
I got the following message about the spec @jusso-dev. Some confusion can depend on we are not using the API directly. We are using the logic-app provided in  power platform -> https://learn.microsoft.com/en-us/connectors/azuresentinel/#threat-intelligence---upload-indicators-of-compromise-(v2)-(preview) where the spec is lacking format. 

So the API spec is probably correct if you use the API directly via CURL or other request libs.

// Oskar

jusso-dev:
Thanks Oskar that makes sense, I think some clarification even if it's a sentence or two would greatly improve the documentation.

austinmccollum:
updated the upload indicators api doc to reflect the different endpoint and array name depending on the logic app.

PR #9974, opened 2024-02-19
Ccp version change, auto release notes and update condition

v-amolpatil:
   
   Change(s):
   - Updated CCP data connector version instead of solutionversion.
   - Added automatic release notes on package creation
   - updated condition && with -and as the it is not valid condition in powershell. This change is for cef based where it is coming as using azure function in display name.

   Reason for Change(s):
   - Specified above

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes

After package for CEF based connector it will shown like below:
image

Release notes are added like below automatically:
image


Tichandr (approved):
Looks good

PR #10027, opened 2024-02-23
Update MS-PIM reference

RafaelRuales:
   
   Change(s):
   - Updated the entity that references MS-PIM

   Reason for Change(s):
   - the current  entity "InitiatedByDisplayName" does not exist

   Version Updated:
   - No
 
   Testing Completed:
   - Yes

   - No


v-rusraut:
Hi @RafaelRuales,
have you completed your changes or still you are working on this? As I see you have updated one line in query but still that line is commented.

image

Thanks

RafaelRuales:
Good morning @v-rusraut , I left it commented out as I figured it would be up to the individual to determine whether or not they want to see PIM activity. With that being said, the entity "InitiatedByDisplayName" was giving me an error in the query, I figured that the right entity if one was to uncomment that line was "InitiatingAppName" which is referenced earlier in the query, as opposed to "InitiatedByDisplayName" which doesn't seem to be. Does that make sense?

v-rusraut (dismissed):

v-atulyadav (dismissed):

v-rusraut (approved):

PR #10081, opened 2024-03-04
Added validation for parameter type check

v-amolpatil:
   
   Change(s):
   - This validation is added to check if there are any parameters in maintemplate that need to have type as "securestring" instead of string. eg: Password, ClientSecret, AuthorizationToken, APIKey, Token. This request to have a validation came from Sara.

   Reason for Change(s):
   - Added above

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes

**Testing Screenshot:**
image

### DIFF
--- a/.github/workflows/validateFieldTypes.yaml
+++ b/.github/workflows/validateFieldTypes.yaml
@@ -1,0 +1,50 @@
+name: Validate Parameter Fields Type in for Security
+
+on:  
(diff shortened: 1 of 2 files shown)


mkchiliveri (approved):
Looks good

v-dvedak (approved):

PR #10147, opened 2024-03-14
Detections test case fix

v-amolpatil:
   
   Change(s):
   - There were 2 files in our repo i.e. one in detections folder and one in solution analytic rules folder and test case use to pass if it identifies 1st file value so we have fixed by checking validation on all files with same name in detection and solution folder so that it fails.

   Reason for Change(s):
   - Specified above

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes


mkchiliveri (approved):
Looks good.

v-atulyadav (approved):

PR #10190, opened 2024-03-21
Add CCP v2 connector to Jira solution

joanabmartins:
   
   Change(s):
   - New CCP v2 connector
   - Updated parser

   Reason for Change(s):
   - New codeless connector option, instead of function, and parser to include new table.

   Testing Completed:
   - Need Help
   Only manual testing. Packaging with v3 gave ID errors, and in the mainTemplate it didn't properly include CCP details (table, DCR, polling). Also, it doesn't seem like the parser is being added properly to the mainTemplate. We know we need to declare the polling parameters manually in the mainTemplate, but we didn't do that because the polling resource doesn't show.

   - Need help
   Reasons described above.


v-prasadboke:
Hello @joanabmartins, Thanks for raising the PR. As discussed over teams chat I'll take a look at CCP connector mapping as expected changes doesnt reflects in the maintemplate

v-prasadboke:
Hello @joanabmartins, 
Sorry for the delay we are still working on CCP connector Mapping

v-prasadboke (approved):

v-atulyadav (approved):

PR #8933, opened 2023-09-03
O365 Data: Fix ZIP and add default values for new variables

doc352:
Added default values for ManagementAPI and LoginEndpoint to match the handling of the Log Analytics URI and better support existing deployments Updated the ZIP with the changes to run.ps1 and removed the parent folder which had caused issues with WEBSITE_RUN_FROM_PACKAGE

   Required items, please complete
   
   Change(s):
   - O365 Data Connector: fixed run.ps1 by adding default values for $ManagementAPI and $LoginEndpoint to better support existing deployments that don't have the new environment variables from the updated AzureDeploy.json in #8756 
   - O365 Data Connector: Updated the function ZIP with the updated run.ps1 and removed the parent folder from the ZIP to fix an issue with WEBSITE_RUN_FROM_PACKAGE

   Reason for Change(s):
   - Bug fixes

   Testing Completed:
   - Deployed from an old version of AzureDeploy.json, updated WEBSITE_RUN_FROM_PACKAGE to point to the updated ZIP, confirmed the function runs correctly with the default values

   Checked that the validations are passing and have addressed any issues that are present:
   - Yes

PR #10350, opened 2024-04-19
Update run.ps1 - fix LoginURL check

pensivepaddle:
Will be necessary to re-create the zip as well, if this fix is accepted.

   Change(s):
   - Fix for LoginURL check

   Reason for Change(s):
   - In azuredeploy.json the loginEndpoint dynamic parameter is set to `[environment().authentication.loginEndpoint]` which will return with a trailing `/` ([docs](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-functions-deployment)) and as such this test will always fail and halt the process.
   - Removing the `/` from the variable broke the auth. process. 
   - Suspect an issue was introduced in the PR from 8 months ago? #8933 

   Testing Completed:
   - See guidance below

   - See guidance below


pensivepaddle:
Created on false pretences, closing as there was no bug. Apologies.

v-prasadboke (approved):

v-atulyadav (approved):

PR #10421, opened 2024-05-01
AliCloud function - run in time-based iteration & add graceful shutdown

hovavza:
   
   Change(s):
   - AliCloud azure function connector will now run in 60-seconds (configurable) chunks. In every iteration, it will retrieve the data for that time range only, send it to LA, and then update the state manager with that chunk end-time. Then, it will move to the next 60-second chunk. 
   - Added graceful cancellation mechanism. Since max time for function is 20 minutes, we will stop it once we get to 18 minutes (configurable), before we get a timeout error. That way, the next function execution can continue from the last successful timestamp, so less chance of duplicate data and data loss.
   - Fixed bug in which the state is updated at the start of the run (in the generate_date method). We should only update the state after we successfully published data to LA (in our case, when we finished a time range chunk).

   Reason for Change(s):
   - Multiple timeouts for customer with high volume. Function tries to retrieve data from last 30 minutes, and as a result, fails on timeout just before finishing to process it.

   Testing Completed:
   - No - this isn't a KQL/table change, so existing testing & validation do not exist.

   - No - this isn't a KQL/table change, so existing testing & validation do not exist.

### DIFF
--- a/Solutions/Alibaba Cloud/DataConnectors/AliCloudSentinelConnector/__init__.py
+++ b/Solutions/Alibaba Cloud/DataConnectors/AliCloudSentinelConnector/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 from aliyun.log import *
 import os
+import time
 import requests
 import datetime
 import hashlib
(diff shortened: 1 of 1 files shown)


v-prasadboke:
Hello @hovavza, Please update the function app zip as well.

v-prasadboke:
Thanks for committing the requested changes, Please accept license agreement too

v-prasadboke:
Hello @hovavza, Please accept license agreement

v-prasadboke:
We wanted to check on the status of PR #10421. PR is pending for more than 30 days in draft state. Please let us know if you need any assistance to review this PR. Per our standard operating procedures if no response is received in the next 5 business days, we will close this PR. Thank you for your cooperation.

v-prasadboke:
Since we have not received a response in the last 7 days, we are closing your PR #10421 per our standard operating procedures. If you still need support for this issue, you can re-open the PR at any time. 

PR #10497, opened 2024-05-20
alicloud data conncetor - re-implement with storage-queue

hovavza:
Change(s):
This is a full re-implementation of this function, since it's not resilient, can't handle high scale, and is very vulnerable to timeouts (which apparently, Ali endpoints have a tendency to cause in high scale).
In short, this function has been split into 2 functions (time triggered sends messages to storage queue, & queue triggered that handles each storage queue message).
   1. TimeTrigger - Very similar to the existing function for GoogleWorkspaceReports. 
- Runs every X minutes (configurable), retrieves list of projects from Ali endpoint (if necessary; if customer included list of needed projects in config, no need) and creates a message in a new storage queue for every project + time window (by default 1 minute, can be changed by customer). Each message indicates a data-polling job that needs to be performed by the 2nd function. So if we run every 4 minutes for 2 projects, each run of the function will create 8 messages in the queue (and allow the 2nd function to run with 8 separate instances, one per queue message). 
- We rely on a bookmark to continue for each project from the previous date-range we scheduled. The difference from before is that now we have a bookmark per project, and not for entire run. 
- If storage queue has too many items (it indicates that the 2nd function isn't running or has a problem), we don't send more into the queue, so not to cause to much load. If queue has 1k items, our timer function will just wait until there's room in the queue (again, similar to the googleWS connector).
- As in the previous version, we only schedule data retrieval for minimum 10-minutes ago from Ali cloud (this remained as before, since we don't want to miss late-arriving data in Ali, so if now is 8:00, we don't schedule message for 7:51, only for up to 7:50.

2. QueueTrigger - runs when a new message is received in the storage queue. Performs the data collection from Ali endpoint according to the data included in the message (project, time range). 
- Resiliency - on failure, this function has the ability to continue from almost exactly where it left off (to reduce duplicate data sent to LA and to reduce latency in case a given task just takes very long to complete). Thus, on every failure, we send a new message to the storage queue with the same details, but with additional fields that allows us not to query Ali again for a list of stores (since we already retrieved them), and also to continue from collecting data from the exact store we left off (assuming we finished getting logs from some of the stores & sending those logs to LA). 
To do that, we now send to LA after finishing querying every store in the project (before, we first got data from all stores in the project, and only then sent to LA).
- LA-post resiliency - previously, if we failed to send a given chunk of events to LA, we would only print an entry to the log and that's it. That's problematic since it's data loss for the customer that will probably get unnoticed. Instead, we now treat LA-send failures just like any other failures. We retry the function until max-retry, etc. Also, in case of transient failures for LA-send operations (http status >= 500 only), we perform up to 3 internal retry attempts (during same function execution, not causing the function to stop and waiting for a new run).
- Failure handing - assuming we experienced a failure or caught an exception, our azure function doesn't throw an exception. Instead, we schedule a similar queue message (with invisibility of 1 min - so it's taken for retry in 1 minute) with same details and increased dequeue_count field and gracefully stop (so function doesn't appear as failed in azure; azure runtime will delete the original message on its own automatically). However, if dequeue_count reached 15, we move the message to a poison queue and do raise an exception so function does appear as failed. This way, there's less noise for the customer. They only see failures when they actually have data-loss due to exceeding their max re-tries of the function.
The customer can monitor the poison queue and also drain it and send to original queue if they wish (manually, not as part of the function).
If we didn't catch an exception (that shouldn't happen usually, unless the queue message is corrupt or missing mandatory fields, misconfiguration, etc.) so function failed unexpectedly or if there was a runtime failure of the entire function/function-app, we will rely on the azure-runtime internal resiliency capabilities to do the same. In other words, azure runtime itself will retry this message up to 15 attempts (configured in the host.json file; it reilies on a different dequeue_count field, the OOTB storage-queue one, not the one included in the message body). Here as well, if there are 15 failures for this message, azure runtime itself will move this message to the same poison-queue.

3. Timeout handling - both functions have an internal timeout mechanism that causes it to gracefully stop before the runtime timeout and kills the function on its own. So if we defined 10 minute runtime timeout, each function has 8-9 internal timeout; so if we reached that internal timeout mark, we gracefully stop. This allows us to continue from where we left off in the next retry, and also reduces the failure noise for the customer (again, if we get 15 straight timeouts on the message, it will not be re-tried and the function will be flagged as failed).

4. Change ARM template to have different env variables - 
- AliCloudWorkers - removed. No need as concurrency is now handled automatically by azure runtime (since one function per queue message).
- MaxWindowSizePerApiCallInSeconds - determines the time range of each queue message. by default, 1 minute. It's helpful when there's high scale and retrieving X-seconds worth of data takes too long. So customer can increase if there aren't scale issues, and reduce it if there are.
- MaxQueueMessageRetries - determines how many times the message is re-tried (due to caught exceptions and failures) before stopping and moving to poison queue. By default it's 15. Note that this does NOT change the number of retries for uncaught exceptions (and runtime failures), this cannot be dynamically configured (as it's defined in host file) and is always 15. However, in most cases, failures will be caught failures.
- TimerCronSchedule - allows the customer to contorl how often the TimeTrigger functions runs. By default it's every 4 minutes.

5. Added mock for Ali SDK - this isn't used (commented out), but it's in the package for easier debugging (has same relevant interface as the Ali SDK).

   Reason for Change(s):
   - In high scale, the function keeps failing on timeout as sometimes takes 10 minutes to retrieve 1 minute worth of data from Ali endpoints. Also, we need ability to run multiple function instances to improve latency (due to the latency mentioned above).

   Version Updated:
   - No

   Testing Completed:
   - Yes - Manual & with customer with high volume (with original timeouts issues).

   - See guidance below


hovavza:
Yes

hovavza:


‫בתאריך יום ב׳, 20 במאי 2024 ב-9:06 מאת
‪microsoft-github-policy-service[bot]‬‏ ***@***.***‬‏>:‬

v-prasadboke:
We wanted to check on the status of PR #10497. PR is pending for more than 30 days in draft state. Please let us know if you need any assistance to review this PR. Per our standard operating procedures if no response is received in the next 5 business days, we will close this PR. Thank you for your cooperation.

v-prasadboke:
Since we have not received a response in the last 7 days, we are closing your PR #10497 per our standard operating procedures. If you still need support for this issue, you can re-open the PR at any time. 

PR #10430, opened 2024-05-03
Fixing bytes ecoding.

NikitaGrunskyHolm:
Fixing `bild_signature` function.

PR #10534, opened 2024-05-27
upgrade version

NikitaGrunskyHolm:
Recently we've fixed bug in function in this pull request: https://github.com/Azure/Azure-Sentinel/pull/10430. But in zip file downloaded from https://aka.ms/sentinel-holmsecurityassets-functionapp still contains old version with bug.

We want to know how to update zip file returned by this link. Do we need always provide new package to trigger files update? I will be very thankful if you could clarify this!


v-prasadboke:
Hello @NikitaGrunskyHolm, Please provide me your mail id so that we can discuss about this. Or you can ping me on teams 
my teams handle is v-prasadboke@microsoft.com

PR #10559, opened 2024-05-31
ASIM-Tester-Workflow-Fix

vakohl:
   
   Change(s):
   - Changing Workflow File Step version for azure/loginV1 to V2

   Reason for Change(s):
   - azure/loginv1 was deprecated and was causing failure for ASIM Workflow execution

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes


manishkumar1991 (approved):
approving this change, as Git hub has deprecated the use of node 16 in v1 version of github  action called azure/login because the node 16 has reached its end of life , so V2 is the new version of same git action , which is running on 20 node version and is recommended as well

v-atulyadav (approved):

PR #10596, opened 2024-06-05
Delete Playbooks/Move-LogAnalytics-to-Storage directory

malowe101:
   
   Change(s):
   - Remove playbook from the repository

   Reason for Change(s):
   - Customers and users continue to use this playbook despite it not being recommended and we offer out of box functionality that should be used instead. I would like for this to be removed to prevent people using content that is not recommended.


v-shukore (approved):

rahul0216:
Hi @malowe101, instead of removing it completely, I would recommend updating the playbook name and add 'Deprecated'. Similarly update readme file with deprecated message and add link to inbuilt features' guidance.  
Upright deleting the playbook may raise unnecessary customer escalations.

v-shukore:
Hi @malowe101 please check the above comment provided by Rahul for deleting the playbooks. Thanks.

malowe101:
I will resubmit the PR with the changes recommended above.

PR #10833, opened 2024-07-17
Updated python version from 3.8 to 3.9 for Security Scorecard Data Connectors

niralishah-crest:
Change(s):

Updated python version fro 3.8 to 3.9 for Security Scorecard Data Connector
Reason for Change(s):

python version 3.8 will be deprecated so need to upgrade python version
Version Updated:

No
Testing Completed:

Yes

Yes


v-prasadboke:
Hello @niralishah-crest, Thanks for raising this PR. This Pr will be investigated and we will update you about the same before 23 July, 2024

v-prasadboke (approved):

PR #10874, opened 2024-07-25
Add T1659 technique for initial access and command and control tactic

v-amolpatil:
   
   Change(s):
   - Added T1659 technique which was missing for tactics "initial access" and "command and control". Have Referred https://attack.mitre.org/techniques/T1659/. Due to this we were getting pipeline error as "Invalid data model. No valid tactic corresponding to the technique T1659 was provided in the tactics field., but found System.Exception with message "Invalid data model. No valid tactic corresponding to the technique T1659 was provided in the tactics field.".
   
   Reason for Change(s):
   - Specified above

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes


rahul0216 (changes requested):
Sentinel supports till **MITRE ATT&CK version 13.1** currently. Technique T1659 was added in later version. I would suggest removing this. Rather update the analytic rule/hunting query which is causing this break.

v-amolpatil:
As per discussion and comment from Rahul closing this PR. We are not supporting this technique in current version 13.1 of Mitre Att&ck.

PR #11045, opened 2024-08-29
Update ScanSecrets.yaml to output only verified results

manishkumar1991:
   
   Change(s):
   - added a new flag to output only verified results

   Reason for Change(s):
   - earlier flag is flagging the plain text as unverified secret key 
   - 
image

image

   Version Updated:
   - N/A

   Testing Completed:
   - Yes

   - Yes


rahul0216 (approved):
Changes look good.

v-atulyadav (approved):

PR #11257, opened 2024-10-10
Update imDns_DomainEntity_DnsEvents.yaml

Bievens:
Change(s):
   - Added column Domain to project operation to allow for alert results to include the field and for entity mapping
   - Added entityMapping for type DNS on column Domain
   - Incremented version number

   Reason for Change(s):
   - DNS Entity typically aligns better with logs associated with domain TI indicators and DNS logs may contain the Domain but not a URL.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


v-prasadboke:
Please accept license agreement here

Bievens:

v-prasadboke (approved):

PR #11314, opened 2024-10-22
Enforcing ASim validations when failure comes

vakohl:
   
   Change(s):
   - We want to fail ASim PR when any ASIM parser review test case fails

   Reason for Change(s):
   - We want to fail PR when any ASIM parser review test case fails

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes


manishkumar1991 (approved):
Approved

v-atulyadav (approved):

PR #10701, opened 2024-06-25
Update BloodHound Enterprise Solution function app

gwhite-so:
Required items, please complete
   
   Change(s):
   - Update bhe_client.py function app data connector with new findings API endpoint

   Reason for Change(s):
   - The old findings API will be deprecated

   Testing Completed:
   - Yes

   Checked that the validations are passing and have addressed any issues that are present:
   - Yes

PR #11380, opened 2024-11-06
SpecterOps Bloodhound Solution

daviditkin:
   
   Change(s):
   - Replaced python function with golang connector that uses bloodhound-go-sdk
   - Updated all workbooks, analytic rules, function_app, connector to use new schema
   - We use blob storage to keep last ingestion state

   Reason for Change(s):
   - The python function was replaced with golang and updates to custom table to reflect new fields

   Version Updated:
   - No
   - The version remains the same since this solution has never been approved
   
   Testing Completed:
   - Tested a deployment of the masterTemplate in a clean azure environment. 

   - Does not validate cleanly, but seems to be the same issues that all other masterTemplate.json have.


StephenHinck:
This PR replaces https://github.com/Azure/Azure-Sentinel/pull/10701. We are refactoring the application entirely.

daviditkin:
@microsoft-github-policy-service agree [company="Specter Ops, Inc."]

daviditkin:
@microsoft-github-policy-service agree company="your company"

PR #11482, opened 2024-11-25
Update OCI_logs_API_FunctionApp.json

garis:
Disclaimer for the 500 columns limit causing ingestion issues for the OCI table.


v-prasadboke:
https://github.com/Azure/Azure-Sentinel/blob/master/Tools/Create-Azure-Sentinel-Solution/V3/README.md

Please repackage the solution using V3 tool

PR #11505, opened 2024-11-28
Fix alert importer managed identity

niklas-recordedfuture:
   
   Change(s):
   - Fixed `RecordedFuture-AlertImporter` to default its `azuremonitorlogs` API connection to system-assigned managed identity

   Reason for Change(s):
   - Bugfix

   Testing Completed:
   - Yes

   - Yes


niklas-recordedfuture:
@v-prasadboke Hi, thanks for looking into this.

Was it intentional to change our solution version by bumping the minor version? In my mind this is just a small bugfix release, which is why I bumped the patch version to 3.2.12. If you think this should be a minor bump then that's fine, but I think changelog should also be updated in that case.

Thanks!

v-prasadboke (approved):

PR #11520, opened 2024-12-04
Use variables instead of parameters

Cv-securityIQ:
   
   Change(s):
   - Use variables('FunctionName') instead of parameters('FunctionName')

   Reason for Change(s):
   - Use variables('FunctionName') instead of parameters('FunctionName'), because the function app name in the variable is function name and resource group id., which is the correct id.

   Version Updated:
   - Not needed

   Testing Completed:
   - Yes, I have tested it, if this works


Cv-securityIQ:
Hi. Just fixed a small mistake in the last PR where I accidentally used the wrong variable for the function app name. Could you please review the fix?

v-prasadboke (approved):

PR #11538, opened 2024-12-09
False Positives from "Correlate Unfamiliar sign-in properties & atypical travel alerts" When Changing User Risk

v-visodadasi:
   
   Change(s):
   - Modified query to exclude events where the Comments field starts with "Risk detail: Admin"

   Reason for Change(s):
   - To exclude alerts that are the result of admin actions, helping to reduce false positives.

   Version Updated:
   - Yes
   - 
   Testing Completed:
   - Yes


quantumburnz:
@v-prasadboke , Hi, I see you closed this issue without merging the commits. Is this change still going to be implemented or will a fix be made?

quantumburnz:
@v-visodadasi / @v-prasadboke - following up on my last question. Is this change still going to be implemented or will a fix be made? I don't understand why this was closed with unmerged commits.

ramonion:
this change is an improvement, but i would suggest to use "!contains" instead of "!startswith" as there may be comments like "Entra ID updated this alert due to the following risk detail: Admin dismissed all risk for user" which will not be matched with "startswith"

PR #11654, opened 2025-01-10
Use arrivalTime to avoid missing events

nitsan-tzur:
Required items, please complete
   
   Change(s):
   - __init__.py

   Reason for Change(s):
   - Use arrivalTime instead of eventDate to avoid missing events

PR #11739, opened 2025-01-29
Cyberark EPM - Fix zipped package structure

SpeedyFireCyclone:
   
   Change(s):
   - Repackage zip file to conform to Function App expectations.

   Reason for Change(s):
   - Correct issue introduced in #11654
   - Nested directory is not being picked up by the Function App runtime and has caused log ingestion to stop.


rahul0216 (approved):
Looks good.

PR #11777, opened 2025-02-07
Netskope CCP Integration

mitchellgulledge2:
   
   Change(s):
   - Added Netskope CCP data connector to existing solution

   Reason for Change(s):
   - GA for Netskope CCP Integration

   Testing Completed:
   - Netskope Validated in Internal MSFT environment


mitchellgulledge2:
@v-prasadboke let me know if any other changes need to be made and thanks in advance

v-prasadboke:
Hello @mitchellgulledge2, i have made some required changes. 
Can you share demo credentials to test connector. 

you can share it over mail or teams. 
my handle : v-prasadboke@microsoft.com

v-prasadboke:
Hello @mitchellgulledge2, Please try to deploy and test the connector using the maintemplate i commited. 

it is throwing me 409 error

mitchellgulledge2:
The 409 is because we were running too many deployments with the same apikey on the same tenant for the same endpoint. Data should be flowing correctly but we discovered a few issues.
1) With some of the sample data sets Sentinel converts integer data types to type 32. Some of our data sets for like incident ID are 64 bit.
2) When we converted data type from int to long we noticed that after the 16th digit in the integer that Sentinel converts anything after to 0. So customers reported incorrected Incident IDs since it appears sentinel does a javascript evaluation (didnt see this documented anywhere) 
So now we are having to test to try and see if converting to string resolves issue but we should likely document that sentinel cannot support anything past a long value of 9007199254740991.

v-prasadboke (approved):
looks good

v-atulyadav (approved):

PR #11781, opened 2025-02-09
Suspicious registry additions

JouniMi:
Added queries to find for suspicious registry additions.

Change(s):

- Adding Hunting Queries/Microsoft 365 Defender/Defense evasion/suspicious-base64-encoded-registry-keys.yaml 
- Adding Hunting Queries/Microsoft 365 Defender/Defense evasion/suspicious-command-interpreters-added-to-registry.yaml
- Adding Hunting Queries/Microsoft 365 Defender/Defense evasion/suspicious-keywords-in-registry.yaml Reason for 

Pics of queries running:
Screenshot 2025-02-09 at 14 33 30
Screenshot 2025-02-09 at 14 34 36
Screenshot 2025-02-09 at 14 34 00

### DIFF
--- a/Hunting Queries/Microsoft 365 Defender/Defense evasion/suspicious-base64-encoded-registry-keys.yaml
+++ b/Hunting Queries/Microsoft 365 Defender/Defense evasion/suspicious-base64-encoded-registry-keys.yaml
@@ -23,4 +23,4 @@ query: |
   | extend DecodedCommand = replace(@'\x00','', base64_decode_tostring(RegistryValueData))
   | extend ExtractedDecodedCommand = base64_decode_tostring(ExtractedB64)
   | where isnotempty(DecodedCommand) or isnotempty(ExtractedDecodedCommand)
-  | project Timestamp, DeviceName, DecodedCommand, ExtractedDecodedCommand, RegistryValueData, RegistryKey, RegistryValueName, RegistryValueType, PreviousRegistryValueData, InitiatingProcessFileName, InitiatingProcessCommandLine, InitiatingProcessParentFileName, InitiatingProcessRemoteSessionDeviceName, InitiatingProcessRemoteSessionIP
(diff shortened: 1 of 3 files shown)


JouniMi:
The KQL checks fails because they they do not recognize "InitiatingProcessRemoteSessionDeviceName". This is a new field to Defender for Endpoint. I think that test needs to be updated, as this is standard field within the tool. I am happy to remove those fields from the queries to make this test pass - the actual fix should be to fix the test though:
image

JouniMi:
> The KQL checks fails because they they do not recognize "InitiatingProcessRemoteSessionDeviceName". This is a new field to Defender for Endpoint. I think that test needs to be updated, as this is standard field within the tool. I am happy to remove those fields from the queries to make this test pass - the actual fix should be to fix the test though: image

Same for the field InitiatingProcessRemoteSessionIP.

v-shukore:
Hi @JouniMi, if the `InitiatingProcessRemoteSessionDeviceName` field is not required and does not impact the query result, could you please remove this field to pass the validation check?
Thanks!!

v-shukore (approved):

v-prasadboke:
Hello @JouniMi, I see the column was removed from query due to validation failure. 
But what if there are users who will be needing this column when they use hunting query.

JouniMi:
> Hello @JouniMi, I see the column was removed from query due to validation failure. But what if there are users who will be needing this column when they use hunting query.

Users may add more columns themselves. The KQL validation does not work with these columns **which are by default included in Defender for Endpoint.** I think that the validation should be fixed to which I opened an issue which was closed as I removed these columns from these hunting queries - which also I don't understand https://github.com/Azure/Azure-Sentinel/issues/11787#issuecomment-2650176515. I don't know how to fix that KQL check.

As for these queries, those columns are not mandatory to analyze the results, just may bring additional value. As long as the KQL validation do not account for these default columns in Defender for Endpoint I don't think it is feasible to require to have them.

This fix was also proposed in this comment: https://github.com/Azure/Azure-Sentinel/pull/11781#issuecomment-2648109997

v-prasadboke (approved):

v-dvedak (approved):

PR #11826, opened 2025-02-18
Fix Azure CLI version comparison bug

0x-art:
Change(s):
   - This commit swaps the arguments to `verlte "$azver" "2.33.1"` so that the function correctly returns true only when the installed version is less than or equal to 2.33.1

   Reason for Change(s):
   - The scripts currently check for an outdated Azure CLI version by calling `verlte "2.33.1" "$azver"`, which incorrectly evaluates newer versions (e.g., 2.38.2) as outdated due to reversed argument order.
   - This fix prevents unnecessary removal and reinstallation of the Azure CLI.


0x-art:

v-prasadboke (approved):

PR #11859, opened 2025-02-26
Query Period changed to Standard Format

### LINKED ISSUE #11759: Query Period Value Incosistency
**Describe the bug**
The query period set in the template link below is 70m. When this is deployed and then referenced the return value is 1h10m. Could the template be updated to match the value once deployed so that any automated deployment of these templates does not see this as a changing value.

https://github.com/Azure/Azure-Sentinel/blob/b228a2f19663559ebd06145a80cfe8acd7549b56/Solutions/Microsoft%20Entra%20ID/Analytic%20Rules/SuspiciousServicePrincipalcreationactivity.yaml#L12 

**To Reproduce**
We currently utilise an autoamted solution using terraform to install the Content Hub solution and then deploy the rule templates using the data contained within the template. After an initial deployment an additional re-run will return the value change due to the template containing different settings to the deployed value.

**Expected behavior**
The template query value should match the value of the setting once deployed.

**Screenshots**

Image

**Additional context**
Updating the template value to 1h10m rather than 70m will then allow the values to match.

v-visodadasi:
   
   Change(s):
   - query period from a non-standard format (70m) to a standardized format (1h10m)

   Reason for Change(s):
   - To ensure consistency between the template value and the deployed value, preventing discrepancies during automated deployments.

   Version Updated:
   - Yes

   Testing Completed:
   - Yes


rahul0216 (approved):
Change look good.

f-bader:
@v-visodadasi The change of the queryPeriod value breaks this YAML as ARM files do not support this format

image

PR #12021, opened 2025-04-08
Updated Entity Mappings of Analytic Rules of Cisco ISE solution

v-shukore:
   
   Change(s):
   - Updated Entity Mappings of Analytic Rules

   Reason for Change(s):
   - Updated Entity Mappings of Analytic Rules

   Version Updated:
   - Yes

   Testing Completed:
   - Yes


ashwin-patil (changes requested):
These are applicable to all files within the PR. I havent added individual comment per file so adding summary here. 
Please add additional fields to the output and add aggregations by entities and event name in all the fields.
- For account names , add domain names as strong identifiers. 
- Add FileName Entity where it is aggregated and present
- Add IP address entity where it is present but missing in entity mapping

PR #12035, opened 2025-04-10
Fix SAP LogServ retention, SAP ETD OData entity change

MartinPankraz:
Change(s):
   - Drop retention setting from LogServ table
   - Replace OData entity on SAP ETD solution

   Reason for Change(s):
   - Allow LogServ customers to use their LogAnalytics workspace defaults rather than be forced to accept 30d setting.
   - Enable SAP ETD customers to use real time events (TriggeringEvents vs. NormalizedTriggeringEvents entity)

   Version Updated:
   - 3.0.1

   Testing Completed:
   -


MartinPankraz:
Fixing branch validation with https://github.com/Azure/Azure-Sentinel/pull/12037

PR #12056, opened 2025-04-15
Issue Resolved for vimNetworkSessionCheckPointFirewall 

### LINKED ISSUE #11940: Check Point ASIM Firewall parser incorrect assumption that `CommunicationDirection` maps to `NetworkDirection`
Cannot correctly use this to set network direction given it there is seldom any context if the interface (ifname) is an external vs internal interface.

- `CommunicationDirection` seems to be Micrsoft's obscured remapping from `deviceDirection` in the CEF extensions (guessing what Microsoft maps where from extensions in the CommonSecurityLog table is not fun / black-box?).
- `deviceDirection` is in fact the `ifdir` (interface direction), so this is a direction from the perspective of an interface?
- But there is often not enough context to know if a given interface is on an internal or external zone?
- Therefore misleading NetworkDirection mappings can result.
  - E.g. "inbound" from the interface perspective where the interface is on an internal zone could actually be "outbound" traffic from a network perspective?

As per

https://github.com/Azure/Azure-Sentinel/blob/0776c20d26cc5112ee2bbf6dcff2e06c896118ed/Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCheckPointFirewall.yaml#L310

In the case block:

```kql
                NetworkDirection = case(
                                    isnotempty(NetworkDirection), NetworkDirection,
                                    inzone == "Internal" and (outzone == "Internal" or outzone == "Local"), "Local",
                                    (inzone == "Internal" or inzone == "Local") and outzone == "External", "Outbound",
                                    inzone == "External" and (outzone == "Internal" or outzone == "Local"), "Inbound",
                                    CommunicationDirection == "0", "Inbound",
                                    CommunicationDirection == "1", "Outbound",
                                    ""
                                    ),
```

`CommunicationDirection` which Microsoft presumably mapped from the CEF Extension field `deviceDirection` assumes this the "Network" direction and maps it directly. At least it only does that if better choices couldn't be matched, but it could end up incorrect regardless.

As per other references:

[Log Exporter CEF Field Mappings](https://community.checkpoint.com/t5/Management/Log-Exporter-CEF-Field-Mappings/td-p/41060) and [Description of Fields in Check Point Logs](https://support.checkpoint.com/results/sk/sk144192)

- `deviceDirection` is the exported name from an original field `ifdir`.
- `ifdir` was in the list described as **"Disclaimer - These fields are only used for Check Point internal purposes"** implying it shouldn't be relied on.

Given checkpoint does log the `ifname`, it seems `ifdir` means which direction is the traffic arriving through the interface. However, the problem is a firewall can have multiple interfaces and the Extensions in CEF don't seem to tag/label if the interface is external or not.

To double check the issue, I was able to find CheckPoint events where the `ifdir` / `deviceDirection` =  0 (inbound) meanwhile `conn_direction` = Outgoing was set so it's very possible these fields can contradict given they are for different perspectives (interface vs network zones).

v-tsawant:
   
   Change(s):
   - Removed CommunicationDirection from NetworkDirection
   - Added Additionfields to store communication direction

   Reason for Change(s):
   - GitHub issue: 11940 This parser is zone specific and not interface specific there for remove the logic 

   Version Updated:
   - 1.2


v-tsawant:

v-utpalkumar:
Investigating the causes of the failed checks.

v-utpalkumar (approved):

v-atulyadav (approved):

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCheckPointFirewall.yaml
```diff
@@ -194,7 +194,7 @@ ParserQuery: |
         {
         CommonSecurityLog
         | where not(disabled)
-        | where DeviceVendor=="Check Point" and DeviceProduct=="VPN-1 & FireWall-1"
+        | where DeviceVendor=="CheckPoint" and DeviceProduct=="VPN-1 & FireWall-1"
```
BWC-TomW, line 197: @v-sabiraj - Can see your commit #01e2fca92c9a9c3341725b149a02839ca0036086 re-introduces the space here but seems this was overwritten again in this PR which has broken the parsing. 

The logs we see have the space but perhaps something like 

| where replace_string(DeviceVendor, " ","") =~ "CheckPoint" and DeviceProduct=~"VPN-1 & FireWall-1"

would keep everyone happy

PR #12111, opened 2025-04-28
Update package documentation and listing

nitsan-tzur:
   
   Change(s):
   - mainTemplate.json -> Update instructions to use Python 3.10 version
   - 3.0.1.zip
   - ReleaseNotes.md

   Reason for Change(s):
   - Python 3.8 EOL


nitsan-tzur:
> Hi @nitsan-tzur, Ensure that the Python version in the data connector instructions is updated, and then repackage the solution.
> 
> image

Hi @v-atulyadav, I've updated the quoted file doesn't seem to be part of the current zip package.
image

v-atulyadav (approved):

v-atulyadav:
Hi @nitsan-tzur,
It appears that the main template has been manually updated without repackaging, which has resulted in the solution version not being reflected in the main template. To ensure that the correct changes are applied, you will need to repackage this solution.

image

PR #12203, opened 2025-05-22
Patchon aws detection rules

Elleinshar:
Add PolicyCreationDate column for better Output
Add Correction on join flavor to avoid triggering endless incdent for 14 days straight 
Add some fixes after Error Review

Change(s):

    Change queryPeriod
    Add new Column PolicyCreationDate
    Change join flavor
    add CustomDetail PolicyName

Reason for Change(s):

    Change of query Period : as this query is designed to lookback on 14Days via the let lookback =14d command, with a set up at 1 day Query Period, this will not have the intended purpose and only for data up to 1 day
    New column for better clarity
    Change of the join flavor : as we change the query period, we will generate dupplicated incident at every run of the rule if we keep the flavor as leftouter. With a inner flavor, we keep the designated purpose of the rule : detecting a Policy creation in the last 14 days and alert if it have been attach in the last day. And thus incident will be generated only if the attach envent is detected throught the join

Version Updated:

    Yes

Testing Completed:

    Yes

    Need Help


Elleinshar:
microsoft-github-policy-service agree

oliviahuegel:
@ashwin-patil tagging you on this PR

v-maheshbh:
@Elleinshar   All the files in detections and solution (Analytics rules) folder are supposed to end with .yaml  not .yml

Elleinshar:
Oh okay I didn't see the second file being push 
I will update that
Thank you for the information

v-maheshbh:
@Elleinshar any update please?

Elleinshar:
Yes just updated the File 
Sorry for the delay

Elleinshar:
Patched Dupplciate GUID with : bb3e5d55-3f87-4dcb-860b-a7decabf3304

PR #12296, opened 2025-06-10
Update for support of SecureStrings type 

NobufumiMurata:
Response to the change in the return value of Get-AzAccessToken to SecureStrings in the following versions: 
Az.Accounts version 5.0.0
Az version 14.0.0

https://learn.microsoft.com/en-us/powershell/azure/protect-secrets?view=azps-14.1.0#transition-from-strings-to-securestrings

   Change(s):
   - Update Rest API Bearer Header logic for [Playbook_ARM_Template_Generator.ps1](https://github.com/Azure/Azure-Sentinel/blob/master/Tools/Playbook-ARM-Template-Generator/src/Playbook_ARM_Template_Generator.ps1)

   Reason for Change(s):
   - The default output type of the Get-AzAccessToken cmdlet changed from a plain text String to a SecureString, starting with Az.Accounts version 5.0.0 and Az version 14.0.0.
   - [Learn Page](https://learn.microsoft.com/en-us/powershell/azure/protect-secrets?view=azps-14.1.0#transition-from-strings-to-securestrings)

   Version Updated:
   - Not Detections/Analytic Rule templates
   
   Testing Completed:
   - Need Help
   - It was processed correctly in my environment.

   - Need Help
   - It was processed correctly in my environment.

### DIFF
--- a/Tools/Playbook-ARM-Template-Generator/src/Playbook_ARM_Template_Generator.ps1
+++ b/Tools/Playbook-ARM-Template-Generator/src/Playbook_ARM_Template_Generator.ps1
@@ -259,20 +259,32 @@ Function BuildPlaybookArmId() {
 }
 
 Function SendArmGetCall($relativeUrl) {
-    $authHeader = @{
-        'Authorization'='Bearer ' + $tokenToUse
+    if ($tokenToUse -is [System.Security.SecureString]) {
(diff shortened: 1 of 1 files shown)


NobufumiMurata:

v-atulyadav (approved):

PR #12442, opened 2025-06-28
Entra solution analytics update

CyberNinjaCat:
Change(s):
   - Updated PrivilegedGroups to include the "Privileged Authentication Administrator" role
   - Updated OperationsList to remove "Add member to role in PIM", this would have never been included in the results as the query is looking for Role.WellKnownObjectName but this operation doesn't have that field.
   - Updated OperationsList to include "Add eligible member to role" to ensure eligible assignments as surfaced so the analytic provides full coverage.

   Reason for Change(s):
   - Detection logic and accuracy improvements

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   - Yes


CyberNinjaCat:
@v-prasadboke  Hi, any updates on this PR please?

PR #250, opened 2019-08-22
Detection Multidata updates - Tactics, Techniques and formatting updates

shainw:
Fixes #

## Proposed Changes

  -
  -
  -

PR #12558, opened 2025-07-20
Add M365 Defender hunting query for Shimcache flush detection

VlogicK:
   
   Change(s):
   - Added M365 Defender hunting query for Shimcache flush detection.

   Reason for Change(s):
   - To improve visibility into potential anti-forensic techniques, this hunting query detects attempts to flush Shimcache on Windows endpoints. Such behavior is often associated with attackers trying to erase execution traces and evade forensic analysis.

   Version Updated:
   - N/A (Hunting query)

   Testing Completed:
   - Yes

   - Yes

### DIFF
--- a/Hunting Queries/Microsoft 365 Defender/Defense evasion/shimcache-flushed.yaml
+++ b/Hunting Queries/Microsoft 365 Defender/Defense evasion/shimcache-flushed.yaml
@@ -1,0 +1,19 @@
+id: cd90d859-0d8f-458e-9d96-7f2945fe87a6
+name: shimcache-flushed
+description: |
+  This query searches for attempts to flush Shimcache, which may indicate anti-forensic or defense evasion activity by an attacker.
+  Author: Vaasudev_Kala
+  Ref: https://blueteamops.medium.com/shimcache-flush-89daff28d15e
+requiredDataConnectors:
+- connectorId: MicrosoftThreatProtection
+  dataTypes:
+  - DeviceProcessEvents
+tactics:
+- Defense evasion
+relevantTechniques:
+  - T1112
+query: |
+  DeviceProcessEvents
+  | where FileName has "Rundll32.exe"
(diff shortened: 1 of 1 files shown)


VlogicK:

VlogicK:

Hunting Queries/Microsoft 365 Defender/Defense evasion/shimcache-flushed.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - DeviceProcessEvents
+tactics:
+- Defense evasion
+relevantTechniques:
+  - T1112
+query: |
+  DeviceProcessEvents
+  | where FileName has "Rundll32.exe"
+  | where ((ProcessCommandLine has_any("ShimFlushCache","BaseFlushAppcompatCache"))
+  or (ProcessCommandLine has_any (@"#250",@"#46") and ProcessCommandLine has_any ("apphelp.dll","kernel32.dll"))) // checking for ordinals as well
```
4R9UN, line 19: Better to add DLL names, you could use as below , ProcessCommandLine has_any (@"apphelp.dll,#250",@"kernel32.dll,#46")

VlogicK: I want to check the dlls only if the command line has ordinals ( #250 and #46) that's why I added it as an AND clause inside the OR. Because other dlls could also have some other function corresponding to those ordinal numbers and cause false positives.

Does that make sense?

In case the function is being called directly without ordinal numbers, we don't need to have Dlls checked as the function name is self explanatory in that case I feel.

v-atulyadav: Hi @4R9UN, 
Any views or feedback on Vlogic's explanation shared above? Thanks

PR #12638, opened 2025-08-11
Added a new workbook for Alibaba Cloud

samikroy:
Added a new workbook for Alibaba Cloud to visualize the below

- High-Risk API Calls (Security Sensitive)
- Events by AliCloud Service Name
- Accounts Without MFA
- Root vs Assumed Role Usage
- Unusual Source IPs Accessing AliCloud
- Top Event Sources in AliCloud
- Root Account Actions Over Time
- API Call Trends (Potential Attack Timeline)
- ListAccessKeys Event Spike Detection


samikroy:
@v-atulyadav - Could you please share the metedata property which is failaing as have added the similar ones as previous.
Thnak you.

v-atulyadav:
Hi @samikroy,
We're currently investigating this failure and will keep you updated. Thanks

PR #12757, opened 2025-09-08
Suppress CodeQL Alert

rahul0216:
Change(s):
   - Suppress code QL alerts. These alerts are for cryptography module code. Alerted module is not being used in Functions app.

   Reason for Change(s):
   - To suppress code QL alerts.

   Version Updated:
   - NA

   Testing Completed:
   - NA

   - Yes


rahul0216:
**Doesn't require packaging. Only Merge the PR.**

manishkumar1991 (approved):
approved

PR #12772, opened 2025-09-10
Sentinel Transition To Defender Helper Script

mariocuomo:
The tool is a .ps1 script to highlight some aspects to consider before fully adopting Sentinel in Defender—such as Defender XDR Retention data, Analytics Rules, and Automation Rules. The tool allows you to print findings in the shell or generate a PDF report.


mariocuomo:
this is the output in the Shell
output

This is an example of report that is generated from the script
[report.pdf](https://github.com/user-attachments/files/22252938/report.pdf)

rahul0216:
@mariocuomo Copilot has provided some comments. Please review once. Make changes wherever you find correct, rest resolve conversation.

mariocuomo:
@rahul0216 resolved. Copilot provided some suggestion on the comments side. Nothing in terms of functionality.
Just a few words on a security check provided by Copilot to not store the app registration secret directly in the code - suggesting to use an Azure Key Vault. This implementation would make the script less user-friendly adding complexity.

PR #12806, opened 2025-09-15
Tanium Version 3.2.0

Tanium-Nicole:
Change(s):
   - Enhancement(s)
     - All playbook templates now connect to a Key Vault to obtain the API Token for the Tanium API
     - All playbook templates treat the Tanium API URL as a SecureString
     - The analytics rule template has been updated to generate the Azure Sentinel Alert name based on the alert itself, rather than all using the same name
   - Resolved Issue(s)
     - The analytics rule template has been corrected to no longer group alerts

   Reason for Change(s):
   - Playbooks templates should use best security practices to limit bad actors obtaining access to secrets
   - When all Azure Sentinel Alerts use the same name it's difficult to differentiate them in a list view
   - Tanium Threat Response Alerts and Azure Sentinel Alerts should have a 1-to-1 relationship. Also this ensures that the Tanium Threat Response playbook can be used to resolve any alert, not just the first alert on the Azure Sentinel Incident

   Version Updated:
   - Yes, solution version increased from 3.1.0 to 3.2.0

   Testing Completed:
   - Yes, this solution has been installed and tested in an Azure resource group with Log Analytics, Sentinel, an integration account, and access to a Tanium environment

   - Yes


v-shukore:
Hi @Tanium-Nicole, the JSON file validation is failing due to an issue in the cspell.config.json.JSON file. It appears that there is an extra comma causing the error. Could you please remove the extra comma from the file? Thanks.

caleb-clausen-t:
@v-shukore, I removed the extra comma! It should be ready to go.

v-shukore (approved):

PR #12813, opened 2025-09-16
Fix Rule logic for abnoraml protocol

miritsadon:
   
   Change(s):
   - Fix an abnormal protocol rule in GSA solution to ignore empty values. 

   Reason for Change(s):
   - noise.

   Version Updated:

   Testing Completed:
   - See guidance below

   - See guidance below


v-atulyadav:
Raised a new PR https://github.com/Azure/Azure-Sentinel/pull/12816

PR #12874, opened 2025-09-25
Clarify PowerShell version requirements and AWS CLI

eevanwong:
Change(s):
   - Updated installation instructions for PowerShell and AWS CLI

Reason for Change(s):
   -  A CRI about this which resulted in invalid json being passed to `--tags` flag when creating sqs queue. This was investigated and was found because they were using an old version of aws cli.

CRI is pending completion. Will finalize this PR once its resolved.


eevanwong:
Made additional PR here with version checker:
#12924

PR #12924, opened 2025-10-06
Version enforcement for powershell v7

eevanwong:
Change(s):
  - Added powershell version check to ensure user is not using a powershell version < 7
  - Updated README.md to be explicit in using the up to date versions of aws cli and powershell.

Reason for Change(s):
   - There was a CRI in which a consumer ran into this issue. However, while they did install the correct powershell, they had used a prior command of `powershell.exe -ExecutionPolicy Bypass ...` which references the windows powershell executable (pwsh.exe is v7+). They effectively ran powershell v5 on the powershell v7 cli which caused a lot of confusion. This check helps prevent this.

Check was tested and blocks if using windows powershell.

### DIFF
--- a/DataConnectors/AWS-S3/ConfigAwsConnector.ps1
+++ b/DataConnectors/AWS-S3/ConfigAwsConnector.ps1
@@ -36,6 +36,13 @@ param (
 . ".\Utils\AwsSentinelTag.ps1"
 . ".\Enviornment\EnviornmentConstants.ps1"
 
+# Verify that powershell version 7 is being used
+if ($PSVersionTable.PSVersion.Major -lt 7) 
+{ 
+    Write-Error "This script requires PowerShell 7 or higher. You are running version $($PSVersionTable.PSVersion). Please install the latest version of PowerShell from https://aka.ms/powershell and try again."
+    exit
+}
+
 # Verify that the AWS CLI is available
 if ($null -eq (Get-Command "aws" -ErrorAction SilentlyContinue)) 
 { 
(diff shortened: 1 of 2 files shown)


eevanwong:

eevanwong:
Bump

v-atulyadav:
Hi @eevanwong,
The script needs to be compatible with PowerShell 5 also, so proceeding with the merge isn't feasible at this time. We’ve already updated the script to work with both PowerShell versions 5 and 7. Can we proceed to close this request as we don't need this instruction?

eevanwong:
> Hi @eevanwong, The script needs to be compatible with PowerShell 5 also, so proceeding with the merge isn't feasible at this time. We’ve already updated the script to work with both PowerShell versions 5 and 7. Can we proceed to close this request as we don't need this instruction?

Hi, @v-atulyadav , A customer who, both installed the powershell scripts on this repo and from the web page: https://learn.microsoft.com/en-us/azure/sentinel/connect-aws?tabs=s3#instructions failed to go through this script because of faulty json parsing. They were following the instructions and the script itself was failing to pass in the tags for the sqs queue (no user input, the script itself failed). I was able to replicate this error when running this on windows powershell (v5), and the definitive fix was to update the powershell version.

I would like to understand why this needs to be compatible with powershell 5? In the original documentation, it asks the user to install powershell (which references to version 7).

eevanwong:
@v-atulyadav, any update on this? It would be nice to resolve this to prevent customers from running into this issue again.

PR #12939, opened 2025-10-08
v2.5 update

markolauren:
introduces "byos" (bring-your-own-schema) instead of always copying the schema from existing table.

   Change(s):
   - See guidance below

   Reason for Change(s):
   - new version with new feature

   Version Updated:
   - See guidance below

   Testing Completed:
   - See guidance below

   - See guidance below


markolauren:
What's pending the merge?

PR #12972, opened 2025-10-16
Removing cffi module files

rahul0216:
   
   Change(s):
   - Deleted cffi module files as it is raising codeql alerts

   Reason for Change(s):
   - To mitigate CodeQL alerts

   Version Updated:
   - NA

   Testing Completed:
   - NA

   - Yes


rahul0216:
**No need to repackage the solution.**

PR #12990, opened 2025-10-22
Correct the inconsistency between the package name and the offer name

nikita-konoplin:
Change(s):
   - OfferId changed

   Reason for Change(s):
   - Inconsistency between the package name and the offer name

   Version Updated:
   - 3.0.2
   Testing Completed:
   - Yes

   - Yes

### DIFF
--- a/Solutions/Veeam/Package/mainTemplate.json
+++ b/Solutions/Veeam/Package/mainTemplate.json
@@ -138,7 +138,7 @@
     "_email": "[variables('email')]",
     "_solutionName": "Veeam",
     "_solutionVersion": "3.0.2",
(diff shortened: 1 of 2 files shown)


nikita-konoplin:
Hi @v-shukore, @v-atulyadav,
During the certification publishing step, we found an issue where the package name and the offer name do not match. We need to fix this as soon as possible. Please let me know if anything is needed from our side.

Thank you!
Nikita

v-atulyadav:
Hi @nikita-konoplin,
Please repackage the solution to ensure this change is included in the ZIP file as well. You can refer to the link below for the packaging steps. Also after packaging accept license/cla.

https://github.com/Azure/Azure-Sentinel/blob/master/Tools/Create-Azure-Sentinel-Solution/V3/README.md

PR #13159, opened 2025-11-18
Update IPEntity_AppServiceHTTPLogs.yaml

TommyJohansson:
Missing column added, alertSeverityColumnName rely on that field

   Change(s):
   - Added missing column in IPEntity_AppServiceHTTPLogs.yaml

   Reason for Change(s):
   - Rule was broken, not able to install without the fix

   Version Updated:
   - Yes, updated

   Testing Completed:
   - Added column during installation from template and it worked.

   - See guidance below


TommyJohansson:
@microsoft-github-policy-service agree company="SentorSecurity"

Issue #13169, opened 2025-11-20
ASimAuditEvent - Improperly Formatted Value for Scheduled Tasks


andrewj-t:
**Describe the bug**
When the `vimAuditEventMicrosoftSecurityEvents` or `ASimAuditEventMicrosoftSecurityEvents` return parsers results for Scheduled Task Events. The Value/NewValue field contains improperly unescaped XML which can not be parsed with parse_xml. The versions of the ASIM AuditEvent parsers which operate on the `WindowsEvent` table, do not have this issue and will return properly formatted XML

**To Reproduce**
1. Run this Kql
```kql
ASimAuditEventMicrosoftSecurityEvents
| where EventOriginalType == 4698
```
2. Observe that partially escaped XML is returned for the Value field
```text
<?xml version="1.0" encoding="UTF-16"?>
<Task version="1.4" xmlns="http://schemas.microsoft.com/windows/2004/02/mit/task">
	<RegistrationInfo>
```
4. Run this Kql to compare results to WindowsEvent based logs
```kql
ASimAuditEventMicrosoftWindowsEvents
| where EventOriginalType == 4698
```
5. Observe that correctly formatedXML is returned for the Value field
```text
&lt;?xml version="1.0" encoding="UTF-16"?&gt;
&lt;Task version="1.4" xmlns="http://schemas.microsoft.com/windows/2004/02/mit/task"&gt;
	&lt;RegistrationInfo&gt;
```

**Expected behavior**
As these Parsers are both operating on Windows Event Logs, they should return the information consistently, and correctly formatted XML, so it can be further parsed by parse_xml or similiar by the user if required

**Additional context**
I have created and fix and will submit a PR shortly

v-utpalkumar:
Closing this issue since the associated PR https://github.com/Azure/Azure-Sentinel/pull/13170 has been closed by the author.

Issue #13168, opened 2025-11-20
ASimAuditEvent - Missing fields in vimAuditEventMicrosoftSecurityEvents


andrewj-t:
**Describe the bug**
When running vimAuditEventMicrosoftSecurityEvents, Events related to Scheduled Task information results are missing the Value, NewValue, Object and ObjectType fields. These fields are present when running ASimAuditEventMicrosoftSecurityEvents

When 

**To Reproduce**
1. Connect to a Sentinel Environment which has ingested 4698 - A scheduled task was created Events
2. Execute the following tsql:
```kql
vimAuditEventMicrosoftSecurityEvents
| where EventOriginalType == 4698
```
3. Observer that the Object,ObjectType,NewValue and Value fields are not returned

**Expected behavior**
The fields should be returned, as they do when running the non-paramaterized version of the query

**Additional context**
I have created and fix and will submit a PR shortly

v-utpalkumar:
Closing this issue since the associated PR https://github.com/Azure/Azure-Sentinel/pull/13170 has been closed by the author.

PR #13170, opened 2025-11-20
Update ASimAuditEvent parsers to improve Scheduled Task Event Parsing

andrewj-t:
   
   Change(s):
   - Updated vimAuditEventMicrosoftSecurityEvents.yaml to include additional fields
   - Updated vimAuditEventMicrosoftSecurityEvents.yaml to correctly unescape TaskContent XML
   - Updated ASimAuditEventMicrosoftSecurityEvents.yaml.yaml to correctly unescape TaskContent XML

   Reason for Change(s):
   - Fix Issue https://github.com/Azure/Azure-Sentinel/issues/13168
   - Fix issue  https://github.com/Azure/Azure-Sentinel/issues/13169

   Version Updated:
   - Yes, Updated to `0.2.2`

   Testing Completed:
   - Yes

   - Yes


andrewj-t:

v-atulyadav:
Hi @andrewj-t,
Please upload a sample data. Thanks

v-atulyadav:
Hi @andrewj-t,
Validations are failing due to the issues listed below. Please include the necessary sample data file. Thanks

image

image

Sample data file path: https://github.com/Azure/Azure-Sentinel/tree/master/Sample%20Data/ASIM

andrewj-t:
Hi @v-atulyadav 

I will look to generate some sample data that covers off the scheduled task related event IDs and will add to the PR.

As for the replace_strings error, this may be an issue in the KqlValidator not being up to date with the latest functions available in the Kql language. It is documented as a supported function [here](https://learn.microsoft.com/en-us/kusto/query/replace-strings-function?view=microsoft-sentinel) and I am able to use it in my production Sentinel Workspace sucessfully. There is the older replace_string function, but I thought it would be more performant to use replace_strings, rather than calling replace_string 3 times for every row processed

v-atulyadav:
Hi @andrewj-t,

It appears that the KQL query fails when using the condition shown in brackets, but it succeeds when we replace it with the underlined condition.

image

Please verify at your end.

andrewj-t:
Hi @v-atulyadav 

I am having problems generating sample logs to match the existing Microsoft_Windows_AuditEvent_SecurityEvent_IngestedLogs.csv in the Sample data directory. It seems the columns may have changed since the sample data was generated previously.

I have emailed you directly with the results of the query from my company's internal environment for an example.
```
SecurityEvent
| where EventID in (4698,4699)
```

If you have access to an internal tool that can sanitize it and match to the existing data please go ahead and use it.. I hope this is acceptable to progress the PR, if not, let me know and I will withdraw it as I don't think I will be able to have time to be able to regenerate the sample data.

Parsers/ASimAuditEvent/Parsers/vimAuditEventMicrosoftSecurityEvents.yaml
```diff
@@ -272,7 +272,7 @@ ParserQuery: |
         | extend
             User = ActorUsername,
             Dvc = DvcFQDN
-        | project-away Subject*, EventID, Computer,NewValue,ObjectType,Object,OldValue,Value
```
hunngu-ms, line 275: NewValue and OldValue should be project-away still. Their values are assigned to Value.

andrewj-t: That would re-introduce the issue in described here - https://github.com/Azure/Azure-Sentinel/issues/13168

This change simply matches the project-away statement on the similiar parsers and makes their behaviour consistent.

[vimAuditEventMicrosoftWindowsEvents](https://github.com/Azure/Azure-Sentinel/blob/fcd9ddf16d06762570ee0c7395d9c9bf6b7f49df/Parsers/ASimAuditEvent/Parsers/vimAuditEventMicrosoftWindowsEvents.yaml#L269)
[ASimAuditEventMicrosoftSecurityEvents](https://github.com/Azure/Azure-Sentinel/blob/fcd9ddf16d06762570ee0c7395d9c9bf6b7f49df/Parsers/ASimAuditEvent/Parsers/ASimAuditEventMicrosoftSecurityEvents.yaml#L199)
[ASimAuditEventMicrosoftWindowsEvents](https://github.com/Azure/Azure-Sentinel/blob/fcd9ddf16d06762570ee0c7395d9c9bf6b7f49df/Parsers/ASimAuditEvent/Parsers/ASimAuditEventMicrosoftWindowsEvents.yaml#L194)

PR #13259, opened 2025-12-06
Adding other Assets support for Watchlist

vakohl:
   
   Change(s):
   - Added logic in Alerts and Incidents to support additional Entity Types as personal data assets

   Reason for Change(s):
   - Previously only Hostname/Devices were supported as personal data assets. Now included Azure Resources, AWS, Blobs etc.

   Version Updated:
   - NA

   Testing Completed:
   - Yes

   - Yes


manishkumar1991 (approved):
approved

PR #13409, opened 2026-01-08
[ASIM] Authentication AADSigninLogs parser rewrite

yummyblabla:
Additions:
- LogonMethod
- ActingAppName
- TargetAppId

Removals:
- Unnormalized columns


Parsers/ASimAuthentication/Parsers/ASimAuthenticationAADSigninLogs.yaml
```diff
@@ -56,56 +56,108 @@ ParserQuery: |
...
+    | lookup FailedReason on ResultType
+    | project-rename
+        EventOriginalUid = Id,
+        HttpUserAgent    = UserAgent,
+        TargetAppId      = AppId,
+        TargetAppName    = AppDisplayName,
+        TargetSessionId  = CorrelationId,
+        TargetUserId     = UserId,
+        TargetUsername   = UserPrincipalName,
+        TargetOriginalAppType = ClientAppUsed
```
oshezaf: You map client app (which is source) to targetapp. This also make the type lookup incorrect.

Parsers/ASimAuthentication/Parsers/ASimAuthenticationAADSigninLogs.yaml
```diff
@@ -56,56 +56,108 @@ ParserQuery: |
...
+        SrcHostname = tostring(DeviceDetail.displayName),
+        SrcDvcId = tostring(DeviceDetail.deviceId),
+        SrcDvcIdType = "",
+        SrcIpAddr = IPAddress,
+        SrcDvcOs = tostring(DeviceDetail.operatingSystem),
+        TargetUserIdType = 'AADID',
+        TargetUsernameType = 'UPN',
+        LogonMethod = coalesce(AuthenticationMethodsUsed, AuthenticationRequirement)
+    | extend
+        SrcDvcIdType = iif(isempty(SrcDvcId), "", "MDEId"),
```
oshezaf: This would be strange. Why would Entra log have MDE device IDs? Well, maybe, but worth verifying.

Parsers/ASimAuthentication/Parsers/ASimAuthenticationAADSigninLogs.yaml
```diff
@@ -56,56 +56,108 @@ ParserQuery: |
...
+        Type = 'SigninLogs',
+        Location = todynamic(LocationDetails),
+        SrcHostname = tostring(DeviceDetail.displayName),
+        SrcDvcId = tostring(DeviceDetail.deviceId),
+        SrcDvcIdType = "",
+        SrcIpAddr = IPAddress,
+        SrcDvcOs = tostring(DeviceDetail.operatingSystem),
+        TargetUserIdType = 'AADID',
+        TargetUsernameType = 'UPN',
+        LogonMethod = coalesce(AuthenticationMethodsUsed, AuthenticationRequirement)
```
oshezaf: Just a comment: I am not happy with the fact that LogonMethod is not normalized. I assume it was requested by research, but if not normalized, how will it help them?

yummyblabla: https://learn.microsoft.com/en-us/azure/sentinel/normalization-schema-authentication#authentication-specific-fields

LogonMethod is an optional string, and not an enumeration. Perhaps you would want some enumerations enforced here..?

Parsers/ASimAuthentication/Parsers/ASimAuthenticationAADSigninLogs.yaml
yummyblabla: Good catch. It should be "Other"

Parsers/ASimAuthentication/Parsers/ASimAuthenticationAADSigninLogs.yaml
yummyblabla: Not sure if I follow.
We have source `ClientAppUsed`, which is mapped to `TargetOriginalAppType`, which is also used to map to the normalized `TargetAppType` column.
We will have both `TargetOriginalAppType` and `TargetAppType`

Parsers/ASimAuthentication/Parsers/ASimAuthenticationAADSigninLogs.yaml
yummyblabla: Normalized it, but we should look into updating the docs as well.

Parsers/ASimAuthentication/Parsers/ASimAuthenticationAADSigninLogs.yaml
oshezaf: I believe that ClientAppUsed is ActingApp and not TargetApp

Parsers/ASimAuthentication/Parsers/ASimAuthenticationAADSigninLogs.yaml
oshezaf: Makes sense. Can you summarize for me the changes I need to make?

oshezaf:
@yummyblabla : one comment still open about the acting vs. target app.

Parsers/ASimAuthentication/Parsers/ASimAuthenticationAADSigninLogs.yaml
yummyblabla: Changes made

oshezaf (approved):

PR #13411, opened 2026-01-09
[ASIM] ProcessEvent - Microsoft365D Parser changes

yummyblabla:
Fixes:
- Use project instead of project-away
- Change use of TimeGenerated to Timestamp instead

Additions:
- TargetUsernameType
- ActorUsernameType

Removal:
- Unnormalized column: MDE-MachineGroup


Parsers/ASimProcessEvent/Parsers/vimProcessEventMicrosoft365D.yaml
```diff
@@ -79,8 +79,8 @@ ParserQuery: |
     DeviceProcessEvents 
     // -- pre-filtering
     | where
-      (isnull(starttime)   or TimeGenerated >= starttime )
-      and (isnull(endtime) or TimeGenerated <= endtime )
+      (isnull(starttime)   or Timestamp >= starttime )
```
oshezaf, line 82: You coalesce later on, which is inconsistent with the filter here. 

I think (but need to really think it over):

isnull(starttime) or isnull(timestamp) or TimeGenerated > startime 
and (isnull(starttime) or not(isnull(timestamp)) or timestamp > starttime

yummyblabla: Removed the coalesce. As for Timestamp or TimeGenerated, we should probably be consistent and go with the documentation[Timestamp] (it has Timestamp and not TimeGenerated)
https://learn.microsoft.com/en-us/defender-xdr/advanced-hunting-deviceprocessevents-table

PR #13418, opened 2026-01-09
[ASIM] AlertEvent - Microsoft Defender XDR Parser updates

yummyblabla:
Refer to changelog

### DIFF
--- a/Parsers/ASimAlertEvent/CHANGELOG/ASimAlertEventMicrosoftDefenderXDR.md
+++ b/Parsers/ASimAlertEvent/CHANGELOG/ASimAlertEventMicrosoftDefenderXDR.md
@@ -1,5 +1,11 @@
 # Changelog for ASimAlertEventMicrosoftDefenderXDR.yaml
 
+## Version 0.2.0
+
+- (2026-01-09) [ASIM] AlertEvent - Microsoft Defender XDR Parser updates - [PR #13418](https://github.com/Azure/Azure-Sentinel/pull/13418)
+- Removal of unnormalized columns
+- Add value: `IpAddresses` to AdditionalFields
+
 ## Version 0.1.0
 
 - (2024-11-07) AlertSchemaParsers - [PR #11387](https://github.com/Azure/Azure-Sentinel/pull/11387)
--- a/Parsers/ASimAlertEvent/CHANGELOG/vimAlertEventMicrosoftDefenderXDR.md
+++ b/Parsers/ASimAlertEvent/CHANGELOG/vimAlertEventMicrosoftDefenderXDR.md
@@ -1,5 +1,12 @@
 # Changelog for vimAlertEventMicrosoftDefenderXDR.yaml
 
+## Version 0.2.0
+
+- (2026-01-09) [ASIM] AlertEvent - Microsoft Defender XDR Parser updates - [PR #13418](https://github.com/Azure/Azure-Sentinel/pull/13418)
+- Removal of unnormalized columns
+- Add value: `IpAddresses` to AdditionalFields
+- Add `ASimMatchingIpAddr` column to describe what IpAddress column was matched to
+
 ## Version 0.1.0
 
 - (2024-11-07) AlertSchemaParsers - [PR #11387](https://github.com/Azure/Azure-Sentinel/pull/11387)


Parsers/ASimAlertEvent/Parsers/ASimAlertEventMicrosoftDefenderXDR.yaml
```diff
@@ -54,157 +54,207 @@ ParserQuery: |
...
-            Account*,
-            ApplicationId,
-            Application,
-            *_s
+    disabled: bool=false) {
+    AlertEvidence
+    | where not(disabled)
+    // Mapping Inspection Fields
+    | extend AdditionalFields = todynamic(AdditionalFields)
+    | extend
```
oshezaf, line 64: project-rename for those applicable

Parsers/ASimAlertEvent/Parsers/ASimAlertEventMicrosoftDefenderXDR.yaml
```diff
@@ -54,157 +54,207 @@ ParserQuery: |
...
+    // Mapping Inspection Fields
+    | extend AdditionalFields = todynamic(AdditionalFields)
+    | extend 
+        EventUid = AlertId,
+        AlertName = Title,
+        AlertVerdict_Custom = tostring(AdditionalFields.ThreatAnalysisSummary[0].Verdict),
+        AlertVerdictDate_s = todatetime(AdditionalFields.ThreatAnalysisSummary[0].AnalysisDate),
+        AttackTactics = iff(Categories has_any (AttackTacticSet), replace_regex(Categories, @"[\[\]\""]", ""), ""),
+        AlertOriginalStatus = tostring(AdditionalFields.LastRemediationState),
+        AlertStatus = iif(isnotempty(AdditionalFields.LastRemediationState), iif(AdditionalFields.LastRemediationState == "Active", "Active", "Closed"), ""),
```
oshezaf: case statement, and can use AlertOriginalStatus  (splitting to to extend of course)

Parsers/ASimAlertEvent/Parsers/ASimAlertEventMicrosoftDefenderXDR.yaml
```diff
@@ -54,157 +54,207 @@ ParserQuery: |
...
+        HttpUserAgent_s = tostring(AdditionalFields.UserAgent)
+    | extend
+        UserIdType = iif(isnotempty(UserId), "EntraUserID", iif(isnotempty(UserSid), "SID", "")),
+        UserId = coalesce(UserId, UserSid),
+        UserType = _ASIM_GetUserType(Username, UserSid),
+        UsernameType = _ASIM_GetUsernameType(Username)
+    // Mapping Device Entity
+    | extend 
+        DvcId = coalesce(DeviceId, tostring(AdditionalFields.Host.MachineId)),
+        DvcIpAddr = coalesce(LocalIP, tostring(AdditionalFields.Host.IpInterfaces[0].Address), RemoteIP),
```
oshezaf, line 98: I don't see how RemoteIP can be a DevIpAddr

Parsers/ASimAlertEvent/Parsers/ASimAlertEventMicrosoftDefenderXDR.yaml
```diff
@@ -54,157 +54,207 @@ ParserQuery: |
...
+        DvcOs = tostring(coalesce(AdditionalFields.OSFamily, AdditionalFields.Host.OSFamily)),
+        DvcOsVersion = tostring(coalesce(AdditionalFields.OSVersion, AdditionalFields.Host.OSVersion)),
+        DeviceName = coalesce(DeviceName, tostring(AdditionalFields.Host.NetBiosName)),
+        DvcScopeId = coalesce(tostring(split(AdditionalFields.AzureID, "/")[2]), (tostring(split(AdditionalFields.ResourceId, "/")[2])))
+    | extend DvcIdType = iif(isnotempty(DvcId), "MDEid", "")
```
oshezaf: I suspect this would depend on the DvcId selected by the coalesce operator above

Parsers/ASimAlertEvent/Parsers/vimAlertEventMicrosoftDefenderXDR.yaml
```diff
@@ -51,225 +51,279 @@ ParserParams:
...
+            attacktactics_has_any: dynamic=dynamic([]),
+            attacktechniques_has_any: dynamic=dynamic([]),
+            threatcategory_has_any: dynamic=dynamic([]),
+            alertverdict_has_any: dynamic=dynamic([]),
+            eventseverity_has_any: dynamic=dynamic([]),
+            disabled: bool=false) {
+      AlertEvidence
+      | where not(disabled)
+      // Mapping Inspection Fields
+      | extend AdditionalFields = todynamic(AdditionalFields)
```
oshezaf: You don't parse JSON (which is what todynamic does) prior to pre-filtering. You prefilter on the string and post verify after the parsing.

yummyblabla: Artifact of the RDA table treating this column as string and not dynamic. Will remove.
https://learn.microsoft.com/en-us/azure/azure-monitor/reference/tables/alertevidence

Parsers/ASimAlertEvent/Parsers/ASimAlertEventMicrosoftDefenderXDR.yaml
yummyblabla: Addressed

Parsers/ASimAlertEvent/Parsers/ASimAlertEventMicrosoftDefenderXDR.yaml
yummyblabla: Addressed

Parsers/ASimAlertEvent/Parsers/ASimAlertEventMicrosoftDefenderXDR.yaml
yummyblabla: Checking with sample data, both are the same. However, it should be FQDN, and not MDEid

According to doc, 
DeviceName | string | Fully qualified domain name (FQDN) of the machine.

https://learn.microsoft.com/en-us/azure/azure-monitor/reference/tables/alertevidence

Parsers/ASimAlertEvent/Parsers/ASimAlertEventMicrosoftDefenderXDR.yaml
yummyblabla: Seems like the fault of the AlertEvent schema. It does not contain DstIpAddr, where we could potentially map RemoteIP to

Parsers/ASimAlertEvent/Parsers/vimAlertEventMicrosoftDefenderXDR.yaml
```diff
@@ -97,7 +97,6 @@ ParserQuery: |
       AlertEvidence
       | where not(disabled)
       // Mapping Inspection Fields
-      | extend AdditionalFields = todynamic(AdditionalFields)
       | where (isnull(starttime) or Timestamp >= starttime)
           and (isnull(endtime) or Timestamp <= endtime)
           and ((array_length(ipaddr_has_any_prefix) == 0) or (has_any_ipv4_prefix(LocalIP, ipaddr_has_any_prefix)) or (has_any_ipv4_prefix(tostring(AdditionalFields.Host.IpInterfaces[0].Address), ipaddr_has_any_prefix)) or (has_any_ipv4_prefix(RemoteIP, ipaddr_has_any_prefix)))
```
oshezaf, line 102: Was the field always dynamic? So you could just remove the todynamic? Anyways, the efficient way to do the prefiltering is not just to not convert to dynamic, but to actually use test over the entire string. Otherwise, you are back to calculated fields conditions.

yummyblabla: The field was always dynamic. It is string in RDA

yummyblabla: Comparing to WindowsEvent where EventData is dynamic, the filtering performance here should be similar.'

https://github.com/Azure/Azure-Sentinel/blob/master/Parsers/ASimFileEvent/Parsers/vimFileEventMicrosoftSysmonWindowsEvent.yaml

Parsers/ASimAlertEvent/Parsers/ASimAlertEventMicrosoftDefenderXDR.yaml
oshezaf: I suspect RemoteIP is SrcIpAddr

oshezaf (changes requested):
Two comments: 
1. Requiring change - we again have here the issue of searching in a value that may not get into the normalized event (i.e. RemoteIP being matched by not selected as DcvIpAddr. Let's use again the AdditionalFields solution.
2. No change required, but please look into string vs. dynamic "has" performance.

yummyblabla:
> Two comments:
> 
> 1. Requiring change - we again have here the issue of searching in a value that may not get into the normalized event (i.e. RemoteIP being matched by not selected as DcvIpAddr. Let's use again the AdditionalFields solution.
> 2. No change required, but please look into string vs. dynamic "has" performance.

Addressed first item.
Second item put in backlog

Parsers/ASimAlertEvent/Parsers/vimAlertEventMicrosoftDefenderXDR.yaml
```diff
@@ -203,6 +203,9 @@ ParserQuery: |
       // Mapping Email Entity
       | extend
           EmailSubject
+      // Creating IpAddress list in AdditionalFields
+      | extend IpAddresses = pack_array(RemoteIP, LocalIP, tostring(AdditionalFields.Host.IpInterfaces[0].Address))
+      | mv-apply IpAddress = IpAddresses on (where isnotempty(IpAddress) | summarize IpAddresses = make_list(IpAddress))
```
oshezaf: 1. Isn't this just make_set?
2. Didn't we do the previous one with also assigning ASimMatchingIpAddress to AdditionalFields if the value was only there?

yummyblabla: 1. make_set is an aggregation function that is only used with summarize, but yes, make_list should be updated to make_set in case there are duplicates.
2. It looks like a schema difference where parsers of the same schema have ASimMatching* as a column.
Authentication parsers have ASimMatchingUsername
Network and WebSession have ASimMatchingIpAddr

I will add it to this parser.

yummyblabla: Addressed

PR #13419, opened 2026-01-09
[ASIM] AuditEvent - AzureAdminActivity Parser changes

yummyblabla:
Fixes:
- Enumeration for EventType (existing AzureActivity table already has EventTypecolumn)
- Enumeration for EventResult (Other does not exist, replace with NA)
- Removal of unnormalized columns (eg. ActorUserUpn)


Parsers/ASimAuditEvent/Parsers/ASimAuditEventAzureAdminActivity.yaml
```diff
@@ -74,23 +74,23 @@ ParserQuery: |
...
     | where CategoryValue == "Administrative"
-    | project-away HTTPRequest, Level, SourceSystem, EventSubmissionTimestamp, TenantId, OperationId, Hierarchy, Category, ResourceId, ResourceProvider, Resource
     | project-rename 
         Operation = OperationNameValue,
         SrcIpAddr = CallerIpAddress,
         EventOriginalUid = EventDataId,
         ActorSessionId = CorrelationId,
         EventOriginalType = CategoryValue
     | extend
+        Type = "AzureActivity",
```
oshezaf, line 87: Question and thought for the future: 
- I assume we need to add those across the board to support RDA? 
- They should be in by default for LA workspaces
- Non calculated column is always best to keep :-().
-  So thinking about it, maybe we need some automation to extend the LA parser to RDA (ItemId, Type etc.)

Parsers/ASimAuditEvent/Parsers/ASimAuditEventAzureAdminActivity.yaml
```diff
@@ -130,15 +128,13 @@ ParserQuery: |
...
     // --
     // Calculate EventType
     | extend op = toupper(tostring(split(Operation,"/")[-1]))
     | lookup AzureActivityOperationLookup on op
-    | extend EventType = iff (EventType == "", "Other", EventType)
-    | project-away op
+    | extend EventType = iff (AzureEventType == "", "Other", AzureEventType)
     // Aliases
-    | extend AdditionalFields = pack_dictionary("Authorization", Authorization_d, "Claims", Claims_d, "Error", Properties_d.statusMessage)
+    | extend AdditionalFields = bag_pack("Authorization", Authorization_d, "Claims", Claims_d, "Error", Properties_d.statusMessage)
```
oshezaf, line 137: Going through a few parsers, I found that I have added a pack parameter in some cases to handle the inefficiency of bag_pack. It is documented here: https://learn.microsoft.com/en-us/azure/sentinel/normalization-about-parsers#the-pack-parameter. So we should use it.

Parsers/ASimAuditEvent/Parsers/ASimAuditEventAzureAdminActivity.yaml
```diff
@@ -147,9 +143,43 @@ ParserQuery: |
         Dst = TargetAppName,
         Src = SrcIpAddr,
     // -- Entity identifier explicit aliases
-        ActorUserUpn = ActorUsername,
```
oshezaf, line 150: I think that ActorUserUpn is normalized (Similar to ActorUserAadId). It is a thorny issue. Let's discuss in our meeting today.

Parsers/ASimAuditEvent/Parsers/vimAuditEventAzureAdminActivity.yaml
```diff
@@ -113,27 +113,18 @@ ParserQuery: |
...
         (isnull(starttime) or TimeGenerated >= starttime) 
         and (isnull(endtime) or TimeGenerated <= endtime)
         and (array_length(newvalue_has_any) == 0)
     | where CategoryValue == "Administrative"
-    | project-away HTTPRequest, Level, SourceSystem, EventSubmissionTimestamp, TenantId, OperationId, Hierarchy, Category, ResourceId, ResourceProvider, Resource
     | where 
         (array_length(srcipaddr_has_any_prefix) == 0 or has_any_ipv4_prefix(CallerIpAddress,srcipaddr_has_any_prefix))
         and (array_length(actorusername_has_any) == 0 or Caller has_any (actorusername_has_any))
         and (array_length(operation_has_any) == 0 or OperationNameValue has_any (operation_has_any))
         and (array_length(object_has_any) == 0 or Properties  has_any (object_has_any))
```
oshezaf, line 127: There should be a post parsing filtering for object. The others seems fine as they check for atomic fields.

Parsers/ASimAuditEvent/Parsers/ASimAuditEventAzureAdminActivity.yaml
yummyblabla: Added to the csv tester based on those extra columns:
https://learn.microsoft.com/en-us/azure/sentinel/normalization-entity-user#the-user-id-and-scope

Parsers/ASimAuditEvent/Parsers/ASimAuditEventAzureAdminActivity.yaml
yummyblabla: I will need to review all the schemas/parsers we have and see how we can ensure that this argument is passed to them all. There is a huge gap on how this is standardized currently...

Parsers/ASimAuditEvent/Parsers/vimAuditEventAzureAdminActivity.yaml
yummyblabla: Addressed

PR #13425, opened 2026-01-12
Update ZscalerZPAUnexpectedSessionDuration analytic rule for Zscaler Private Access(ZPA)

### LINKED ISSUE #12588: Zscaler Private Access (ZPA) - Unexpected session duration encountered
**Describe the bug**
Analytics rule appears to be no longer valid.
https://github.com/Azure/Azure-Sentinel/blob/master/Solutions/Zscaler%20Private%20Access%20(ZPA)/Analytic%20Rules/ZscalerZPAUnexpectedSessionDuration.yaml

**To Reproduce**
1. Configure ZPA timeout policy (e.g. 4 hours)
2. Enable analytics rule on Sentinel
3. Wait for lookback period (e.g. 24 hours)
4. Multiple alerts triggered by various users (e.g. Session ID would be the same for each user)

**Expected behavior**
Few or none alerts triggered

**Additional context**
Opened a case with Zscaler Support (Case Number: 05699828) on same Session ID and their response:
> Unfortunately, there is currently no method to track user reauthentication events based on the 4-hour timeout policy using the User Status or any other logs available within the Zscaler portal.

Pending Zscaler response if they can provide an update to this analytics rule.

v-kasghosh:
Change(s):
   - Updated the analytic rule logic in ZscalerZPAUnexpectedSessionDuration.yaml to use ConnectionID instead of NetworkSessionId.
   - Incremented solution versions to 3.0.4
   
   Reason for Change(s):
   - #12588 

   Version Updated:
   - 3.0.4


rahul0216 (approved):
Changes look good.

v-atulyadav (approved):

PR #13450, opened 2026-01-15
Set fetch-depth to 2 in workflow checkouts

rahul0216:
Change(s):
   - Updated all GitHub Actions workflows to use 'fetch-depth: 2' when checking out code. This ensures both HEAD and its parent commit are available for git diff operations, improving validation accuracy in pull request workflows and avoiding storage issues.

   Reason for Change(s):
   - To avoid storage exhaustion issue

   Version Updated:
   - NA

   Testing Completed:
   - Yes
   
   - Yes


ashishsyal (approved):
Approved

PR #13453, opened 2026-01-15
[ASIM] Authentication su parser fixes

yummyblabla:
Changes:
- Successful su events are now Logon EventType
- Added Failed su events
- Fix normalized columns and add Type explicitly.


Parsers/ASimAuthentication/Parsers/vimAuthenticationSu.yaml
```diff
@@ -59,154 +59,178 @@ ParserQuery: |
...
-          ,
-          TargetIpAddr          = DvcIpAddr
+    let prefilter = (T: (SyslogMessage: string, TimeGenerated: datetime, EventResult: string, EventType: string))
+    {
+        T
+        | where 
+            (isnull(starttime) or TimeGenerated >= starttime) 
+            and (isnull(endtime) or TimeGenerated <= endtime)
+            and ((array_length(username_has_any) == 0) or SyslogMessage has_any (username_has_any))
+            and (array_length(targetappname_has_any) == 0) // TargetAppName not available in source
```
oshezaf: I think that TargetAppName is "SU"

Parsers/ASimAuthentication/Parsers/vimAuthenticationSu.yaml
```diff
@@ -59,154 +59,178 @@ ParserQuery: |
...
+            and (array_length(srcipaddr_has_any_prefix) == 0) // SrcIpAddr not available in source
```
oshezaf: It goes back to schema definition, but I don't understand why there only src filter on IP address and hostname (and target on app). For SU, I would just make the source the same as target/dev to resolve.

Parsers/ASimAuthentication/Parsers/vimAuthenticationSu.yaml
yummyblabla: Addressed

Parsers/ASimAuthentication/Parsers/vimAuthenticationSu.yaml
yummyblabla: Addressed

yummyblabla:
> The solutions we added to filtering cannot be filtering only. You should map the values also to the relevant fields ('su' to target app and HostIP to DstIpAddr

Whoops. Added.

oshezaf (approved):

PR #13460, opened 2026-01-16
[ASIM] Authentication - Sshd Parser fixes

yummyblabla:
Changes:
- Add LogonMethod to accepted sshd events
- Coalesce Dvc
- Add Src Alias
- Add Type column explicitly


Parsers/ASimAuthentication/Parsers/ASimAuthenticationSshd.yaml
```diff
@@ -37,7 +37,8 @@ ParserQuery: |
             EventCount    = int(1),
             EventResult   = 'Success',
             EventSeverity = 'Informational',
-            EventType     = 'Logon'
+            EventType     = 'Logon',
+            LogonMethod   = iff (SyslogMessage has 'publickey', 'PKI', 'Username & password')
```
oshezaf: What about the 3rd option SNAM has raised? SyslogMessage has 'keyboard-interactive/pam', 'keyboard-interactive/pam',.

Also, since you are anyway extracting the method field in the parse statement you can use a lookup here. Especially if the list get longer.

Parsers/ASimAuthentication/Parsers/ASimAuthenticationSshd.yaml
```diff
@@ -191,8 +193,9 @@ ParserQuery: |
     // -- Aliases
     | extend
         Dst                   = coalesce (DvcFQDN, DvcHostname, DvcIpAddr),
-        Dvc                   = DvcHostname,
+        Dvc                   = coalesce (DvcFQDN, DvcHostname, DvcIpAddr),
```
oshezaf: You can extend Dst (separate extend statement needed)

Parsers/ASimAuthentication/Parsers/vimAuthenticationSshd.yaml
```diff
@@ -63,7 +63,7 @@ ParserQuery: |
     disabled: bool=false
     )
   {
-      let prefilter = (T: (SyslogMessage: string, TimeGenerated: datetime))
+  let prefilter = (T: (SyslogMessage: string, TimeGenerated: datetime))
```
oshezaf, line 72: - Repeating here the issue with only having Source IP/host as filters. Since we have them here, less critical. 
- Note that we did take target app to be sshd here (which we did not do for su)
- Also, worth thinking about adding LogonMethod parameter to the schema.

Parsers/ASimAuthentication/Parsers/ASimAuthenticationSshd.yaml
yummyblabla: Addressed. Looks like the previous PR just addressed missing logs, and not mapping LogonMethod.

This fix would also apply to failed logins.

Parsers/ASimAuthentication/Parsers/ASimAuthenticationSshd.yaml
yummyblabla: ADdressed

ASIM/dev/ASimTester/ASimTester.csv
```diff
@@ -831,7 +831,7 @@ IpAddr,string,Alias,FileEvent,IP Address,,SrcIpAddr
 IpAddr,string,Alias,NetworkSession,IP Address,,SrcIpAddr
 IpAddr,string,Alias,UserManagement,IP address,,SrcIpAddr
 IpAddr,string,Alias,WebSession,IP Address,,SrcIpAddr
-LogonMethod,string,Optional,Authentication,Enumerated,Managed Identity|Service Principal|Username & Password|Multi factor authentication|Passwordless|PKI|Other,
+LogonMethod,string,Optional,Authentication,Enumerated,Managed Identity|Service Principal|Username & Password|Multi factor authentication|Passwordless|PKI|Keyboard Interactive/PAM|Other,
```
oshezaf: While a judgement call, let's use "PAM" (see https://docs.ssh.com/manuals/server-admin/60/userauth-kbi.html)

yummyblabla: Addressed

oshezaf (approved):

PR #13472, opened 2026-01-20
[ASIM] NetworkSession - NTANetAnalytics - SrcIpAddr can map to SrcIp or SrcPublicIps

yummyblabla:
Previous implementation only mapped SrcIpAddr to SrcIp.

Data shows if SrcIp is empty, SrcPublicIps is populated, so this bug fix will fix SrcIpAddr being empty.

This problem also exists with DestPublicIps as well. The solution is applied to that field.


Parsers/ASimNetworkSession/Parsers/vimNetworkSessionNTANetAnalytics.yaml
```diff
@@ -91,7 +91,7 @@ ParserQuery: |
   // Pre-filter Event result
   | where eventresult == "*" or (eventresult == "Success" and EventResult == "Success") or (eventresult == "Failure" and EventResult == "Failure")
   // Pre-filter IPs
-  | extend temp_SrcMatch = has_any_ipv4_prefix(SrcIp, src_or_any), temp_DstMatch = has_any_ipv4_prefix(DestIp, dst_or_any)
+  | extend temp_SrcMatch = has_any_ipv4_prefix(SrcIp, src_or_any) or has_any_ipv4_prefix(SrcPublicIps, src_or_any), temp_DstMatch = has_any_ipv4_prefix(DestIp, dst_or_any)
```
oshezaf: There is a potential issue here: you may find the IP that was filtered by not in the first item in SrcPublicIps. Now, strictly speaking, this requires post parsing filtering. However, it might not be what the user wants. I do suggest maintaining the full list of IPs in AdditionalFields here in any case, and, if filtering selecting the SrcIpAddr as the one filtered for (which is hard, I know).

yummyblabla: Addressed. Though I acknowledge that we should use pack for this case to show AdditionalFields, but that will require a massive change in the union function. The usage of pack will need to be dealt with after this PR.

PR #13479, opened 2026-01-21
Filter out specific alert name from SecurityAlerts

hitem:
Strict filter to filter out its own inputs (else you have an endless loop of the analytic rule findings its own findings).

   Change(s):
   - Adding exclusion for itself in the title of the alert.

   Reason for Change(s):
   - The analytic rule is creating alerts of its own findings. When it triggers on a true possitive alert it creates and entry in the SecurityAlert, it then finds its own finding and you have an endless flow of new alerts from this TI-analytic rule.

   Version Updated:
   - 1.4.3 -> 1.4.4

   Testing Completed:
   - Indeed, i have run this in production after we have been flodded in 2-3 tenants by this alert - they are no longer being triggered. I also tested by sending the domain again via email, letting Defender for Office block the Domain and then we got 1 new SecurityAlert entry by this analytic rule - and not flooded afterwards. Making them high fidelity.

   - No issues present.


v-shukore:

v-shukore:
Hi @hitem, please accept CLA. Thanks!

hitem:
@microsoft-github-policy-service agree company="HitemSec"

hitem:
@microsoft-github-policy-service agree company="HitemSec"

PR #13509, opened 2026-01-27
Update version of Duo client to one that includes new CA bundle

m-ruiz21:
   
   Change(s):
   - Updated duo client version to 5.5.0 in the zip file (previously not packaged)
   - Updated python version to 3.12
   - Added example local.settings.json for ease of future development
   
Note: will add Solution version update later since discussion is needed here. This is an urgent change, and since there's no ARM templates that are being changed, simply reinstalling the same, current, solution will suffice. I want to just get a working, testable, version out there ASAP.

   Reason for Change(s):
   - Deprecation of duo client versions < 5.5.0 due to updated CA bundle. 

   Testing Completed:
   - Yes. Tested both the individual AzF and the complete ARM template installation flow. Functionality has also been confirmed by CXs who deployed it in their own environments.

   - N/A


m-ruiz21:

Solutions/CiscoDuoSecurity/Data Connectors/local.settings.example.json
jlheard, line 1: Curious, what's this file? Should it also be included in the zip?

PR #13513, opened 2026-01-28
Updated The Non Domain Controller Active Directory Replication Analytic Rule for Windows Security Events Solution

### LINKED ISSUE #12327: Issue with Reuse of SubjectLogonId & TargetLogonId in "Non Domain Controller Active Directory Replication"
**Describe the bug**
The "Non Domain Controller Active Directory Replication" Analytic Rule (AR) queries data over the last 7 days by default. For busy DCs, SubjectLogonId and TargetLogonId may get reused. When this occurs, it can result in erroneous IPs being added to the result because the join is only comparing the SubjectLogonId and TargetLogonId.

**To Reproduce**
Run the following query to observe reuse of TargetLogonId.
```
SecurityEvent
| where EventID == 4624 and LogonType == 3 and AccountType != 'Machine' and TargetLogonId != "0x3e7"
| summarize FirstUse = min(TimeGenerated), LastUse = max(TimeGenerated), Count = count() by TargetLogonId
| where LastUse - FirstUse >= 1d
| order by LastUse desc
```

Then run the following to observe different IP Addresses with the same TargetLogonId.
```
SecurityEvent
| where EventID == 4624 and LogonType == 3
| where AccountType != 'Machine'
| where TargetLogonId == "0xf107948f"
| project-reorder TimeGenerated, Account, Computer, IpAddress
```

**Expected behavior**
This AR "detects potential attempts by non-computer accounts (non domain controllers) to retrieve/synchronize an active directory object leveraging directory replication services (DRS)." Therefore, when this is detected, only legitimate Source IPs associated with traffic should be returned.

**Additional context**
Populating the DCServersList variable can help eliminate this issue. However, the EventIDs 4624 and 4662 should both occur on the same host. Therefore, modifying the join to also match the Computer would help ensure the LogonIds are associated with the same Computer. This could be further improved by also joining on Account but it would be necessary to ensure the format is the same between both Event IDs.
```
| join kind=leftouter
    (
    SecurityEvent
    | where Computer in (DCServersList)
    | where EventID == 4624 and LogonType == 3
    | where AccountType != 'Machine'
    | project TargetLogonId, IpAddress, Computer, Account
    )
    on $left.SubjectLogonId == $right.TargetLogonId, Computer
```

v-kasghosh:
Updated The Non-Domain Controller Active Directory Replication Analytic Rule for Windows Security Events Solution
   
   Change(s):
   - Updated The Non-Domain Controller Active Directory Replication Analytic Rule for Windows Security Events Solution
   - https://github.com/Azure/Azure-Sentinel/pull/13511

   Reason for Change(s):
   - #12327 

   Version Updated:
   -  Yes
   - 3.0.11
 

   Testing Completed:
   - Yes


v-sabiraj (commented):
I hope the changes were tested by customer?

v-sabiraj (approved):

PR #13517, opened 2026-01-28
Increasing the limit for large fields

dhanunjaya1054:
   
   Change(s):
   - There is a problem with the field limit in the incidents. We set the field limit to 1 MB according to platform support

   Reason for Change(s):
   - There is a problem with the field limit in the incidents. We set the field limit to 1 MB according to platform support

   Version Updated:
   - 3.0.8

   Testing Completed:
   - Yes


v-maheshbh:
Hi @dhanunjaya1054 

Kindly update the release notes with the latest version.

Thanks!

dhanunjaya1054:
@v-maheshbh - Updated the release notes. Thanks!

PR #13519, opened 2026-01-29
[ASIM] Authentication - sudo - EventVendor change

yummyblabla:
EventVendor change to Linux


yummyblabla:
Working on larger fix.

PR #13529, opened 2026-01-29
[ASIM] Authentication - sudo Parser changes

yummyblabla:
- Add normalization of `SeverityLevel` to `EventSeverity`
- Add mapping of `SeverityLevel` to `EventOriginalSeverity`
- Remove unnormalized columns
- Add mapping of `HostIP` to `SrcIpAddr`
- Add mapping of `ProcessName` (sudo) to `TargetAppName`, `Application`, `ActingAppName`
- Add mapping of `ProcessID` to `ActingAppId`
- Add Alias `Src`, `Dvc`, `IpAddr`
- Correct mapping of EventProduct to `Linux`
- Improve parser by removing code duplication
- Improve filtering process of vim* parser


oshezaf (commented):
Looks good. Did you check if we ever mapped Severity Level in any syslog parser? If we did so, we need to be consistent. Also, I would add an ASIM function (not a blocker, but on the backlog) to use for any Syslog parser.

yummyblabla:
> Looks good. Did you check if we ever mapped Severity Level in any syslog parser? If we did so, we need to be consistent. Also, I would add an ASIM function (not a blocker, but on the backlog) to use for any Syslog parser.

Yes there is. I will add it to the backlog.

PR #13570, opened 2026-02-04
feedly: migrate to ccf connector

Edouard360:

Change(s):
Add CCF files and updated mainTemplate.json accordindly

Reason for Change(s):
Migrating to CCF connector for simpler architecture

Version Updated:
Yes

Testing Completed:
Yes

image


v-shukore:
Hi @Edouard360, could you please resolve the arm-ttk failures. Thanks!

Edouard360:
@v-shukore should be good now!

Edouard360:
@v-shukore apologies, could you run the tests one last time? Thank you for your patience 🤗 
I must admit before I hadn't run the full arm-ttk test so I missed additional failures, but I now ran:
`docker build --build-arg SolutionName=Feedly --build-arg mainTemplateChanged=True --build-arg createUiChanged=True --file ./.github/actions/Dockerfile .` and all seems to be passing

Edouard360:
@microsoft-github-policy-service agree company="Feedly"

v-shukore:
Hi @Edouard360,
I noticed during solution testing that after deploying the maintemplate and opening the solution in the Content Hub, the newly added CCF DC does not appear correctly. The Content Hub is not picking up the expected name and description for the DC.
Please refer to the screenshot below for details. This might be related to the packaging, which may not have been done correctly. Could you please repackage the solution using the V3 tool and verify once?
image
Also, update the solution version in Data file.
Thanks!

Edouard360:
@v-shukore thank you for the review, I updated accordingly
- rebuilt the mainTemplate.json using the V3 tool
- updated version in Data/Solution_Feedly.json

image

Edouard360:
Thank you for launching the tests @v-shukore I see they all passed 🙌

Is testing on the Content Hub something that only you can do on your side? Happy to help with the additional testing if needed, I'm only familiar with the "Build your own template in the editor" + "Load file" path where I paste in the `mainTemplate.json`, which I think does not get me to that same exact view you're sharing

Screen Shot 2026-02-19 at 9 50 52 PM


image

v-shukore:
Hi @Edouard360,
All checks are passing, however I have a small question regarding the package version. I noticed that the master branch currently has version 3.0.6 for this solution (as shown in the screenshot below).
Based on my understanding, the package version should be incremented when changes are made. In this PR, it appears that the existing package was removed and the version was reduced to 3.0.4. we cannot delete the existing package. Could you please help confirm if this version change is intentional?
Thanks for clarifying.
image

Edouard360:
@v-shukore really appreciate your diligence, and would love to confirm that point with you indeed.

I never actually published on the content hub the versions 3.0.4, 3.0.5, 3.0.6.
I think that's why, when regenerating with the V3 tool it chose to pick up 3.0.4 as the "next" version.
I am totally fine with overwriting manually to 3.0.7, but I assumed that, given the previous versions were never actually published and will not be useful in the future either, I could just delete them and use 3.0.4, so that the package is "cleaner". What do you think?

v-shukore:
Thanks @Edouard360 for the confirmation. I’ll check this internally and confirm whether it works for us.

v-shukore:
Hi @Edouard360, we are good to go with this change I will review the content and will merge the PR. Thanks!

Edouard360:
amazing! thank you very much!

v-shukore (approved):

v-atulyadav (approved):

PR #13576, opened 2026-02-05
[Rapid7InsightVM] - fix package versions

srikarshastry:
Change(s):
   - updating package versions within the az function and also include `cryptography==42.0.8` package

   Reason for Change(s):
   - function app extension build 3.x is deprecated; 4.x version requires updated packages

   Version Updated: N/A

   Testing Completed:
   - I have tested in my ws and it is working


srikarshastry:

PR #13579, opened 2026-02-05
Feature/amirsasson/fix dataconnector name logic

AmirSasson:
Fix packaging script to respect when the Poller config already specifies the full name of the resource.
in addition respect [Condition ](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-tutorial-use-conditions) and Arm [Copy](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/copy-resources)


Tools/Create-Azure-Sentinel-Solution/common/createCCPConnector.ps1
```diff
@@ -617,6 +626,14 @@ function createCCPConnectorResources($contentResourceDetails, $dataFileMetadata,
                     $armResource.type = "Microsoft.OperationalInsights/workspaces/providers/dataConnectors"
                     $armResource.kind = $ccpItem.PollerKind;
 
+                    if ($null -ne $fileContent.condition) {
+                        $armResource | Add-Member -MemberType NoteProperty -Name "condition" -Value $fileContent.condition
```
jlheard, line 630: Thanks for addressing this!!!

PR #13688, opened 2026-02-23
TacitRed-SentinelOne v3.0.2: Fix hardcoded SentinelOne_BaseUrl default causing deployment failure

mazamizo21:
## Summary

Fixes a critical bug present since v1.0.0: the `SentinelOne_BaseUrl` parameter had a hardcoded `defaultValue` of `https://usea1-001.sentinelone.net`, which is a non-existent placeholder URL.

Any customer deploying from Content Hub without explicitly changing this parameter would get a connection timeout on every playbook run (`Post IOC to SentinelOne` step fails with host unreachable).

Verified: `curl https://usea1-001.sentinelone.net` → HTTP 000 (no server at this address). By contrast, real SentinelOne console URLs like `usea1-021`, `usea1-022`, `usea1-050`, `euc1-001` all return HTTP 200. Every SentinelOne customer is assigned their own unique management console subdomain at sign-up.

## Root Cause

The `defaultValue` was set to `https://usea1-001.sentinelone.net` from the very first commit (Dec 8, 2025) and was never corrected through any subsequent PR. It looks like a real SentinelOne URL but the `usea1-001` subdomain does not exist and has never been accessible.

## Changes

- **mainTemplate.json**: Clear `defaultValue` for `SentinelOne_BaseUrl` to `""` (both occurrences — parameter definition and Logic App definition)
- **mainTemplate.json**: Update description to guide customers: *"SentinelOne Console URL (e.g. https://usea1-021.sentinelone.net) — find this in your browser address bar when logged into SentinelOne. Must include https:// with no trailing slash."*
- **createUiDefinition.json**: Update placeholder from `usea1-001` to `YOUR-CONSOLE`
- **Package/3.0.2.zip**: Rebuilt with fixed templates
- **Data/Solution_TacitRedSentinelOneAutomation.json**: Bumped to 3.0.2
- **ReleaseNotes.md**: Added 3.0.2 entry

## Testing

- `defaultValue` for `SentinelOne_BaseUrl` is empty string in both template locations
- Zero references to `usea1-001` remain anywhere in the package
- Version correctly reports 3.0.2
- Existing Content Hub deployment with real credentials and correct URL confirms the playbook flow works end-to-end


mazamizo21:
Thanks for flagging this! You're right — the package was hand-edited instead of generated by the V3 tool, causing the mainTemplate to still show version 3.0.1.

**Fix applied:**
- Ran `createSolutionV3.ps1` against the solution data folder
- V3 tool regenerated `mainTemplate.json` with correct `contentPackages` version **3.0.2**
- Regenerated `3.0.2.zip` with proper V3 tool output

Pushed now. Please re-review when you get a chance. Thanks!

mazamizo21:
Update: replaced V3 tool approach with a surgical fix to avoid unintended changes.

Only modified:
- `mainTemplate.json` → `contentPackages.version`: `3.0.1` → `3.0.2`
- Rebuilt `3.0.2.zip` to match

All other content preserved exactly as-is (BaseUrl fix, description, tags, etc). Pushed.

v-shukore:
Hi @mazamizo21, please update any necessary content first, then use the V3 tool to repackage. This way, no unnecessary changes will be made to your solution. If you update only the maintemplate manually and increase the version without changing the content, it may cause problems in the future. Thanks!

mazamizo21:
Hi @v-shukore, thanks for the feedback. The challenge here is that the V3 tool (createSolutionV3.ps1) regenerates the mainTemplate from scratch, which overwrites the actual fix in this PR — clearing the hardcoded `SentinelOne_BaseUrl` defaultValue that was causing deployment failures.

We tested V3 on this solution previously and it reverted our BaseUrl fix along with other manual corrections (tags, descriptions, etc).

Would it be acceptable to keep the surgical approach for this PR since the core change is specifically to a defaultValue in mainTemplate.json? Or is there a way to make V3 preserve custom parameter defaults? Happy to adjust our approach based on your guidance. Thanks!

PR #13739, opened 2026-03-03
clean up feedly directory and substitute azuredeploy

Edouard360:
The CCP connector already replaces all custom polling logic previously handled by the Azure Function. There is no longer any need to provision or maintain a Function App, App Service Plan, or storage account. The Sentinel platform now handles scheduling, authentication, pagination, and ingestion natively.

This PR removes all files associated with the old function-based connector (including `Feedly_API_AzureFunctionApp.json` and `azuredeploy_Connector_Feedly_AzureFunction.json`). Instead, we only add `azuredeploy_Connector_Feedly_CCP.json`, a clean, standalone ARM deployment template using the native Sentinel RestApiPoller (CCP).

**Testing:**
Connector deployed and validated against the Feedly IoC API. Data flowing into `feedly_indicators_CL` as expected.


Edouard360:
need a second pass at testing `azuredeploy_Connector_Feedly_CCP.json` for the dce / dcr setup to be more aligned with conventions

PR #13765, opened 2026-03-06
Logstash Plugin - Set retransmission delay as parameter

lsoumille:
Change(s):
   - @RETRANSMISSION_DELAY is now taken from plugin parameters instead of being hardcoded
   - Documentation and release note updated

   Reason for Change(s):
   - We want to increase the retry delay during throttling to prevent further throttling. 

   Version Updated:
   - Yes minor update to 1.2.1 to be published to rubygems

   Testing Completed:
   - Yes on our own logstash instances

   - Yes


lsoumille:

v-shukore:
Hi @lsoumille, please share running connector screenshots here. Thanks!

lsoumille:
The connector is a logstash plugin. It doesn't have an UI and it's not available in Sentinel portal.

PR #13775, opened 2026-03-09
Updated offerId in SolutionMetadata.json

omkarj-metron:
## Required items, please complete

**Change(s):**
- Updated the offer ID in SolutionMetadata.json to align with the new offer in the Partner Portal.

**Reason for Change(s):**
- Our client had already created an offer with the same ID in their Partner Portal, and we were not aware of this when we initially set it up. We are updating the offer ID so that we can create a new offer in their Partner Portal.

**Version Updated:**
- Yes

**Testing Completed:**
- Yes

**Checked that the validations are passing and have addressed any issues that are present:**
- Yes


v-shukore:
Hi @omkarj-metron, please package the solution using V3 tool to reflect new offer id everywhere in solution
https://github.com/Azure/Azure-Sentinel/blob/master/Tools/Create-Azure-Sentinel-Solution/V3/README.md.
Thanks!

omkarj-metron:
Hi @v-shukore,
I have made the required change, plz check and let me know if anything is missing.

PR #13789, opened 2026-03-10
fix template file and consistency between connector and main template

Shams-Z:
   
   Change(s):
   - fix template file and consistency between connector and main template

   Reason for Change(s):
   - Installation issue

   Version Updated:
   - No

   Testing Completed:
   - Need help

   - Need help


v-atulyadav:
Hi @sreedharande,

We also need the previous versions, but in this PR the zip for version 3.0.0 has been removed. Thanks

image

Also, the zip file includes the 3.0.0 folder instead of the actual files (createui and maintemplate).

image

PR #13801, opened 2026-03-11
[ASIM] Authentcation parser - OktaSystemLogs

yummyblabla:
New parser for table OktaSystemLogs for Authentication schema


Parsers/ASimAuthentication/Parsers/vimAuthenticationOktaSystemLogs.yaml
```diff
@@ -0,0 +1,220 @@
...
+      disabled: bool=false
+  ) {
+    OktaSystemLogs
+    | where not(disabled)
+    | where EventOriginalType in (OktaSignInEvents)
+    | where 
+          (isnull(starttime) or TimeGenerated >= starttime) 
+          and (isnull(endtime) or TimeGenerated <= endtime)
+          and ((array_length(username_has_any) == 0) or ActorUsername has_any (username_has_any))
+          and (array_length(targetappname_has_any) == 0) // TargetAppName not available in source
```
bqiao-microsoft, line 88: If some field is not available in source, and we add a filter for it, we will return no records? Is that the ASIM convention?

yummyblabla: Yes, in the unifying parser, if the query contains filters, for let's say targetappname_has_any and passes it to this function, the function should not return any records as it doesn't apply to OktaSystemLogs.

PR #13893, opened 2026-03-24
[ASIM] Create WebSession parser for Cisco Umbrella Proxy Logs

yummyblabla:
Create ASIM parser for Cisco Umbrella proxy logs from azure function connector


Parsers/ASimWebSession/Parsers/ASimWebSessionCiscoUmbrella.yaml
```diff
@@ -0,0 +1,132 @@
...
+          FileSHA256 = ['SHA-SHA256_s'],
+          ThreatOriginalRiskLevel = AMP_Disposition_s
+      | extend
+          Verdict = column_ifexists("Verdict_s", ""),
+          Url = column_ifexists("URL_s", ""),
+          HttpRequestMethod = column_ifexists("Request_Method_s", ""),
+          RuleName = column_ifexists("Rule_ID_s", ""),
+          RuleNumber = toint(column_ifexists("Ruleset_ID_s", "")),
+          FileName = column_ifexists("File_Name_s", ""),
+          ThreatCategory = tostring(Blocked_Categories_s)
```
hunngu-ms, line 64: Why is column_ifexists used for some columns but not for others?

yummyblabla: For context, the azure function connector creates v1 tables, so the schema is not predefined at creation
https://github.com/Azure/Azure-Sentinel/blob/49ae03d9490e347da581ae5075d8e6c16ba0972c/Solutions/CiscoUmbrella/Data%20Connectors/ciscoUmbrellaDataConn/__init__.py#L403
 
Customers may be using different versions of their logging, which may not include all columns from Cisco

PR #13946, opened 2026-03-30
Fix Blacklens Logic App invalid secureData configuration

dxniro:
Change(s):
- Removed unsupported "outputs" from secureData.properties on the AutoDecode Compose action in the Logic App workflow
- Bumped solution version to 3.0.1

Reason for Change(s):
- Deployment fails with InvalidSecureDataConfiguration because Compose actions only support "inputs" in secureData.properties


v-shukore:
Hi @dxniro, could you please package using the V3 tool? It looks like the packaging wasn't done correctly, as the maintemplate contentversion is still showing 3.0.0. Thanks!

v-shukore:
Hi @dxniro, do not delete the old package from the solution, keep newly created package and uncommit the deleted package change from this PR. 
Also, implement the changes suggested by the copilot & repackage the solution.
Thanks!!

PR #13950, opened 2026-03-30
[ASIM] Create NetworkSession parser for Check Point Smart Defense

yummyblabla:
- Create NetworkSession parser for Check Point Smart Defense
- Align casing for _Im_NetworkSession_CorelightZeek in imNetworkSession


Parsers/ASimNetworkSession/CHANGELOG/ASimNetworkSession.md
hunngu-ms, line 1: Why skipped 0.7.2

yummyblabla: Looks like a typo. Changed it to 0.7.2

PR #14037, opened 2026-04-10
Fix for Imperva Cloud WAF CCF (S3‑Generic) Connector silently dropping logs which containing quotation characters

v-krishnachi:
The CCF CSV parser defaults to RFC 4180 behavior where double-quote characters are treated as field enclosures. 
Imperva WAF CEF logs contain embedded JSON in below fields which include quotation marks as regular data content. 
      1. cs10
      2. cs11
      3. additionalReqHeaders
      4. additionalResHeaders
This caused the parser to silently drop log lines containing these characters before they reached the DCR.

Fix:   Added HasCsvBoundary: false to the dataFormat configuration in both the PollingConfig and mainTemplate to disable RFC 4180 quote-boundary handling. 

Validated against customer sample data (5 log lines, 4 previously dropped). No regression to existing ingestion of non-quoted log lines.

Bumped solution version to 3.1.1 and created release package.

#   Change(s):
   - ImpervaCloudWAFLogs_PollingConfig
   - Maintemplate
   - REleasenotes
   - Package update

 
   Version updated:
   - Yes
   - Detections/Analytic Rule templates are required to have the version updated


v-krishnachi:
There is new PR given by Jason https://github.com/Azure/Azure-Sentinel/pull/14044 
So will be closing it

PR #14075, opened 2026-04-14
[ASIM] Fixes to VMwareVcenter Authentication parser

yummyblabla:
See Changelog for more details


Parsers/ASimAuthentication/Parsers/vimAuthenticationVMwareVCenter.yaml
```diff
@@ -127,7 +128,8 @@ ParserQuery: |
         | extend
             AdditionalFields = iff(pack, bag_pack("LoginTime", LoginTime), dynamic([])),
             IpAddr = SrcIpAddr,
-            Dvc = EventProduct
+            Dvc = EventProduct,
```
bqiao-microsoft: There was an issue with the 'Dvc' column too right?

bqiao-microsoft (dismissed):

Parsers/ASimAuthentication/Parsers/vimAuthenticationVMwareVCenter.yaml
yummyblabla: Addressed in following commit

bqiao-microsoft (approved):

PR #14150, opened 2026-04-28
Training Lab: E02 add note about empty TI join results

kapetanios55:
## Summary
Adds an explanatory note to Exercise 2 (Threat Intelligence) after the TI-to-telemetry join queries.

## Problem
The join queries in Step 4 match MDTI threat intelligence indicators against the lab's firewall and CloudTrail logs. Since the lab uses synthetic IP addresses, these joins will almost always return empty results — which can confuse users who expect matches.

## Fix
Added a callout note after both join queries explaining:
- Empty results are **expected** in the lab environment (synthetic IPs vs real-world TI)
- The **query pattern** (join kind=inner between TI and telemetry) is what matters
- In production with real traffic, these joins detect connections to known-malicious infrastructure

## Files Changed
- `Tools/Microsoft-Sentinel-Training-Lab/Exercises/E02_threat_intelligence_mdti.md`

### DIFF
--- a/Tools/Microsoft-Sentinel-Training-Lab/Exercises/E02_threat_intelligence_mdti.md
+++ b/Tools/Microsoft-Sentinel-Training-Lab/Exercises/E02_threat_intelligence_mdti.md
@@ -145,6 +145,10 @@ AWSCloudTrail
 | sort by TimeGenerated desc
 ```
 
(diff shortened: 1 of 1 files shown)


kapetanios55:
Closing — branch carried extra commits from a previous PR causing CI failures. Will recreate with a clean single-commit branch.

PR #14183, opened 2026-05-01
[ASIM] Apply fix to ProcessEvent M365D

yummyblabla:
Project TargetUserSessionId as it was missing before


Parsers/ASimProcessEvent/Parsers/vimProcessEventMicrosoft365D.yaml
```diff
@@ -1,7 +1,7 @@
 Parser:
   Title: Process Create Event ASIM parser for Microsoft 365 Defender for endpoint
-  Version: '0.4.0'
-  LastUpdated: Mar 10, 2026
+  Version: '0.4.10'
```
bqiao-microsoft: [nit] Should this be `0.4.1`?

bqiao-microsoft (dismissed):

Parsers/ASimProcessEvent/Parsers/vimProcessEventMicrosoft365D.yaml
yummyblabla: Yes.

bqiao-microsoft (approved):

PR #14197, opened 2026-05-04
RecordedFutureIdentity: fix invalid playbook

aommm:
Change(s):
   - remove stale references in `RFI-confirm-EntraID-risky-user` ARM template

   Reason for Change(s):
   - The action `Check_if_AD_Identity_Protection_risky_users_list_contains_the_user` was deleted in v1.2 but its `@body()` references were left in both response actions, causing `InvalidTemplate` validation failure at deploy time.

   Version Updated:
   - This playbook is NOT part of the main solution, it is only installable from Github. Therefore we have not re-packaged `mainTemplate.json` or bumped the solution version.

   Testing Completed:
   - See guidance below

   - See guidance below

### DIFF
--- a/Solutions/Recorded Future Identity/Playbooks/v3.0/RFI-confirm-EntraID-risky-user/azuredeploy.json
+++ b/Solutions/Recorded Future Identity/Playbooks/v3.0/RFI-confirm-EntraID-risky-user/azuredeploy.json
@@ -1,10 +1,10 @@
 {
     "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
-    "contentVersion": "1.2.0.0",
+    "contentVersion": "1.2.1.0",
     "metadata": {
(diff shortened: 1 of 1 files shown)


v-shukore:
Hi @aommm, please repackage this solution using V3 tool. Thanks!

aommm:
As mentioned in description, I don't think it needs to be packaged. Do you agree @v-shukore?

v-shukore (approved):

v-atulyadav (approved):

v-shukore:
Hi @aommm, since you updated the playbook and this is solution content without packaging, your latest changes won’t be reflected in the main template. You’ll need to repackage the solution. Thanks!

aommm:
@v-shukore I don't think that this playbook is solution content. It is not referenced from the Solution.json (or whatever it's called). I did run the packaging script but it only gave me version bumps and TZ changes, nothing relevant.

PR #14205, opened 2026-05-05
[Local validation] - add support for local validation

srikarshastry:
- add instructions for agents to run local validation of solution
- add scripts for each validation step
- updated copilot instructions

### DIFF
--- a/.github/instructions/packaging.instructions.md
+++ b/.github/instructions/packaging.instructions.md
@@ -16,9 +16,20 @@ Both steps are combined in a single script. **Always run validation after packag
 
 ### Before Running the Script
 
+**Always confirm the version bump with the user BEFORE running the script.** If the user did not explicitly specify `patch`, `minor`, or `major`, ask them:
+
+> What type of version bump should this be?
+> - **patch** — bug fixes, minor text changes, no new content
+> - **minor** — new content added (rules, connectors, playbooks, etc.) or significant updates
+> - **major** — breaking changes, major restructuring, or initial release
+>
+> (Default is `patch` if you're unsure.)
+
+Only proceed once you have a clear answer (or the user explicitly says to use the default). Use the confirmed value for the `-VersionBump` parameter.
+
 **Always tell the user what's about to happen BEFORE you run the script:**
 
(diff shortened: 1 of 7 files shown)


destinywu (dismissed):

.github/instructions/packaging.instructions.md
```diff
@@ -0,0 +1,296 @@
...
+1. **Package** the solution using the V3 packaging tool
+2. **Validate** the packaged solution using the local validation runner
+
+Both steps are combined in a single script. **Always run validation after packaging.**
+
+## How to Build a Solution
+
+### Before Running the Script
+
+**Always tell the user what's about to happen BEFORE you run the script:**
```
jlheard, line 30: Nice 👍!  Let's instruct the agent to verify the version with the user so it knows how to bump the version correctly and use the correct parameters. I foresee this as a possible issue

.script/local-validation/README.md
```diff
@@ -0,0 +1,224 @@
+# Local Validation Runner
+
+Run the same CI validation checks that GitHub Actions performs — **locally**, without GitHub API access, PR context, or secrets.
```
jlheard, line 3: Good idea 👍! Let's look into sharing logic between the actual pipeline and local validation; if possible. These may become out of sync.

If not, we may need another agent/workflow to keep the two in sync so changes in one reflect in the other.

jlheard (commented):
left some comments

.github/instructions/packaging.instructions.md
srikarshastry: updated

.script/local-validation/README.md
srikarshastry: so the current implementation is in sync almost 90%. Only the CLI/powershell scripts are little different. I updated the README.md to include this difference

jlheard (approved):
Looks good 👍

fuqing04 (approved):

v-maheshbh (approved):

v-atulyadav (approved):

v-dvedak (approved):

PR #14307, opened 2026-05-20
Add hunting queries: Entra ID identity boundary expansion hunting pack (3 queries)

descambiado:
## Summary

Three hunting queries targeting identity boundary expansion techniques in Entra ID. Each covers a distinct operation that shifts the permission surface without creating new accounts or directly adding credentials — making them harder to catch with account-focused detections.

### Queries added

**`GuestUserTypeChangedToMember.yaml`** — T1098  
Detects `Update user` events where `UserType` changes from `Guest` to `Member`. A guest-to-member conversion grants the account full tenant membership including access to internal resources, group memberships, and SharePoint sites that exclude guests. Rare in most tenants; correlate against help desk records if it fires.

**`ServicePrincipalOwnerAdded.yaml`** — T1098.001  
Detects `Add owner to service principal` events. SP ownership silently grants full credential management rights — an actor who becomes SP owner can then add a password or certificate credential and authenticate as the SP without triggering the separate `Add service principal credentials` operation immediately. This is the documented precursor step in Midnight Blizzard-style persistence chains, and it sits in a detection gap between general role auditing and credential-focused detections.

**`ApplicationRedirectUriModified.yaml`** — T1528  
Detects `Update application` events where the `ReplyUrls` field changes. Adding an attacker-controlled redirect URI to an existing trusted app registration allows interception of OAuth authorization codes from users who click legitimate consent flows pointing to that app. No new app registration is required, so first-seen-app detections miss it entirely.

### What was verified

- No existing queries in `Hunting Queries/` cover these three operations
- All descriptions ≤ 255 characters, start with "Identifies", sentence-case names
- `OperationName =~` (exact match with `in~`/`=~`, no `has_any`)
- Direct `InitiatedBy.user.userPrincipalName` field access throughout, no `parse_json(tostring(...))` wrapper
- No non-ASCII characters
- Three separate commits, one per query


v-maheshbh:
Hi @descambiado 

Kindly review the above comments and address them as needed.

Thanks!

descambiado:
Pushed updates addressing the Copilot review on this hunting pack:

- `ApplicationRedirectUriModified`: the Account entity `FullName` now maps to `ActorUpn` only, not the merged `Actor` variable (which could be an application display name when `ActorUpn` is empty). Removed the dead `Actor` extend and changed the `AccountName` fallback to `""` so the Account entity is only created for user-initiated events. `ActorApp` is still projected as a separate column for context.
- `ServicePrincipalOwnerAdded`: replaced the positional `TargetResources[0]` (SP) and `TargetResources[1]` (owner) indexing with an `mv-apply` block that uses `take_anyif` filtered by `type =~ "ServicePrincipal"` and `type =~ "User"`, so the SP and the new owner are selected by resource type rather than array position. Normalised the `metadata` block to 2-space indentation to match the rest of the file.

Merged latest master.

v-maheshbh (approved):

v-atulyadav (approved):

PR #14366, opened 2026-05-28
Vaikora-AzureSecurityCenter v3.0.1 — align playbook + analytic rules with Vaikora /actions API schema

mazamizo21:
## Summary

End-to-end testing surfaced that the 3.0.0 solution was built against an assumed alert-shaped API response while the Vaikora \`/api/v1/actions\` endpoint actually returns action-shaped events. This caused failures across the playbook and all three analytic rules. This PR aligns the entire solution on the action-field schema and the \`Vaikora_AgentSignals_CL\` table.

## Defects in 3.0.0

1. **\`Poll_Vaikora_Actions\`** always sent \`agent_id=\` even when \`VaikoraAgentId\` was blank. Vaikora API returned HTTP 422 because Pydantic cannot parse \`""\` as a UUID. Every downstream action skipped.

2. **\`Send_to_Log_Analytics\`** crashed with \`InvalidTemplate\` because \`replace()\` ran against null \`title\` / \`description\` fields that do not exist on the response.

3. **\`Filter_High_Risk_Actions\`** referenced \`threat_detected\`, which is missing from the response.

4. **Three analytic rules** queried \`Vaikora_SecurityAlerts_CL\` with columns (\`AlertId_s\`, \`Title_s\`, \`Description_s\`, \`SourceIP_s\`, \`UserName_s\`, etc.) that the playbook never wrote.

5. **Same solution shipped two divergent target tables**: standalone playbook wrote to \`Vaikora_SecurityAlerts\`; mainTemplate inner playbook wrote to \`Vaikora_AgentSignals\`. Rules failed to match in both install paths.

## Fix

Align everything on the action-field schema and \`Vaikora_AgentSignals_CL\`:

- **\`Poll_Vaikora_Actions\`** URI rebuilt with \`if(empty(parameters('VaikoraAgentId')))\` so the \`agent_id\` segment is omitted when blank. Switched from \`queries\` object to URI concat so the conditional works.
- **\`Send_to_Log_Analytics\`** body rewritten to action fields (\`action_id\`, \`agent_id\`, \`action_type\`, \`resource_type\`, \`resource_id\`, \`status\`, \`severity\`, \`policy_id\`, \`policy_decision\`, \`is_anomaly\`, \`anomaly_score\`, \`anomaly_reason\`, \`log_hash\`, \`timestamp\`) with \`coalesce()\` guards on nullable values. \`Log-Type\` set to \`Vaikora_AgentSignals\`.
- **\`is_anomaly\`** wrapped with \`toLower(string(coalesce(...)))\` so it serializes as JSON \`true\`/\`false\` instead of Logic Apps' default \`True\`/\`False\`, which \`json()\` refuses to parse.
- **\`Parse_Response\`** schema updated to declare the real \`actions\` array shape with nullable types on every property (the API returns explicit \`null\` for some optional fields).
- **\`Filter_High_Risk_Actions\`** drops the \`threat_detected\` check (standalone playbook + mainTemplate inner playbook).
- **Three analytic rules** rewritten to query \`Vaikora_AgentSignals_CL\` with action-field columns.
- **\`entityMappings\`** reduced to \`AgentId\` (from \`agent_id_s\`) mapped to \`Account.Name\`. The response carries no IP/Host/Process fields.
- **\`alertDetailsOverride\`** placeholders updated and kept at or below the 3-placeholder cap.
- **Solution description** updated to reference \`Vaikora_AgentSignals_CL\`.

## Files changed

- \`Solutions/Vaikora-AzureSecurityCenter/Playbooks/VaikoraToAzureSecurityCenter/azuredeploy.json\`
- \`Solutions/Vaikora-AzureSecurityCenter/Package/mainTemplate.json\` — \`_solutionVersion\` bumped to 3.0.1
- \`Solutions/Vaikora-AzureSecurityCenter/Package/3.0.1.zip\` (3.0.0.zip retained per repo policy)
- \`Solutions/Vaikora-AzureSecurityCenter/Analytic Rules/Vaikora - High Severity Security Alerts.yaml\`
- \`Solutions/Vaikora-AzureSecurityCenter/Analytic Rules/Vaikora - Anomaly Detection.yaml\`
- \`Solutions/Vaikora-AzureSecurityCenter/Analytic Rules/Vaikora - Feed Outage Detection.yaml\`
- \`Solutions/Vaikora-AzureSecurityCenter/Data/Solution_VaikoraSecurityCenter.json\` — Version 3.0.1
- \`Solutions/Vaikora-AzureSecurityCenter/ReleaseNotes.md\` — v3.0.1 entry
- \`Solutions/Vaikora-AzureSecurityCenter/SolutionMetadata.json\` — lastPublishDate 2026-05-28

## Test plan

- [x] arm-ttk local: 49/49 pass on \`Package/3.0.1.zip\`
- [x] Verified Vaikora API response shape against \`api.vaikora.com/api/v1/actions\` (live response: \`{actions: [{action_type, agent_id, severity, policy_decision, is_anomaly, anomaly_score, anomaly_reason, log_hash, timestamp, …}], total, page, per_page}\`)
- [x] **Full E2E on a real Sentinel workspace**: Poll_Vaikora_Actions → Parse_Response → Filter_High_Risk_Actions → For_Each_Security_Alert → Send_to_Log_Analytics all return \`Succeeded\`. The JSON body sent to LA contains every action field including \`anomaly_reason\` populated with real explanation text. LA Data Collector API returned HTTP 200.
- [ ] CI \`run-arm-ttk\` green
- [ ] Reviewer Content Hub install + standalone playbook test

cc @v-maheshbh @v-shukore


v-shukore:
Hi @mazamizo21, please resolve playbook validation and arm-ttk is failing for below reason please check. Thanks!
image

mazamizo21:
Hi @v-maheshbh, the PlaybooksValidations failure here was the standalone playbook's Logic App location: it was `[parameters('location')]`, but `playbooksValidator.js` requires `[resourceGroup().location]` on the `Microsoft.Logic/workflows` resource. Fixed in `e68e3c8f70`: changed only the workflow resource location. The `location` parameter stays, since the `Microsoft.Web/connections` resource and the managedApi id paths still reference it.

Could you approve/re-run the validation workflow so the check goes green? Thanks!

v-shukore:
Hi @mazamizo21, thanks for sharing the update I'll review the PR and provide update. Thanks!

v-shukore:
Hi @mazamizo21, still arm-ttk is failing for same concat function please update it and commit. Thanks!
image

mazamizo21:
Hi @v-shukore, fixed the arm-ttk `concat`-in-uri failure in commit `a17038b26e`.

Root cause: the agent_id-omit fix (omit the `agent_id` query param when blank, otherwise the Vaikora API returns 422) put a conditional `concat()` inside the HTTP poll action's `uri`. arm-ttk's `URIs Should Be Properly Constructed` rejects `concat`/`format` anywhere in a `uri`/`url` property, so your local run flagged it even though the bundled CI arm-ttk passed.

Fix: moved the URL construction into a `Compose` action and the HTTP action now reads `"uri": "@{outputs('Compose_Poll_Endpoint')}"` — no `concat` in any uri. Runtime behavior is identical (empty `VaikoraAgentId` still drops the `agent_id=` segment). Applied to both the package `mainTemplate` and the standalone `azuredeploy.json`, and repackaged `3.0.1.zip`.

Verified locally with upstream arm-ttk: `URIs Should Be Properly Constructed` now PASSES (49 tests, the only remaining item is `IDs Should Be Derived From ResourceIDs` on the `azureloganalyticsdatacollector` managed-API connection id — that path can't use `resourceId()` and is the same pattern shipping in merged connection-based solutions, which is why Sentinel CI's arm-ttk subset doesn't run it). Could you re-run the validation workflow? Thanks!

v-shukore (approved):

PR #14406, opened 2026-06-03
SentinelOne: add multi-instance CCP connector support and package as …

Bhagya767:
This PR updates the SentinelOne solution to support multiple CCP connector instances for MSSP scenarios and packages the change as version 3.0.10.

Changes included:

Updated the SentinelOne CCP connector experience from the single-instance toggle flow to a multi-instance flow
Added Data Connectors Grid and Context Pane based configuration for adding and managing multiple SentinelOne instances
Preserved the existing SentinelOne release history and added the new change under version 3.0.10
Fixed the workbook metadata JSON issue that was blocking correct packaging output
Regenerated the SentinelOne package artifacts for version 3.0.10


v-shukore:
Hi @Bhagya767, I checked and noted outside and inside zip file maintemplates are different please check it once and make sure it should be same. Thanks!

jlheard (approved):

PR #14412, opened 2026-06-04
added missing tostring() call to gravityzone asim parser

gbarbieru:
Change(s):
   - added missing tostring() call to gravityzone asim parser

   Reason for Change(s):
   - asim function failing in log analytics workspace query

   Version Updated:
   - No

   Testing Completed:
   - Yes, via Azure Portal -> Deploy from custom template
image

   - Yes


Parsers/ASimAlertEvent/Parsers/ASimAlertEventBitdefenderGravityZone.yaml
```diff
@@ -146,7 +146,7 @@ ParserQuery: |
             MaxRemediationAction == 2, "Medium", // move
             "High" // fallback - disinfect or delete
         ),
-        DvcHostname = d.computerName,
+        DvcHostname = tostring(d.computerName),
```
yummyblabla, line 149: Also update the vim parser as well. I see some misalignment in some columns that are not calling tostring()

v-atulyadav:
Hi @gbarbieru,
Please review the suggested changes above and take the necessary action accordingly. Thanks

Parsers/ASimAlertEvent/Parsers/ASimAlertEventBitdefenderGravityZone.yaml
gbarbieru: you mean to move the tostring() calls from `project` to `extend` region ?
currently, the outcome is as desired:
image

yummyblabla: Oh I see. You cast it to string in the project.

yummyblabla (approved):

v-atulyadav (approved):

PR #14215, opened 2026-05-06
[ASIM] Authentication - Salesforce Service Cloud - Codeless Connector Tables Support

SpeedyFireCyclone:
Required items, please complete
   
   Change(s):
   - Add support for codeless connector via SalesforceServiceCloudV2_CL table
   - Update unifying parser with `pack` parameter for the updated parser
   - Add support for updated connector via SalesforceServiceCloudV3_CL table

   Reason for Change(s):
   - Support parsing logs from the latest codeless connector

   Version Updated:
   - Yes

   Testing Completed:
   - Yes

   Checked that the validations are passing and have addressed any issues that are present:
   - Yes

PR #14180, opened 2026-05-01
[SalesforceServiceCloud] - v3.2.0 to include all events from ELF and support for multiple domains

srikarshastry:
v3.2.0 now supports following options for the CCF data connector:
- new `SalesforceServiceCloudV3_CL` table that include all fields from all events within the event log files
- option to add multiple domains within same workspace/data connector
- updated parser to include the new `SalesforceServiceCloudV3_CL` table

New Connector behavior:
image

New Parser:
image

PR #14144, opened 2026-04-27
[SalesForceEventLogConnector] - add support for missing events and username-password oauth2

srikarshastry:
1. add support for all events within ELF objects
2. add support for Oauth2 username-password flow
3. change UI to support multiple stream (future ready)

PR #14423, opened 2026-06-05
[ASIM] Create authentication parser for SalesforceServiceCloudV2_CL

yummyblabla:
Create authentication parser for SalesforceServiceCloudV2_CL
See CHANGELOG for details


hunngu-ms (approved):

Parsers/ASimAuthentication/Parsers/ASimAuthenticationSalesforceServiceCloudV2.yaml
```diff
@@ -0,0 +1,218 @@
...
+      "Login",  "Logon",
+      "LoginAs", "Logon",
+      "Logout", "Logoff"
+  ];
+  let parser = (
+      disabled: bool = false,
+      pack: bool = false
+  )
+  {
+      SalesforceServiceCloudV2_CL
```
SpeedyFireCyclone, line 61: The latest version of the Salesforce Service Cloud connector refers to the [`SalesforceServiceCloudV3_CL`](https://github.com/Azure/Azure-Sentinel/pull/14180) table, after several fields that were available in the original connector were not ported over to the CCP-based one.

```suggestion
      union isfuzzy=true
          SalesforceServiceCloudV2_CL,
          SalesforceServiceCloudV3_CL
```

Parsers/ASimAuthentication/Parsers/ASimAuthenticationSalesforceServiceCloudV2.yaml
```diff
@@ -0,0 +1,218 @@
...
+      | extend
+          EventResult = iff(isnotempty(EventResult_lookup), EventResult_lookup, iff(EventType == "Logoff", "Success", "Failure")),
+          EventResultDetails = iff(isnotempty(EventResultDetails_lookup), EventResultDetails_lookup, "")
+      | project-away EventResult_lookup, EventResultDetails_lookup
+      // -- Map UserType
+      | lookup SalesforceUserType on UserType
+      // -- Build ASIM fields
+      | extend
+          // -- Event fields (Mandatory)
+          Type = "SalesforceServiceCloudV2_CL",
```
SpeedyFireCyclone, line 79: This field is populated by default by the underlying table.
```suggestion
```

Parsers/ASimAuthentication/Parsers/vimAuthenticationSalesforceServiceCloudV2.yaml
```diff
@@ -0,0 +1,299 @@
...
+      srcipaddr_has_any_prefix: dynamic = dynamic([]),
+      srchostname_has_any: dynamic = dynamic([]),
+      eventtype_in: dynamic = dynamic([]),
+      eventresultdetails_in: dynamic = dynamic([]),
+      eventresult: string = '*',
+      disabled: bool = false,
+      pack: bool = false
+  )
+  {
+      SalesforceServiceCloudV2_CL
```
SpeedyFireCyclone, line 97: Same as noted the non-filtering parser;

The latest version of the Salesforce Service Cloud connector refers to the [`SalesforceServiceCloudV3_CL`](https://github.com/Azure/Azure-Sentinel/pull/14180) table, after several fields that were available in the original connector were not ported over to the CCP-based one.

```suggestion
      union isfuzzy=true
          SalesforceServiceCloudV2_CL,
          SalesforceServiceCloudV3_CL
```

Parsers/ASimAuthentication/Parsers/vimAuthenticationSalesforceServiceCloudV2.yaml
```diff
@@ -0,0 +1,299 @@
...
+      // -- Event result details filter
+      | where (array_length(eventresultdetails_in) == 0) or (EventResultDetails in~ (eventresultdetails_in))
+      // -- Event result filter
+      | where (eventresult == "*" or (EventResult == eventresult))
+      // -- Map UserType
+      | lookup SalesforceUserType on UserType
+      // -- Build ASIM fields
+      | extend
+          // -- Event fields (Mandatory)
+          Type = "SalesforceServiceCloudV2_CL",
```
SpeedyFireCyclone, line 137: Same as noted the non-filtering parser;

This field is populated by default by the underlying table.

```suggestion
```

SpeedyFireCyclone (commented):
Thanks for the work @yummyblabla & @hunngu-ms, I've dropped my pull request in favor of this one. I do have some suggestions regarding supporting of the V2 table and not the V3.

The V2 table was introduced when the Data Connector moved to CCP. Some fields that were available in the legacy connector it replaces were not initially included, and were added to the V2 schema later: [PR 13150](https://github.com/Azure/Azure-Sentinel/pull/13150/changes#diff-53c4465abc101e93c2ddb303cb4aed9af4c0fc626c6fa21f28ff3ae99d2e79bb), [PR 14144](https://github.com/Azure/Azure-Sentinel/pull/14144/changes#diff-53c4465abc101e93c2ddb303cb4aed9af4c0fc626c6fa21f28ff3ae99d2e79bb). Because the table was updated and did not receive a new version number some fields may be missing from certain environments. (I checked and there is no use of those fields in this parser). A later PR (#14180) fixed this and bumped up the table name to V3, which should become the more commonly used table as the connector gets updated across Sentinel instances.

yummyblabla:
Hi @SpeedyFireCyclone 
Yes, there is the V3 of the Salesforce ServiceCloud connector. This is targeting the V2 connector.

I see your PR and its aim is to union the 3 tables, original, V2, and V3.  I think I would prefer your solution better given the history of it. I wanted to separate V2 and V3 parsers because of the schema differences, but since V2 has different variations where some environments may not have certain columns, I don't think we can get away from using column_ifexists, which is what I wanted to stay away from originally by separating V2 and V3.

We can go ahead with your PR instead as I would agree with your solution.

SpeedyFireCyclone:
> Hi @SpeedyFireCyclone Yes, there is the V3 of the Salesforce ServiceCloud connector. This is targeting the V2 connector.
> 
> I see your PR and its aim is to union the 3 tables, original, V2, and V3. I think I would prefer your solution better given the history of it. I wanted to separate V2 and V3 parsers because of the schema differences, but since V2 has different variations where some environments may not have certain columns, I don't think we can get away from using column_ifexists, which is what I wanted to stay away from originally by separating V2 and V3.
> 
> We can go ahead with your PR instead as I would agree with your solution.

@yummyblabla, Sorry I completely missed your comment, I've reopened my pull request: #14215

PR #14450, opened 2026-06-09
[ASIM] Add File and Parser Validation Workflow

yummyblabla:
Do not merge until Azure organization has access to GitHub Models (not free tier)


oshezaf (approved):

yummyblabla:
Models is not aactivated in the Azure organization, so adding LLM analysis in a workflow is not a feasible option.

PR #14475, opened 2026-06-13
QualysVM 3.0.9 - fix Defender portal CCF data connector page KQL parse error

v-krishnachi:
## Summary

Fixes a client-side KQL parse error (`Query could not be parsed at '{' on line [5,18]`) shown in the Defender portal (`security.microsoft.com`) when opening the **Qualys Vulnerability Management (via Codeless Connector Framework)** data connector page.

## Root cause

When a `dataConnectorDefinition` (`kind: Customizable`) is created with `apiVersion: 2024-01-01-preview` or later, the `graphQueriesTableName` field in `connectorUiConfig` is dropped during storage by the internal model. The `{{graphQueriesTableName}}` placeholders inside `graphQueries`, `sampleQueries`, and `dataTypes` are therefore never substituted and reach the portal as literal text — which the KQL engine then fails to parse.

Definitions written before the API change still resolve correctly because they were stored under an older internal model that retained the field. This is a known issue confirmed by the unified connectors team.

## Fix proposed by Unified Connectors Team

Removed the `graphQueriesTableName` property entirely and inlined the literal table name `QualysHostDetectionV3_CL` everywhere `{{graphQueriesTableName}}` was used.

## Files changed

| `Solutions/QualysVM/Data Connectors/QualysVMHostLogs_ccp/QualysVMHostLogs_ConnectorDefinition.json` 
| `Solutions/QualysVM/Package/mainTemplate.json`
| `Solutions/QualysVM/Package/3.0.9.zip`
| `Solutions/QualysVM/data/Solution_QualysVM.json`
| `Solutions/QualysVM/SolutionMetadata.json
| `Solutions/QualysVM/ReleaseNotes.md`

## Local validation
Screenshot 2026-06-12 160033

## Customer follow-up steps (already shared with the connectors team)

1. Reinstall the **QualysVM** solution from Content Hub once 3.0.9 is published — re-installation rewrites the `dataConnectorDefinitions` resource so the corrected definition is stored.
2. If a connector instance was already configured, delete and re-create it so the new definition is used.


v-shukore:
Hi @v-krishnachi, please resolve branch conflicts to proceed. Thanks!

PR #14479, opened 2026-06-15
Recorded Future Identity: Migrate from Data Collector API

aommm:
Change(s):
   - Add DCE/DCR/connector tile for importing Recorded Future Alerts
   - Migrate `RFI-Playbook-Alert-Importer-LAW` logic app to import data via DCE/DCR 

   Reason for Change(s):
   - Microsoft deprecating the Data Collector API

   Version Updated:
   - Yes, updated Analytic Rule version number

-------------------

In this PR we've only migrated one of the logic apps of **Recorded Future Identity**.

We would be glad to hear your feedback on our approach.

We're planning to submit further PRs later for the rest of the logic apps, as well as for our main solution (which will be a lot bigger).


aommm:
Sorry, had to re-create this PR to get the changes to show properly:
https://github.com/Azure/Azure-Sentinel/pull/14495

PR #14569, opened 2026-06-26
[ASIM] Cisco FTD NetworkSession Parser

Steve1145:
Change(s):

Added NetworkSession ASIM parser for Cisco FTD logs ingested via the 'Cisco ASA/FTD via AMA' Data Connector

Reason for Change(s):

Introduce ASIM NetworkSession support for Cisco FTD to complement existing Cisco Firepower logs support.

Version Updated:

Yes

Testing Completed:

Yes

Yes


Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoFTD.yaml
```diff
@@ -0,0 +1,497 @@
...
+                      'FileStorageStatus', FileStorageStatus,
+                      'FileSandboxStatus', FileSandboxStatus
+                  ),
+                  dynamic({})
+              )
+      ;
+      union Intrusion, Connections, File
+      | extend IPParsedHostname = (ipv6_is_match(DeviceName, Computer))
+      | extend
+          Dvc = iff(IPParsedHostname == false, coalesce(DeviceName, Computer), ''),
```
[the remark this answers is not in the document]
Steve1145: Removed ad DvcId always present - so this extension was unnecessary.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoFTD.yaml
```diff
@@ -0,0 +1,416 @@
...
+                      'FileStorageStatus', FileStorageStatus,
+                      'FileSandboxStatus', FileSandboxStatus
+                  ),
+                  dynamic({})
+              )
+      ;
+      union Intrusion, Connections, File
+      | extend IPParsedHostname = (ipv6_is_match(DeviceName, Computer))
+      | extend
+          Dvc = iff(IPParsedHostname == false, coalesce(DeviceName, Computer), ''),
```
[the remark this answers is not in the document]
Steve1145: Removed ad DvcId always present - so this extension was unnecessary.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoFTD.yaml
```diff
@@ -0,0 +1,416 @@
...
+          Dst = DstIpAddr,
+          SrcUsernameType = case(
+              SrcUsername has '@', 'UPN',
+              SrcUsername has @'\', 'Windows',
+              SrcUsername =~ 'Not Found', '',
+              isnotempty(SrcUsername), 'Simple',
+              ''
+          ),
+          Rule = tostring(NetworkRuleNumber),
+          ASimMatchingIpAddr = 'Both' //set using has_any_ipv4_prefix in parameter based parser
```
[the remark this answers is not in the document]
Steve1145: Removed - ASimMatchingIpAddr only used in the filtering parser.

v-atulyadav:
Hi @Steve1145,
Please check the KQL failure and take the necessary action. Thanks
image

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoFTD.yaml
```diff
@@ -0,0 +1,415 @@
...
+          Duration = NetworkDuration,
+          Dst = DstIpAddr,
+          SrcUsernameType = case(
+              SrcUsername has '@', 'UPN',
+              SrcUsername has @'\', 'Windows',
+              SrcUsername =~ 'Not Found', '',
+              isnotempty(SrcUsername), 'Simple',
+              ''
+          ),
+          Rule = tostring(NetworkRuleNumber)
```
yummyblabla: I see Rule = ACPolicy earlier , and then overwritten here Rule = tostring(NetworkRuleNumber).
Is the intention Rule = coalesce(ACPolicy, tostring(NetworkRuleNumber)) ?

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoFTD.yaml
```diff
@@ -0,0 +1,415 @@
...
+          DstNatIpAddr = column_ifexists('DstNatIpAddr', ''),
+          DstNatPortNumber = column_ifexists('DstNatPortNumber', int(null)),
+          ThreatConfidence = column_ifexists('ThreatConfidence', int(null))
+      | extend
+          EventCount = int(1),
+          EventSchema = 'NetworkSession',
+          EventSchemaVersion = '0.2.7',
+          EventProduct = 'FTD',
+          EventVendor = 'Cisco',
+          DvcInterface = DvcInboundInterface,
```
yummyblabla, line 399: Same thing for DvcInterface. It is previously InstanceID and then replaced by DvcInboundInterface

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoFTD.yaml
```diff
@@ -0,0 +1,495 @@
...
+              IngressZone: string,
+              EgressZone: string,
+              NAPPolicy: string
+          ) with (pair_delimiter=',', kv_delimiter=':')
+          | where (
+              ((array_length(srcipaddr_has_any_prefix) == 0) or (has_any_ipv4_prefix(SrcIP, srcipaddr_has_any_prefix)))
+              and ((array_length(dstipaddr_has_any_prefix) == 0) or (has_any_ipv4_prefix(DstIP, dstipaddr_has_any_prefix)))
+              and ((array_length(ipaddr_has_any_prefix) == 0) or (has_any_ipv4_prefix(DstIP, ipaddr_has_any_prefix)) or (has_any_ipv4_prefix(SrcIP, ipaddr_has_any_prefix)))
+              and (isnull(dstportnumber) or DstPort == dstportnumber)
+              and ((array_length(hostname_has_any) == 0) or (Hostname has_any (hostname_has_any)))
```
yummyblabla: Hostname filter can be applied before parse-kv

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoFTD.yaml
```diff
@@ -0,0 +1,495 @@
...
+          ThreatCategory,
+          EventResultDetails,
+          EventType,
+          EventResult,
+          DvcAction,
+          EventSeverity,
+          AdditionalFields,
+          Hostname,
+          //Columns not available in all branches
+          NetworkRuleName = column_ifexists('NetworkRuleName', ''),
```
yummyblabla: Do we need these columns that use column_ifexists? These columns don't exist in CommonSecurityLog

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoFTD.yaml
Steve1145: Thanks Derrick.  I've corrected this by moving ACPolicy to AdditionalFields and following the documentation to ensure Rule is either the tostring(NetworkRuleNumber) or NetworkRuleName.  All IDs have Name but Intrusion have the Number which would be the preference.

Parsers/ASimNetworkSession/Parsers/ASimNetworkSessionCiscoFTD.yaml
Steve1145: Similar to the above.  Moved InstanceID to AdditionalFields and followed the documentation to alias DvcInterface with DvcInboundInterface. Thank you.

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoFTD.yaml
Steve1145: I propose that these fields are required as they are available in at least one of the branches (Intrusion, Connection or File) and map to a NetworkSession field, but not available in all branches and I wanted to avoid isfuzzy=true.

I could extend these fields in all branches as null values where the column does not exist in that branch i.e. DstBytes = long(null) in Intrusion etc - if that would be the preference instead of using column_ifexists for those fields in the union project?

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoFTD.yaml
Steve1145: Thank you - moved higher and immediately after the core Vendor / Product and DeviceEventClassID filtering.

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoFTD.yaml
yummyblabla: I see what you what you mean.
I think you don't need to apply the function column_ifexists. When you are unioning the 3 scenarios, the union combines the schemas from those 3 tables, and the output should contain all columns from those 3 tables regardless if they have rows or not. You can simply project the columns as is without calling the ifexists function.

yummyblabla (commented):
One unresolved issue, but otherwise, LGTM

Parsers/ASimNetworkSession/Parsers/vimNetworkSessionCiscoFTD.yaml
yummyblabla: This is from me trying it out. If your testing shows otherwise, let me know.

You are right that we don't need isfuzzy=true

Steve1145: Thanks Derrick, successfully tested and updated.  

This worked for all columns other than ThreatConfidence - which I had to declare null values in the two scenarios the column wasn't present.  ThreatConfidence = int(null) was added to the Intrusion and File branches in order to project in the union.

yummyblabla (approved):

PR #14599, opened 2026-07-01
Update CEF AMA stream name

didiergbenou-ms:
## Proposed change
Update the CEF stream constant used by `DataConnectors/Syslog/Sentinel_AMA_troubleshoot.py` from the legacy `SECURITY_CEF_BLOB` value to `Microsoft-CommonSecurityLog`.

## Why
The current Microsoft Sentinel CEF via AMA documentation says the DCR `streams` field should be set to `Microsoft-CommonSecurityLog` for CEF messages:
https://learn.microsoft.com/en-us/azure/sentinel/connect-cef-syslog-ama?tabs=api

## Validation
- `python -m py_compile DataConnectors/Syslog/Sentinel_AMA_troubleshoot.py`


didiergbenou-ms:
Closing this PR because the proposed change is no longer needed.

PR #14606, opened 2026-07-01
Add Halcyon ASIM DNS parser

jwilke-halcyon:
Change(s):
- Added ASIM DNS parser for Halcyon (ASimDnsHalcyon / vimDnsHalcyon)
- Registered the Halcyon sub-parser in the DNS union parsers (ASimDns / imDns)

Reason for Change(s):
- Normalize Halcyon OCSF DNS events (HalcyonEventsV2_CL) to the ASIM Dns schema so they work with schema-agnostic Sentinel content out of the box

Version updated:
- N/A

Testing Completed:
- Yes

- Yes


v-atulyadav:
Hi @jwilke-halcyon,
Please review the suggestions above and address them accordingly. Also, check and resolve the failed sample data validation.

v-atulyadav:
Hi @jwilke-halcyon,
Please review the suggestions above and address them accordingly. Also, check the failed sample data validation and resolve any issues. Thanks

v-atulyadav:
Hi @jwilke-halcyon,
Please review the suggestions above and address them accordingly. Also, please check the failed sample data validation and resolve any issues. Thanks

jwilke-halcyon:
@v-atulyadav - I think I've addressed everything now!

Parsers/ASimDns/Parsers/ASimDnsHalcyon.yaml
```diff
@@ -0,0 +1,115 @@
...
+          EventProduct    = "Halcyon",
+          EventVendor     = "Halcyon",
+          EventSchema     = "Dns",
+          EventSchemaVersion = "0.1.7",
+          AdditionalFields = bag_pack(
+                                "halcyon_dxp_rule",  tostring(unmapped.halcyon_dxp_rule),
+                                "halcyon_alert_id",  tostring(unmapped.halcyon_alert_id),
+                                "halcyon_tenant_id", tostring(unmapped.halcyon_tenant_id)
+                            )
+      | project-away DvcHostname_raw
```
yummyblabla: Please project only ASIM columns in the final output. We want to avoid using project-away as it does not future-proof additional columns added to the original source table.

Parsers/ASimDns/Parsers/vimDnsHalcyon.yaml
```diff
@@ -0,0 +1,156 @@
...
+      [
+          "IN", 1, "CH", 3, "HS", 4, "NONE", 254, "ANY", 255
+      ];
+      HalcyonEventsV2_CL
+      | where not(disabled)
+          and class_uid == 4003                                    // OCSF DnsActivity (constant for class 4003)
+          and (isnull(starttime) or coalesce(HalcyonSourceTime, TimeGenerated) >= starttime)
+          and (isnull(endtime)   or coalesce(HalcyonSourceTime, TimeGenerated) <= endtime)
+      // EventType is driven by activity_id (Halcyon emits activity_id == 1 / "Query" constantly).
+      | extend EventType = case(activity_id == 1, "Query", activity_id == 2, "Response", activity_id == 6, "Traffic", "Query")
```
yummyblabla: Traffic is not a valid enumeration for EventType.
We only support:
Query|Status|Notify|Update|DNS Stateful Operations

Parsers/ASimDns/Parsers/vimDnsHalcyon.yaml
```diff
@@ -0,0 +1,156 @@
...
+      // eventtype filter compares against the computed EventType.
+      // 'lookup' is a legacy DNS eventtype alias accepted as match-all for compatibility; the imDns union does not pass it by default.
+      | where (eventtype == '*' or eventtype =~ 'lookup' or EventType =~ eventtype)
+      // Pushdown filters on raw fields.
+      // NOTE: srcipaddr is not applicable - source events carry hostname/uid only, no IP.
```
yummyblabla: Although not applicable, add | where array_length(srcipaddr) == 0 so that if someone intends to filter by ipaddr, they will not get data from this parser.

Parsers/ASimDns/Parsers/ASimDnsHalcyon.yaml
jwilke-halcyon: I've addressed this with an explicit set of projected columns 👍

Parsers/ASimDns/Parsers/vimDnsHalcyon.yaml
jwilke-halcyon: Switched to Query! Thanks!

Parsers/ASimDns/Parsers/vimDnsHalcyon.yaml
jwilke-halcyon: Ah thanks, I addressed this with an equivalent string comparison check. 👍

yummyblabla (commented):
I see the test failing KQLValidations because it is missing the built-in column "Type" in the predefined table.

If you can add the column to https://github.com/Azure/Azure-Sentinel/blob/master/.script/tests/KqlvalidationsTests/CustomTables/HalcyonEventsV2_CL.json, that would resolve the issue.

Otherwise, LGTM, @jwilke-halcyon

jwilke-halcyon:
@yummyblabla - I've added the Type column.

jwilke-halcyon:
@v-atulyadav / @yummyblabla - [Run ASim Template Validation tests](https://github.com/Azure/Azure-Sentinel/actions/runs/29898152878/job/88852713036?pr=14606#logs) Any ideas why this is failing?

jwilke-halcyon:
@v-atulyadav - it seems like the CI ASIM template step is failing due to a workflow issue?

v-atulyadav:
> @v-atulyadav - it seems like the CI ASIM template step is failing due to a workflow issue?

We are looking into this

jwilke-halcyon:
> We are looking into this

@v-atulyadav - thank you! Are you able to determine if there's an action we need to make?

jwilke-halcyon:
@v-atulyadav @yummyblabla - Any update on this failure? Can we re-run the test?

v-atulyadav:
Hi @jwilke-halcyon,
There is no action required from your side at the moment. We are currently investigating the failed validation and will keep you updated as soon as possible. Thanks

v-atulyadav:
Hi @jwilke-halcyon,
The validation issue has now been resolved. Thanks

Parsers/ASimDns/Parsers/vimDnsHalcyon.yaml
```diff
@@ -0,0 +1,210 @@
...
+  let parser = (starttime:datetime=datetime(null), endtime:datetime=datetime(null), srcipaddr:string='*', domain_has_any:dynamic=dynamic([]), responsecodename:string='*', response_has_ipv4:string='*', response_has_any_prefix:dynamic=dynamic([]), eventtype:string='Query', disabled:bool=false) {
+      // DNS query class name -> numeric class ID lookup (standard DNS CLASS values).
+      let DnsClassLookup = datatable(DnsQueryClassName:string, DnsQueryClass:int)
+      [
+          "IN", 1, "CH", 3, "HS", 4, "NONE", 254, "ANY", 255
+      ];
+      HalcyonEventsV2_CL
+      | where not(disabled)
+          and class_uid == 4003                                    // OCSF DnsActivity (constant for class 4003)
+          and (isnull(starttime) or coalesce(HalcyonSourceTime, TimeGenerated) >= starttime)
```
michaelatmsft: Hi @jwilke-halcyon, I am seeing some failures that originate from the `starttime` and  `endtime` conditions in  Parsers/ASimDns/Parsers/vimDnsHalcyon.yaml .

The ASIM DNS specification defines these parameters as filters on `TimeGenerated`, regardless of how `EventStartTime` and  `EventEndTime` are mapped. The parser currently applies them to  `coalesce(HalcyonSourceTime, TimeGenerated)` , which causes different behavior when the original Halcyon event time differs from the Sentinel `TimeGenerated` value.

Please update the conditions to:
```
and (isnull(starttime) or TimeGenerated >= starttime)
```

Please retain `EventStartTime = coalesce(HalcyonSourceTime, TimeGenerated)` , as that correctly preserves the original event time in the normalized output. The remaining `eventtype` failure message is expected because this source only produces `Query` events and should become nonblocking once the datetime failures are resolved.

michaelatmsft (changes requested):

Parsers/ASimDns/Parsers/vimDnsHalcyon.yaml
jwilke-halcyon: Awesome thank you for the review! We've updated the predicate change as you described in this latest push.

jwilke-halcyon:
@michaelatmsft @v-atulyadav @yummyblabla - We've updated the predicate change as requested in the most recent commit.

jwilke-halcyon:
@michaelatmsft - We need your sign off to move forward. We've made the changes you've requested.

jwilke-halcyon:
@v-atulyadav / @yummyblabla / @michaelatmsft - Are there any further changes we need to make?

michaelatmsft:
Taking another look now. Will respond shortly with any more changes or an approval

Parsers/ASimDns/Parsers/ASimDnsHalcyon.yaml
```diff
@@ -0,0 +1,166 @@
...
+          EventUid        = HalcyonEventId,
+          EventOriginalUid = HalcyonEventId,
+          EventMessage    = tostring(message),
+          EventOriginalType = tostring(type_name),
+          EventOriginalSubType = tostring(activity_name),
+          EventProduct    = "Halcyon",
+          EventVendor     = "Halcyon",
+          EventSchema     = "Dns",
+          EventSchemaVersion = "0.1.7",
+          AdditionalFields = bag_pack(
```
michaelatmsft, line 108: **Non-blocking follow-up:** Current ASIM authoring guidance recommends exposing `pack: bool = false` when a parser populates `AdditionalFields`, allowing callers to skip the dynamic-property reads and `bag_pack()` allocation when those fields are not needed. The source-agnostic `ASimDns` and `imDns` functions already expose `pack`, but existing DNS child parsers do not implement it consistently and the validators do not require it. I do not think this should delay the current PR, but it would be worth addressing consistently in a follow-up.

Parsers/ASimDns/Parsers/vimDnsHalcyon.yaml
```diff
@@ -0,0 +1,210 @@
...
+          DnsResponseName     = tostring(answers[0].rdata),
+          SrcUsername         = tostring(actor.user.name),
+          SrcUserId           = tostring(actor.user.uid),
+          SrcProcessName      = tostring(actor.process.file.name),
+          SrcProcessId        = tostring(actor.process.pid),
+          SrcProcessCommandLine = tostring(actor.process.cmd_line),
+          DvcHostname_raw     = tostring(device.hostname),
+          DvcId               = tostring(device.uid),
+          EventOriginalSeverity = tostring(severity)
+      | where (responsecodename == '*' or DnsResponseCodeName =~ responsecodename)
```
michaelatmsft, line 105: **Non-blocking performance follow-up:** `responsecodename` is currently applied after the DNS-class lookup and the larger normalized-field mapping block. A future optimization could calculate only `DnsResponseCodeName`, apply this filter, and then perform the lookup and remaining mappings. The current implementation is correct, all filtering tests pass, and the lookup is small, so this is not a merge blocker.

Parsers/ASimDns/Parsers/ASimDnsHalcyon.yaml
```diff
@@ -0,0 +1,166 @@
...
+          EventType           = "Query",                                             // ASIM Dns EventType enum has no Response/Traffic; Halcyon emits only queries
+          DnsQuery            = tostring(query.hostname),                              // OCSF query.hostname
+          DnsResponseCode     = toint(rcode_id),                                       // OCSF rcode_id (constant 0 = NoError; resolution succeeded)
+          DnsResponseCodeName = coalesce(tostring(rcode), tostring(_ASIM_LookupDnsResponseCode(toint(rcode_id)))),  // Halcyon emits rcode "NoError"; datatable fallback on rcode_id
+          DnsResponseName     = tostring(answers[0].rdata),
+          SrcUsername         = tostring(actor.user.name),
+          SrcUserId           = tostring(actor.user.uid),
+          SrcProcessName      = tostring(actor.process.file.name),
+          SrcProcessId        = tostring(actor.process.pid),
+          SrcProcessCommandLine = tostring(actor.process.cmd_line),
```
michaelatmsft, line 58: **Non-blocking cleanup:** `SrcProcessCommandLine` is calculated here but intentionally omitted by the final projection. It should not be added to the output because it is not defined for the DNS schema or Common rows in `ASimTester.csv`. Consider removing this unused calculation in a follow-up, although leaving it as-is should not block this PR.

michaelatmsft (changes requested):
The original datetime filtering feedback has been addressed and all required ASIM/KQL workflows now pass. I found one remaining release-metadata item that should be completed before merge: both source-agnostic DNS union parsers were bumped from `0.5.2` to `0.5.3`, but their changelogs still end at `0.5.2`.

Please add these entries above the existing `0.5.2` sections:

`Parsers/ASimDns/CHANGELOG/ASimDns.md`

```markdown
## Version 0.5.3

- (2026-06-09) Added ASimDnsHalcyon to the source-agnostic DNS parser - [PR #14606](https://github.com/Azure/Azure-Sentinel/pull/14606)
```

`Parsers/ASimDns/CHANGELOG/imDns.md`

```markdown
## Version 0.5.3

- (2026-06-09) Added vimDnsHalcyon to the source-agnostic filtering parser - [PR #14606](https://github.com/Azure/Azure-Sentinel/pull/14606)
```

The three inline notes are explicitly non-blocking follow-up suggestions and do not need to delay this PR. Once the changelog is updated we can approve the PR

v-atulyadav:
Hi @jwilke-halcyon,
Please address the reviewer’s comments and make the necessary changes. Thanks

jwilke-halcyon:
@michaelatmsft / @v-atulyadav - I've made the mandatory blocking change. I've captured the recommendations (thank you!) for an upcoming follow up PR.

michaelatmsft (approved):
The required `0.5.3` changelog entries were added exactly as requested. The latest commit changes only `Parsers/ASimDns/CHANGELOG/ASimDns.md` and `Parsers/ASimDns/CHANGELOG/imDns.md`.

All current-head ASIM checks now pass:
- Template validation
- Sample data ingestion
- Schema and data tests
- Parser filtering tests
- Fork security gate

The prior KQL validation passed on the unchanged parser code. The remaining inline suggestions were explicitly non-blocking follow-up ideas. No ASIM review blocker remains.

PR #14631, opened 2026-07-07
[ASIM] Cisco Secure Endpont - Missing Fields

Steve1145:
Change(s):

Added TimeGenerated and Type fields to existing Cisco Secure Endpoint AlertEvent parser

Reason for Change(s):

Fields now mandatory.

Version Updated:

Yes

Testing Completed:

Yes

Yes


yummyblabla (dismissed):

yummyblabla:
@Steve1145 Looks like it is failing KQL validations as it does not know that Type is a built-in column from LogAnalytics. You will need to add it explicitly to the CiscoSecureEndpointEventsV2_CL custom table definition

Steve1145:
> @Steve1145 Looks like it is failing KQL validations as it does not know that Type is a built-in column from LogAnalytics. You will need to add it explicitly to the CiscoSecureEndpointEventsV2_CL custom table definition

Thank you! Table now updated.

yummyblabla:
@v-atulyadav It looks like YAML file Syntax validation is failing for forked PRs. Can you look into this?

yummyblabla (approved):

PR #14673, opened 2026-07-13
Upwind Catalog Loader: expand to 6 API endpoints and fix ingestion/deployment bugs (v1.1.1)

ozahavi:
## Summary

Reworks the **Upwind** Microsoft Sentinel solution into the **Upwind Catalog Loader**, expanding it from a single-endpoint asset importer into a full ingestion connector covering six Upwind cloud-security datasets. Also fixes several deployment- and data-correctness bugs discovered through live end-to-end validation.

The solution deploys an Azure Function (Python 3.11, timer trigger) that authenticates to the Upwind platform via OAuth2 and streams data into custom Log Analytics tables using the Azure Monitor Ingestion API (DCE + DCR).

## What's new

**Expanded from 1 to 6 Upwind API endpoints**, each ingested to its own custom Log Analytics table via its own DCR stream:

| Dataset | Pagination strategy |
|---|---|
| Inventory / catalog assets (now **all** categories) | Cursor-based |
| Vulnerability findings | HTTP `Link`-header |
| Threat detections | Page-number |
| Threat events | Page-number |
| Threat stories | Time-windowed POST + cursor |
| Configuration (posture) findings | Time-windowed POST + cursor |

- Each dataset is fetched and uploaded **independently per run** — one endpoint failing (e.g. an org not entitled for it) no longer blocks the others.
- New `UpwindThreatLookbackMinutes` parameter controls the time window for the four time-windowed datasets (default: 90 minutes).

## Bug fixes

- **Critical: inventory asset columns were never populated.** The `Custom-UpwindCatalogAssets_CL` DCR stream declared columns in the destination table's PascalCase names with a pure passthrough transform, while the Upwind API sends snake_case fields — so every real field silently landed as `null` and only `TimeGenerated` ever had a value. **Broken since v1.0.0**, affecting 100% of ingested inventory rows on every install. Fixed by declaring the stream in raw snake_case and adding an explicit `transformKql` that maps each field into the table schema. Re-validated against a live Upwind demo org.
- **Support a Log Analytics workspace in a different resource group** (`WorkspaceResourceGroup` parameter) — the ARM template previously assumed the workspace shared the deployment's resource group, causing `workspaces/tables` write failures.
- **Fix Linux Function App deployment failure** (`"The parameter LinuxFxVersion has an invalid value"`) — the App Service Plan was missing `"reserved": true`, so it provisioned as Windows and silently blocked creation of every other resource in the template.
[...]

PR #14675, opened 2026-07-13
Upwind: support Log Analytics workspace in a different resource group

MitchellGulledge3:
## Summary

A customer deploying the **Upwind Logs Loader** data connector hit this failure whenever their Sentinel/Log Analytics workspace lives in a different resource group than the one they deploy the connector into:

```
Failed to perform 'write' on resource(s) of type workspaces/tables, because the parent resource
.../resourceGroups/<connector-rg>/providers/Microsoft.OperationalInsights/workspaces/<workspace-name>
[was not found].
```

**Root cause:** `azuredeploy_UpwindLogsLoader_API_FunctionApp.json` declares `Microsoft.OperationalInsights/workspaces/tables` as a top-level resource and resolves the workspace with the 2-argument `resourceId()` form. Both always resolve against the deployment's own target resource group — so any customer whose workspace lives in a separate resource group (a very common pattern) hits this immediately.

## Fix

- Added an **optional** `WorkspaceResourceGroup` parameter, defaulting to `[resourceGroup().name]` — existing single-resource-group deployments are unaffected.
- `logWsResId` now resolves with the 3-argument `resourceId()` form scoped to `WorkspaceResourceGroup`.
- Moved the `UpwindLogsAssets_CL` table into a nested `Microsoft.Resources/deployments` resource scoped to `WorkspaceResourceGroup` (the only piece that must live in the workspace's own resource group — Function App, Storage, App Insights, Key Vault, managed identity, DCE, and DCR all stay in the resource group the customer deploys into).
- Fixed the DCR's `dependsOn` to reference the nested deployment with the matching resource-group-qualified `resourceId()`.
- Docs: added `WorkspaceResourceGroup` to the parameter tables in `README.md`, the connector definition, and `Package/mainTemplate.json`; added a `3.0.3` entry to `ReleaseNotes.md`; bumped `Data/Solution_UpwindLogsLoader.json` `Version` to `3.0.3`.

## Validation

Deployed the patched template against a real Azure subscription with the Function App's resource group deliberately different from the Sentinel workspace's resource group:
- `az deployment group create` → `provisioningState: Succeeded`
- Confirmed `UpwindLogsAssets_CL` was created in the **workspace's** resource group while the Function App/DCE/DCR/Storage/Key Vault all landed in the **deployment's** resource group.
- Test resources torn down afterward.

## Note for the Upwind team

I'm submitting this on Upwind's behalf as a suggested fix after a mutual customer hit this exact error. Please feel free to check out this branch, review, amend, or push additional commits directly — happy to have Upwind take this over entirely, or I can keep iterating here if that's easier. No solution package zips were regenerated; assuming those are rebuilt as part of your normal release process.


MitchellGulledge3:
Closing in favor of #14673 — Upwind's own comprehensive PR (built directly on our reference fixes, plus the full 6-endpoint expansion) supersedes this narrower WorkspaceResourceGroup-only fix. Confirmed #14673 already correctly incorporates the same WorkspaceResourceGroup logic. Will follow up on #14673 with the missing Content Hub package zip.

PR #14680, opened 2026-07-14
Qualys KB CCF connector - fix incomplete data ingestion

Bhagya767:
- Widen query window from 10min to 1440min (24h) to prevent data loss from API truncation
- Reduce query delay from 600min to 60min so recent vulnerabilities are ingested sooner
- Increase timeout from 60s to 120s for large API responses
- Switch from published_after/before to last_modified_after/before to capture updated records
- Separate filter parameters (is_patchable, discovery_method, discovery_auth_types) into individual query parameters instead of concatenating into show_qid_change_log
- Repackage solution as v4.0.1


m-ruiz21:
were you able to confirm the same data is being ingested? The issue we've observed is that Qualys seems to have a 8-10hr delay before publishing the issues on their API, does this still address that? 
 
I'm wondering that even if the query window is larger, the data close to the beginning of the query window won't populate and we'll just skip over it. 

For example: 

Like for example, say we're querying up to hour 0d 0h: 
 
Query Time: 1d 1h
Query Window Fetches: 0d 0h - 1d 0h 
 
Concern: data from 1d 0h doesn't return in the Qualys API until 1d 10h

PR #14697, opened 2026-07-16
New solution: HoneyLabs threat intelligence (TAXII)

honeylabshq:
Change(s):
- New solution at Solutions/HoneyLabs: threat intelligence generated by HoneyLabs internet-facing honeypot sensors (exploitation source IPs and malware loader/C2 URLs), ingested through the built-in Threat Intelligence - TAXII data connector against the HoneyLabs TAXII 2.1 server (https://honeylabs.net/taxii2/).
- Data connector page (instruction-style) with copyable configuration values for the two TAXII collections.
- Three scheduled analytic rules mapping HoneyLabs indicators to CommonSecurityLog (IP and URL) and SigninLogs (IP), modeled on the standard TI-map templates and scoped with SourceSystem startswith 'HoneyLabs'.
- Incident enrichment playbook (HoneyLabs-EnrichIncident-IP) that comments the HoneyLabs verdict, event counts and report link on IP entities.
- Logo at Logos/HoneyLabs.svg, package generated with the V3 packaging tool (3.0.0).

Reason for Change(s):
- New community solution. HoneyLabs indicators are evidence-backed (captured exploit/loader commands rather than scan traffic), exclude known research scanners, and expire via valid_until as activity stops. The TAXII server requires a free HoneyLabs API key as the HTTP Basic password.

Version Updated:
- Not applicable, new solution (all content at version 1.0.0, package 3.0.0).

Testing Completed:
- Yes. The TAXII endpoint is live and was verified end to end (discovery, collections, objects with added_after paging and manifest). The analytic rule KQL was validated against the ThreatIntelligenceIndicator schema; rules are adapted from the shipped TI-map templates. Package built and validated with the V3 tool.

- Yes. arm-ttk passes except 'IDs Should Be Derived From ResourceIDs' flagging the contentProductId pattern that the V3 tool itself generates (variables('analyticRuleObject1')._analyticRulecontentProductId1); shipped solutions such as Solutions/GitLab carry the identical pattern, so this appears to be an arm-ttk 0.21 false positive. Happy to adjust if the pipeline says otherwise.

Solution provider contact: info@honeylabs.net (HoneyLabs, https://honeylabs.net).


honeylabshq:

honeylabshq:
Thanks for the review. Pushed fixes for the two failing checks:

- **DetectionTemplateSchemaValidation**: removed the custom `HoneyLabsThreatIntelligenceTAXII` connectorId from the analytic rules' `requiredDataConnectors`. Indicators arrive through the built-in `ThreatIntelligenceTaxii` connector, which the rules already require, so the rules no longer reference an id outside `ValidConnectorIds.json`.
- **SolutionValidations**: corrected 'Sentinel' to 'Microsoft Sentinel' in the data connector description, and set a conventional `BasePath`.

On the Copilot comments against `Package/mainTemplate.json` (duplicate `metadata`/`dataConnectors` resources and the `[[concat(...)` expressions): these are the standard output of the V3 solution packaging tool, not hand-authored. The doubled resources are the Content Hub pattern (a resource inside `contentTemplates.mainTemplate.resources` plus the top-level catalog entry), and `[[` is the correct ARM escaping for literals inside a nested template. They match the structure of currently shipped solutions, so I've left the generated package as-is. The ReleaseNotes table also renders correctly (single leading pipe per row). Happy to adjust if the pipeline flags anything further.

honeylabshq:
Updated from the legacy `ThreatIntelligenceIndicator` table to `ThreatIntelIndicators`. Added a workbook, four hunting queries, and an ASIM rule so workspaces without CEF are covered.

Connector setup is now one TAXII entry instead of one per collection, with malware infrastructure optional. The CommonSecurityLog rule no longer matches IPs out of the free-text `Message` field, and indicators carry a confidence score based on how many attacks we observed.

v-shukore:
Hi @honeylabshq, the workbookmetadata previewimages field cannot be blank. Please add at least one black and white image and include the solution logo and preview images at the location below.
https://github.com/Azure/Azure-Sentinel/tree/master/Workbooks/Images
Also, please confirm this is community supported solution.
Could you please confirm which type of connector uses this solution because this is not Sentinel Codeless Connector Framework (CCF). We would be happy to work with you to migrate your solution to CCF. Please contact the Microsoft Sentinel Partners at [AzureSentinelPartner@microsoft.com](mailto:AzureSentinelPartner@microsoft.com) to discuss available options. If this connector needed, please take approval from the azure app assure team on mail and share the screenshot with us to proceed with this connector. 
Thanks!!

honeylabshq:
Thanks @v-shukore. Both points addressed, and the connector question resolved by removing the connector.

**Preview images.** Added `HoneyLabsThreatIntelligenceWhite.png` and `HoneyLabsThreatIntelligenceBlack.png` under `Workbooks/Images/Preview/`, and the logo at `Workbooks/Images/Logos/HoneyLabs.svg`. `previewImagesFileNames` in `WorkbooksMetadata.json` now lists both. They are real screenshots from a live workspace with the feed connected, not mockups.

**Support model.** Community. `SolutionMetadata.json` has `publisherId: microsoftsentinelcommunity` and `support.tier: Community`, contact info@honeylabs.net.

**Connector type.** It was a `GenericUI` connector, and you are right that it should not ship as-is. Looking at what it actually did: it had no data plane at all. It was an instructions page plus a connectivity query. Indicators have always arrived through Microsoft's built-in **Threat Intelligence - TAXII** connector, which polls our TAXII 2.1 server and writes to `ThreatIntelIndicators`.

That means CCF has nothing to express here. Per `Tools/Create-Azure-Sentinel-Solution/V3/CCF_README.md`, a CCF connector needs a definition plus a poller whose `kind` is one of `RestApiPoller`, `WebSocket`, `GCP`, `AmazonWebServicesS3`, `Push` or `StorageAccountBlobContainer`, plus a DCR. There is no TAXII poller kind, and threat intelligence does not land in a DCR-targeted table. Building one would mean duplicating Microsoft's own TAXII connector and asking users to choose between two ingestion paths.

So rather than migrate it or ask for an exception, I removed it. Version 3.0.2 ships no data connector, matching the other two TAXII-based threat intelligence solutions in this repo (`Australian Cyber Security Centre` and `NCSC-NL NDN Cyber Threat Intelligence Sharing`), neither of which ships one. The setup values moved to the workbook and a new `README.md`, and the connection tile in the workbook has a `noDataMessage` pointing at them. The analytic rules already required only built-in connector ids (`ThreatIntelligenceTaxii`, `CEF`, `AzureActiveDirectory`), so nothing else changed.

No approval from App Assure or Sentinel Partners should be needed now, since there is no custom connector left in the solution. Happy to be corrected if you read it differently.

**Verification.** I deployed `Package/mainTemplate.json` into a live Sentinel workspace rather than only validating it. Result: 11 resources, `provisioningState: Succeeded`, solution registered at 3.0.2, 10 content items installed (1 workbook, 4 analytic rules, 4 hunting queries, 1 playbook), zero `DataConnector` content, and the workbook renders against real data.

Also in this round, from a pass over the workbook: it now bins on the indicator's own activity time rather than ingest time, so a fresh connection shows real history instead of one point; there is a match panel for normalised network data (ASIM) next to the CEF one, and both explain themselves when the underlying table is absent; and matched addresses and probed CVEs link through to the evidence behind them.

honeylabshq:
Hi @v-shukore, we removed the data connector in 3.0.2, so there is no custom connector left in the solution. Could you please re-verify, and approve the workflow run so the checks can report?

v-shukore:
Hi @honeylabshq, this is initial version of solution, so you have to package this solution with version 3.0.0. Thanks!

honeylabshq:
Hi @v-shukore, this is now packaged as 3.0.0 as requested and merged with the latest master to clear the conflict, so it should be ready for another look. Thanks!

v-shukore (approved):

v-atulyadav (approved):

PR #14738, opened 2026-07-22
Intel471 v3.0.1: community hunting queries and playbook Key Vault fixes

mmolenda:
## Change(s):
- Added 25 community hunting queries to Solutions/Intel471/Hunting Queries/ (refreshed community set).
- Updated Solutions/Intel471/Data/Solution_Intel471.json — added the Hunting Queries array + HuntingQueryBladeDescription, and the Description now names the Verity 471 backend so the solution is discoverable by it in Content Hub.
- Solutions/Cyborg Security HUNTER/ is left in place, unchanged (restored per review feedback). Retiring that solution will be handled separately — its offer belongs to a different publisher, so it is out of scope here.
- Reissued the content IDs of the two hunting queries that would otherwise duplicate their Cyborg Security HUNTER counterparts, since Content Hub keys hunting queries on id:
  - potential-maldoc-execution-chain-observed.yaml → 02909982-664b-4858-b703-429cbe031fef
  - powershell-encoded-command-execution.yaml → 3a5a21fd-5a82-45dc-ad37-3fb95f475e11
- Added hidden-SentinelTemplateName / hidden-SentinelTemplateVersion tags to both playbooks so the packaging tool emits real playbook template versions — Sentinel playbook 2.0, Graph Security playbook 1.1 — instead of defaulting to 1.0 on every release.
- Key Vault fixes in the Sentinel playbook:
  - GetUsername / GetApiKey built their path from a single concat() expression, which the Logic App designer cannot bind to the Key Vault Get secret operation, leaving those actions uneditable in the portal. They now use the standard /secrets/@{encodeURIComponent(...)}/value form used elsewhere in the repo.
  - Secret names are overridable at deployment via new optional KeyVaultUsernameSecretName / KeyVaultApiKeySecretName parameters, defaulting to the per-backend names as before.
- Fixed YAML indentation in autorun-or-asep-registry-key-modification.yaml and aws-identity-and-access-management-iam-discovery.yaml: continuation lines inside the query: | block had been dedented to column 0, terminating the block scalar and making the files unparseable. Only indentation changed; the KQL from those commits (the not() grouping and the switch to SourceIpAddress) is preserved.
- Documentation for the Sentinel playbook: how to obtain Titan and Verity credentials (Verity's Client ID / Client Secret are the API user name and key), granting Key Vault access on vaults that use the legacy vault access policy model, an Upgrading from version 2.0.0 section, and a note that installing the solution does not itself create the logic app. Added sample azuredeploy.parameters.json for both playbooks and a RELEASE_CHECKLIST.md.
- User-Agent headers now follow Intel471-<Integration>/<version> and carry the solution version.
- Fixed a stale storage listKeys apiVersion (2022-09-01 → 2025-01-01) in both playbook azuredeploy.json files so the regenerated package passes ARM-TTK's apiVersion-recency check.
- Regenerated the solution package with the V3 tool: Package/mainTemplate.json, createUiDefinition.json, and Package/3.0.1.zip → 2 Playbooks + 25 Hunting Queries + 0 Data Connectors. The previously committed 4.0.0.zip was removed as it was never published.
- Added a 3.0.1 row to Solutions/Intel471/ReleaseNotes.md.

## Reason for Change(s):
- Intel 471 acquired Cyborg Security; the community hunt content is being brought under the Intel 471 brand into a single Content Hub solution. The Cyborg solution itself stays in the repo for now and will be retired separately.
- Refreshes the community hunt queries to the latest set (adds MDE DeviceEvents support and corrects invalid SecurityEvent → SecurityEvents connectorIds).
- The playbook template version had never moved off 1.0 across solution versions 2.0.0 and 3.0.0, because the hidden-SentinelTemplateVersion tag was absent. Content Hub therefore could not detect that a playbook template had changed, existing installations never received the updated playbook, and customers had to uninstall the solution first. This release fixes the tags and the release notes state the new template versions.
- The Key Vault and documentation changes address customer-reported issues: the credential steps could not be edited in the designer, the required secret names were not configurable, and there was no guidance on which Intel 471 credentials to use or on Key Vaults using the legacy access-policy model.
- Metadata already lists Intel 471 as the support owner (SolutionMetadata.json), satisfying the "Intel 471 as support owner" requirement.

## Version Updated:
- Yes — Intel471 solution version 3.0.0 → 3.0.1, generated with the V3 tool as requested in review (rather than the previously proposed 4.0.0). Synced across Data/Solution_Intel471.json, Package/mainTemplate.json, ReleaseNotes.md and the package filename. Playbook template versions moved independently: Sentinel playbook 1.0 → 2.0, Graph Security playbook 1.0 → 1.1. (Note: content is Hunting Queries, not Analytic Rule/Detection templates.)

## Testing Completed:
- Yes — packaged with the V3 tool (createSolutionV3.ps1) and validated locally:
  - ARM-TTK: Pass=48 Fail=0 Total=48 via .script/local-validation/run-arm-ttk.ps1.
  - All 25 hunting query YAML files parse successfully, and the package is confirmed to contain 25 hunting queries and 2 playbooks.
  - Package/3.0.1.zip verified byte-identical to the mainTemplate.json and createUiDefinition.json committed alongside it.
  - KQL / Detection-Schema / Non-ASCII validators require .NET 3.1/8.0 (not available on the build machine, which has .NET 10) and will run in CI.
- The Sentinel playbook was deployed end to end against the Verity backend and confirmed to retrieve credentials from Key Vault and upload indicators successfully.
- The 25 hunting queries use only standard tables (SecurityEvent, DeviceEvents, AWSCloudTrail) via built-in connectors — no custom parsers, functions, or tables are required.

## Checked that the validations are passing and have addressed any issues that are present:
- Yes. Resolved the ARM-TTK failure surfaced by repackaging (playbook storage listKeys apiVersion recency), and fixed two hunting query files whose YAML had become unparseable — worth noting, because the packaging tool stops at the first unparseable hunting query without failing, and had silently produced a package containing 2 of the 25 queries. All locally-runnable validators now pass; remaining validators (KQL/schema/non-ASCII) will execute in CI.


v-shukore:
Hi @mmolenda, please review the copilot comment and implement them if required and resolve KQL validations failures. Thanks!
image

mmolenda:
Issues resolved.

v-shukore:
Hi @mmolenda, please resolve kql validation failure as failing due to column name is not present in table schema please check once this. Thanks!

v-shukore:
Hi @mmolenda, 
Could you explain why the entire "Cyber Security Hunter" solution was deleted from the repo?
For the "Intel471" solution, the zip package version has been updated to 4.0.0, so please update the package using the V3 tool with version 3.0.1 because other solutions in the repo all using V3 tool, so this will help keep the state consistent and make the changes easier to commit.
Thanks!

mmolenda:
Hi @v-shukore, Intel 471 acquired Cyborg Security, the HUNTER content is consolidated into the Intel 471 solution, and the Cyborg offer is going to be retired via Partner Center. 

Please let me know if you prefer us to restore Cyborg solution in this PR and handle it separately before I move on with packaging and another review round.

v-shukore:
Hi @mmolenda, thanks for the update.
It would be a good practice to restore the cyborg solution in this PR, raise a separate PR for it, and keep only the changes related to the "Intel471" solution in this PR.
Also, update the package using the V3 tool with version 3.0.1 because other solutions in the repo all using V3 tool, so this will help keep the state consistent and make the changes easier to commit.

v-shukore:
Hi @mmolenda, I deployed the maintemplate in my workspace to test the queries and found an issue in the Hunting Query description: it should be no more than 255 characters. Please update the hunting query descriptions. Also keep releasenote short and simple to understand.
image

v-shukore:
Hi @mmolenda, one last thing: could you please update the playbook name it gives error due to length while creating it, as shown in the screenshot below? Also, keep release note short and simple. Thanks!
image

v-shukore (approved):

v-atulyadav (approved):

PR #14763, opened 2026-07-27
[ASIM] Top-level parser consistency

SpeedyFireCyclone:
   
   Change(s):
   - Add missing sub-parsers to top-level ProcessEvent & NetworkSession parsers

   Reason for Change(s):
   - While doing static analysis, we discovered missing parsers compared to what is supposed to be supported.

   Version Updated:
   - Yes

   Testing Completed:
   - No actual parser changes; Did not re-test the already-existing sub-parsers.


v-atulyadav:
Hi @SpeedyFireCyclone,
Please review the failed validation and take the necessary action to resolve it. Thanks

v-atulyadav:
Hi @SpeedyFireCyclone,
Please review the failed validation and take the necessary action to resolve the issue. Thanks

SpeedyFireCyclone:
Hi @v-atulyadav, this is an issue in the validation script. I can see that the NetworkSession unifying parsers get skipped with the following line:
`Skipping '<FILENAME>' as this is a union or empty parser file. This file won't be tested.`

This seems to be different for the ProcessEvent parsers, and is not related to my changes.

SpeedyFireCyclone:
@v-atulyadav, it seems like forks are not allowed to update the validation files.

> [2](https://github.com/Azure/Azure-Sentinel/actions/runs/32341923366/job/96342593437?pr=14763#step:5:43)
> ❌ Fork PRs cannot modify the asimParsersTest test infrastructure folder
> ❌ Modified test files detected:
>  - .script/tests/asimParsersTest/VerifyASimParserTemplate.py

Can someone that has permissions to add a branch directly on this repository please fix the validation script?

mporfirov:
> v-atulyadav, it seems like forks are not allowed to update the validation files.
> 
> > [2](https://github.com/Azure/Azure-Sentinel/actions/runs/32341923366/job/96342593437?pr=14763#step:5:43)
> > ❌ Fork PRs cannot modify the asimParsersTest test infrastructure folder
> > ❌ Modified test files detected:
> > 
> > * .script/tests/asimParsersTest/VerifyASimParserTemplate.py
> 
> Can someone that has permissions to add a branch directly on this repository please fix the validation script?

@SpeedyFireCyclone , could you revert the changes to  _.script/tests/asimParsersTest/VerifyASimParserTemplate.py_ and I'll take a look onto the root cause? I don't believe this is the right fix.

SpeedyFireCyclone:
Hi @mporfirov, thanks for having a look at this.

The issue is that the ProcessEvent schema has the option to use split Create/End top-level parsers, and the current validations don't take that into account. The parser name extractor attempts to interpret these as product parsers, and fails to extract a DeviceVendor-equivalent from them, causing the validation to fail. I do now see that there's a number of other scripts that have the same logic duplicated that I never updated in my attempt, so I'll leave the actual changes to you.

Parsers/ASimProcessEvent/Parsers/ASimProcessEventCreate.yaml
```diff
@@ -35,7 +35,7 @@ ParserQuery: |
   union isfuzzy=true
     vimProcessEmpty,
     ASimProcessEventMicrosoft365D(imProcessEventBuiltInDisabled or ('ExcludeASimProcessEventMicrosoft365D' in (DisabledParsers) )),
-    ASimProcessCreateMicrosoftSysmon(imProcessEventBuiltInDisabled or ('ExcludeASimProcessCreateMicrosoftSysmon' in (DisabledParsers) )),
+    ASimProcessEventCreateMicrosoftSysmon(imProcessEventBuiltInDisabled or ('ExcludeASimProcessCreateMicrosoftSysmon' in (DisabledParsers) )),
```
mporfirov: For consistency with ASimProcessEvent.yaml union-parser, ExcludeASimProcessCreateMicrosoftSysmon should be changed to ExcludeASimProcessEventCreateMicrosoftSysmon. I assume that ASimProcessEventCreateMicrosoftSysmon has never worked in this union-parser, so there should be no backward compatibility concerns. 

Similar change should be made in ASimProcessEventTerminate.yaml union-parser.

SpeedyFireCyclone: I found a good number of other typos and incorrect keys for the exclusion logic that would have caused incorrect behaviour compared to the [documentation](https://learn.microsoft.com/en-us/azure/sentinel/normalization-manage-parsers#use-a-modified-version-of-a-built-in-parser). These have all been updated to be in line with what's been described in that article.

mporfirov (approved):

hunngu-ms (approved):

PR #14765, opened 2026-07-27
usta account takeover prevention playbook fix

p-osman:
Changes:

- Updated the "PRODAFTUstaATP-Backfill" playbook so it is self-contained: it now
  provisions its own Data Collection Endpoint (DCE) and Data Collection Rule (DCR)
  and assigns its managed identity the "Monitoring Metrics Publisher" role on the DCR.

- Removed the previously required manual inputs for the Logs Ingestion endpoint URI
  and DCR immutable ID.

- Bumped solution version 3.0.0 -> 3.0.1 and playbook template version 1.0 -> 1.1
  (hidden-SentinelTemplateVersion), regenerated the Package (mainTemplate.json,
  createUiDefinition.json, 3.0.1.zip), and added a 3.0.1 entry to ReleaseNotes.md.

---

Reason for Change(s):
   - Running the backfill previously required the operator to manually create/wire up a
  DCE + DCR and grant the playbook identity RBAC before it would work. Making the
  playbook provision these resources itself removes that manual setup and makes the
  historical backfill deployable in one step.

---

Version Updated:
   - N/A for the specific requirement (no Detections/Analytic Rule template versions
  changed in this update). For completeness: solution version 3.0.0 -> 3.0.1 and
  playbook template version 1.0 -> 1.1 were updated.

---

Testing Completed:
   - Yes (playbook). The backfill playbook was deployed and run in a Microsoft Sentinel
  environment; it created its DCE/DCR, self-assigned Monitoring Metrics Publisher, and
  ingested historical records into PRODAFTUstaCompromisedCredentials_CL.
  
  -  Prerequisite confirmed: the deploying user needs Owner or User Access Administrator
  on the target resource group, since the template creates a role assignment.
  
  -  Analytic rules / KQL were unchanged in this update and are unchanged from 3.0.0.

---

- Yes. Package regenerated with createSolutionV3.ps1. arm-ttk reviewed: the only two
  findings ("IDs Should Be Derived From ResourceIDs" on metadata contentProductId/id,
  and "URIs Should Be Properly Constructed") are the generator-produced patterns that
  the repo's CI gate (.github/actions/entrypoint.ps1) explicitly skips/filters, so they
  are non-blocking. No other validation failures.


p-osman:
Hi @v-atulyadav,

Could you please proceed with merging this if everything is OK?

v-atulyadav (approved):

PR #14776, opened 2026-07-28
Update ASIM skills for entity placeholders

michaelatmsft:
## Summary
- require ASIM parser creation to emit every schema-defined entity placeholder
- preserve the exact role-specific field names and casing from `ASimTester.csv`
- require ASIM PR reviews to validate completeness in parameter-less parsers and consistency in parameterized parsers

## Placeholder behavior
- fields are selected from the target schema and applicable `Common` rows
- every `*EntityKey` is emitted as an empty string
- every `*AdditionalIds` field and `AdditionalEntities` is emitted as an empty dynamic array
- placeholders remain empty until mappings are defined

## Dependency
The entity field definitions follow the schema changes proposed in #14628. This PR should merge after #14628, or be reconciled with its final field inventory if that proposal changes.

PR #14797, opened 2026-07-30
Restrict unsupported patterns in ASIM parser skills

michaelatmsft:
## Summary
- keep ASIM parser creation as a single event-source, per-record normalization pipeline
- allow query-local static `datatable` + `lookup` mappings with unique keys
- prohibit same-table and cross-table event enrichment
- prohibit one-to-many fan-out, all `mv-*` operators, and event-record aggregation or reaggregation
- add matching High-priority checks to the ASIM PR reviewer for both ASim and vim parsers
- preserve the schema-aware entity-placeholder guidance merged in #14776

## DCR gap analysis alignment
This applies the guidance from section 6 of `AsimDcrGapAnalysis.pdf` while preserving translator-supported static lookup mappings:
- no same-table or cross-table event enrichment
- no normalization fan-out from one source record to multiple records
- no `mv-*` followed by `summarize` or other event-record aggregation/reaggregation patterns
- query-local static lookups remain valid and can be translated to scalar `case` expressions

## Local skill tests
Ran the project skills locally through Copilot CLI using synthetic KQL cases:

| Scenario | Expected | Result |
|---|---|---|
| Static `datatable` + `lookup` with unique keys | Allowed | Passed |
| External event-table `join` | High-priority finding | Passed |
| `mv-expand` fan-out | High-priority finding | Passed |
| `summarize arg_max` event deduplication | High-priority finding | Passed |
| Scalar `case` mapping | Allowed | Passed |
| Parameterized parser preserving static lookup with early native filtering | Allowed | Passed |

## Rebase validation
- rebased the two intended commits onto current `master`
- preserved both entity-placeholder consistency checks from #14776 and prohibited-pattern checks
- resolved the overlap by keeping query-local static mappings allowed
- confirmed the final diff contains only the three ASIM skill files

## Validation
- Prettier 1.19.1 formatting check
- skill YAML frontmatter check
- `git diff --check`

### DIFF
--- a/.github/skills/asim-parser-create-parser/SKILL.md
+++ b/.github/skills/asim-parser-create-parser/SKILL.md
@@ -40,10 +40,22 @@ Parsers are KQL functions that follow a clear flow: **Filter → Parse → Map**
 - Use indexes so only relevant extents are scanned.
 - Filter early on native columns before parsing to improve performance.
 - Use high-performance parsing operators (`split`, `parse-kv`, `parse`) and avoid regular expressions for string parsing.
-- Normalize values with `iff`, `case`, or lookup tables rather than copying source values directly.
+- Normalize values with scalar expressions such as `iff` or `case` rather than copying source values directly.
 - Use `project-rename` for mapping, `extend` for calculated or normalized fields.
 - Try to map as many fields as possible, including optional ones. This increases usefulness and future-proofs the parser for schema changes.
 - Do not use `project-away` to remove unmapped columns. Use `project` instead, as `project-away` does not protect the parser from schema changes in the source data.
+
+### Prohibited patterns
+
+ASIM parsers must remain a single pipeline over the declared source table. Do not use:
+
+- **Same-table enrichment** — including a second read of the source table, a self-join, or a tabular subquery used to add fields.
+- **Cross-table or other table-shaped enrichment** — including `join`, `lookup`, a secondary `union` branch, `externaldata`, or inline `datatable` mappings.
+- **One-to-many fan-out** — each source record must produce at most one normalized record. Do not expand one record into multiple normalized records; this indicates that the connector or source event shape must be corrected.
+- **Any `mv-*` operator** — including `mv-expand` and `mv-apply`.
+- **Aggregation, reaggregation, correlation, or deduplication** — including `summarize`, `distinct`, `arg_min`, `arg_max`, or any other operation that combines source records or collapses expanded rows.
+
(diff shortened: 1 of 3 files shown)


.github/skills/asim-parser-create-parser/SKILL.md
```diff
@@ -40,11 +40,23 @@ Parsers are KQL functions that follow a clear flow: **Filter → Parse → Map**
 - Use indexes so only relevant extents are scanned.
 - Filter early on native columns before parsing to improve performance.
 - Use high-performance parsing operators (`split`, `parse-kv`, `parse`) and avoid regular expressions for string parsing.
-- Normalize values with `iff`, `case`, or lookup tables rather than copying source values directly.
+- Normalize values with scalar expressions such as `iff` or `case` rather than copying source values directly.
```
hunngu-ms: We should allow lookup tables

.github/skills/asim-parser-create-parser/SKILL.md
```diff
@@ -40,11 +40,23 @@ Parsers are KQL functions that follow a clear flow: **Filter → Parse → Map**
...
 - Use `project-rename` for mapping, `extend` for calculated or normalized fields.
 - Try to map as many fields as possible, including optional ones. This increases usefulness and future-proofs the parser for schema changes.
 - Do not use `project-away` to remove unmapped columns. Use `project` instead, as `project-away` does not protect the parser from schema changes in the source data.
 
+### Prohibited patterns
+
+ASIM parsers must remain a single pipeline over the declared source table. Do not use:
```
hunngu-ms: What does "single pipeline" mean ?

.github/skills/asim-parser-create-parser/SKILL.md
michaelatmsft: Updated in 3bec9718db. Query-local static mappings defined with `datatable` and applied with `lookup` are now explicitly allowed. The skill distinguishes those mappings from same-table or cross-table event enrichment and requires unique lookup keys so the mapping cannot fan out records.

I also tested the updated creator and reviewer skills locally: both accepted a static `datatable` + `lookup`, while the reviewer reported an external event-table join as High priority.

.github/skills/asim-parser-create-parser/SKILL.md
michaelatmsft: Updated in 3bec9718db. I replaced "single pipeline" with the precise requirement that the parser reads event records from one declared source table. Query-local static `datatable` + `lookup` mappings are allowed because they do not read or correlate additional event records.

hunngu-ms (approved):

PR #14816, opened 2026-08-03
usta integration fixes commit - 04072026

p-osman:
Change(s):
   - **Data Connector** (`PRODAFTUstaATP_PollingConfig.json`): build the paging `nextPageUrl`
     with the ARM `uri()` function instead of `concat()`.
   - **Backfill Playbook** (`PRODAFTUstaATP-Backfill/azuredeploy.json`, template version 1.1 -> 1.2):
     the deployment now creates the `PRODAFTUstaCompromisedCredentials_CL` table when it does not
     already exist, via a nested deployment scoped to the workspace resource group. The playbook's
     DCR now `dependsOn` that deployment. Adds a `TableRetentionDays` parameter (default 90).
   - **Backfill Playbook readme**: corrected the prerequisites, which previously stated that
     installing the solution creates the table, and documented the new permission and parameter.
   - **Solution version** 3.0.1 -> 3.0.2, with a matching `ReleaseNotes.md` entry.
   - **Package** regenerated: `Package/mainTemplate.json` and new `Package/3.0.2.zip`.
   - **Workbook images**: added the black/white preview images and the logo under the solution's
     `Workbooks/Images/` folder (identical to the copies already at the repo root).

   Reason for Change(s):
   - The ARM-TTK check *URIs Should Be Properly Constructed* failed with
     `Function 'concat' found within 'nextPageUrl'`. The test permits `uri()` but not `concat()`
     in properties whose name ends in `uri`/`url`.
   - The backfill playbook failed when run before the codeless connector had ingested its first
     record: the connector only materialises `PRODAFTUstaCompromisedCredentials_CL` on first
     ingestion, and a DCR cannot reference a destination table that does not exist yet. Deploying
     the table is a PUT, so it creates the table when absent and is a no-op when the connector has
     already created it.
   - Review feedback: add both black-and-white workbook preview images inside the solution folder.

   Version Updated:
   - Not applicable - no Detections/Analytic Rule templates were changed in this PR.
   - For completeness: solution version 3.0.1 -> 3.0.2, and the backfill playbook template
     version 1.1 -> 1.2 (`hidden-SentinelTemplateVersion` plus its `releaseNotes` entry).

   Testing Completed:
   - Yes

   - Yes
   - ARM-TTK on `mainTemplate.json` and `createUiDefinition.json`: passing, including the
     *URIs Should Be Properly Constructed* check that previously failed.
   - KQL validation, JSON/YAML syntax, solution, playbook, workbook-template, field-type and
     hyperlink validations: all passing locally.


p-osman:
Hi @v-atulyadav @v-maheshbh,

I submitted a commit addressing all the issues.

- **`dependsOn` resource group**: fixed. `TableDeploymentResourceId` now resolves in `WorkspaceResourceGroup`, matching where the nested deployment actually runs. Previously the dependency was silently dropped when the RGs differed.
- **Outer-scope evaluation**: added `"expressionEvaluationOptions": { "scope": "outer" }` explicitly. Note this is already the ARM default (the parent's parameters/variables do resolve), so the original wasn't broken, but explicit is better. No secure parameters are used inside the nested template.
- **Prerequisites**: added `Microsoft.OperationalInsights/workspaces/tables/write` (Log Analytics Contributor).
- **"no-op" wording**: reworded to create-or-update, noting retention comes from `TableRetentionDays`.

The `mainTemplate.json` comments are the packaged copy of the same playbook, so repackaging carried both fixes through.

Validated locally: ARM-TTK (29/0 and 18/0, including the *URIs Should Be Properly Constructed* check that originally failed), KQL, non-ASCII, and the solution/playbook/workbook validators.

Could you please approve the workflow run?

p-osman:
Hi @v-atulyadav @v-maheshbh,

Seems like all the CI checks are passed. If everything is OK, could we please continue merging this PR?

Thanks.

v-maheshbh:
Hi @p-osman 

Kindly attach the invocation logs for reference.

Thanks!

p-osman:
Hi @v-maheshbh,

Here are the images
atp_running_1
atp_running_2

v-maheshbh (approved):

PR #14850, opened 2026-08-06
Revert 1Password CCP connector to time-window polling to stop duplicate ingestion

eileen-zhaoo:
Change(s):
- Reverted the CCP data connector polling configuration in `Solutions/1Password/Data Connectors/1Password_ccpv2/` for all three pollers (`OnePasswordSignInEvents`, `OnePasswordAuditEvents`, `OnePasswordItemUsageEvents`):
  - `pagingType`: `PersistentToken` → `NextPageToken`
  - Removed `pagingInfoPlacement: RequestBody`
  - Restored `queryTimeFormat: yyyy-MM-ddTHH:mm:ssZ`
  - Restored the bounded request body `{"limit": 1000, "start_time": "{_QueryWindowStartTime}", "end_time": "{_QueryWindowEndTime}" }` in place of `{}`
- Applied the same change to `azuredeploy_1Password_poller_connector.json`, and corrected its hardcoded `_solutionVersion` (was stale at `3.0.3`) and `dataConnectorVersionConnections`.
- Solution version `3.0.4` → `3.0.5`, `DataConnectorCCFVersion` `1.0.2` → `1.0.3`.
- Regenerated `Package/mainTemplate.json`, `Package/createUiDefinition.json` and added `Package/3.0.5.zip` using the V3 packaging tool.
- Added a `ReleaseNotes.md` entry for 3.0.5.

This reverts **only** the paging change from #14625. The Events API v2 endpoint migration and added table columns from #14626 are retained. The PowerShell/Function App connector is not touched.

Reason for Change(s):
- #14625 replaced time-windowed requests with a cursor and began sending `{}` as the request body on every poll. The 1Password Events API treats a body with no `start_time` as a new reset cursor with a rolling lookback, so every poll re-fetched and re-ingested the same events instead of continuing from a stored cursor.
- Effect: every event was written once per poll for roughly the first hour of its life. Measured in a live workspace, a single `reveal` event was ingested **59 times**, with the last copy arriving **59m51s** after the event occurred, and copy counts scaling linearly with event age across all three streams.
- Root cause under investigation: `PersistentToken` appears not to retain the cursor between polls for this API.

Version Updated:
- Yes — solution version bumped `3.0.4` → `3.0.5` and `DataConnectorCCFVersion` bumped `1.0.2` → `1.0.3`.
- No Detections/Analytic Rule templates were modified, so no analytic rule version bumps were required.

Testing Completed:
- The duplicate ingestion was reproduced and measured in a live Microsoft Sentinel workspace running the affected connector version (counts per unique event, ingestion-time spread per event, and per-stream duplication rate).

- Yes.
- JSON validity confirmed for all modified and generated files.
- `arm-ttk` (`Test-AzTemplate`) run locally against `Solutions/1Password/Package`: 49 of 50 tests pass. The single failure is `IDs Should Be Derived From ResourceIDs` (23 errors in `mainTemplate.json`). This failure is **pre-existing** — it reproduces identically against the already-released 3.0.4 package and is not introduced by this change.

Known limitation:
- #14625 was originally written to stop late-synced item usage events (e.g. `reveal`, `secure-copy`) from being dropped. Because the 1Password Events API v1/v2 endpoints filter on the event's own timestamp and item usage events are withheld server-side, reverting to forward-moving time windows reinstates that gap.


v-atulyadav:
Hi @eileen-zhaoo,
Please review and address the suggestions above. Also, please share screenshots of the connector running successfully, along with the invocation logs, for verification. Thanks

eileen-zhaoo:
Hey @v-atulyadav ,

On the Copilot review: all three comments are the same finding on the same paging block in three files, and the concern is valid however we know this bug was pre-existing to #14625; every release through 3.0.2 had `NextPageToken` with `nextPageParaName: "cursor"` and no `pagingInfoPlacement` on a POST endpoint. #14625 incidentally removed it by adding `pagingInfoPlacement: "RequestBody"`, and 3.0.4 inherited that. This PR restores the pre-#14625 configuration exactly, so it does reintroduce it.

However, the suggested fix can't be applied as written. Copilot proposes restoring `pagingInfoPlacement: "RequestBody"` "so that start_time/end_time remain present while paging." The 1Password Events API treats a cursor and the reset fields (`start_time`/`end_time`/`limit`) as mutually exclusive and returns 400 when both are present. A page-2 body of `{"limit":1000,"start_time":...,"end_time":...,"cursor":"..."}` would fail on every paged request. That constraint is also why #14625 moved to a bare {}.

The form that satisfies both the framework and the API appears to be `pagingInfoPlacement: "RequestBody"` together with `pagingQueryParamOnly: true`, so page 1 sends the bounded window and subsequent pages send only the cursor.

We're happy to include that fix in this PR, however, I'd have questions because I haven't been able to determine an answer from reading the documentation. If `pagingQueryParamOnly: true` also strips the body on the first request of a cycle, page 1 would send `{}`, which the Events API treats as a reset cursor with a rolling one-hour lookback — reintroducing the exact duplication this PR exists to stop. That would be strictly worse than shipping the revert as-is. So the answer to question 2 below decides whether we add it here or track it separately.

Three questions, in order of how much they block us:

1. Does `pagingQueryParamOnly: true` omit the `queryParametersTemplate` body on the first request of a polling cycle, or only on subsequent paging requests? This is the one blocking the fix above. If the body is preserved on the first request and omitted only when paging, we'll add `pagingInfoPlacement` + `pagingQueryParamOnly` to this PR.

2. Do solution upgrades apply changes to resources nested in `contentTemplates`? In `mainTemplate.json` the DCR, the table, and all three `RestApiPoller` connectors are nested inside `contentTemplates` rather than deployed directly. I upgraded a live workspace from 3.0.3 to 3.0.4 via Content Hub; the solution version updated, but the five columns 3.0.4 adds to `OnePasswordEventLogs_CL` are still not present on the table. Is that expected — i.e. does an upgrade only re-register the templates, leaving resources that were already materialised untouched until the connector is disconnected and reconnected? I'm asking because if the pollers behave the same way, this revert wouldn't reach existing installs on upgrade alone, and we'd need to document a reconnect step in the release notes.

3. Does `PersistentToken` retain the token across polling cycles, or only within a single execution? This is the root cause behind #14625's duplication and I can't resolve it from the documentation. The [connection rules reference](https://learn.microsoft.com/en-us/azure/sentinel/data-connector-connection-rules-reference) says the token "persists server side ... The client continues where it left off, even if it makes new requests later," while [this Q&A answer](https://learn.microsoft.com/en-us/answers/questions/5862296/ccf-data-connector-pagination-and-checkpoint-handi) states "Pagination tokens exist only inside ONE execution. They are NOT persisted to the next run." Observed behaviour matched the latter — every poll started a fresh cursor, producing the rolling-hour re-read. This doesn't block this PR, but it determines whether cursor-based polling is viable for us at all.

Also could you clarify what invocation logs are and where I can find them? I am new to Sentinel and I can't find anything through Googling.

Thank you.

v-atulyadav:
Hi @eileen-zhaoo,
If you have tested the changes and there are no issues, we can proceed with the merge. Please click the **Resolve conversation** button and share screenshots confirming that the connector is running successfully.

eileen-zhaoo:
Hey @v-atulyadav , here's the screenshot confirming the connector is running successfully. So we can continue working on addressing the root cause of the bug, could you provide some guidance to the 3 questions I asked above? Thank you! 
Screenshot 2026-08-11 at 3 55 41 PM

v-atulyadav:
Hi @eileen-zhaoo,
Could you please execute a query to identify and return any duplicate records based on the **Event ID** or another appropriate unique column?

Regarding your queries, we have reached out to the concerned team. They informed us that they are currently analyzing the queries, and it may take some time to confirm the appropriate implementation. They need to review the relevant product documentation before they can provide a confirmation. We’ll share an update once we hear back from them.
Thank you.

eileen-zhaoo:
Screenshot 2026-08-20 at 12 07 45 PM

Hi @v-atulyadav, 

Ran this over 7 days of continuous collection in a fresh workspace deployed from this PR's mainTemplate.json.

On the Event ID part: OnePasswordEventLogs_CL has no populated unique identifier in this version. The schema includes uuid_s, but the Events API returns the field as uuid and the DCR has no stream declaration mapping it, so the column is empty on every row. I hashed each row's full contents instead, excluding ingestion metadata, so two rows only collide if every field the API returned is identical — a stricter test than a single-column key.

signinattempts — 1091 rows, 1091 unique, 0 duplicates
auditevents — 423 rows, 423 unique, 0 duplicates
itemusages — 179 rows, 174 unique, 5 duplicate rows, max 2 copies of any event
That's 5 duplicate rows out of 1693 over 7 days (0.3%), confined to itemusages. For comparison, on 3.0.4 a single event was ingested 59 times, with copies arriving over 59 minutes.

The residual appears pre-existing rather than introduced here: this PR restores the exact polling configuration that shipped through 3.0.2, and itemusages is the only stream on a 1-minute query window, so it crosses roughly 5x more window boundaries than the other two.

eileen-zhaoo:
@antipharmakon will be working on this moving forward.

v-atulyadav:
Hi @eileen-zhaoo,

Below are the answers to your questions.

Q1. Does pagingQueryParamOnly: true omit the queryParametersTemplate body on the first request of a cycle, or only on subsequent paging requests?
 Only on paging (2nd+) requests. The first request keeps the bounded window body; follow-up pages send just the cursor. The docs imply this but don't state it outright, so validate on a live poll before relying on it.
 
 
Q2. Do solution upgrades apply changes to resources nested in contentTemplates?
 No. A Content Hub upgrade only re-registers the templates and updates solution metadata — it doesn't touch already-deployed resources (table, DCR, active connection). That's why your 3.0.4 columns never showed. Existing installs must disconnect and reconnect the connector to get the fix, so add a reconnect note to the release notes.
 
 
Q3. Does PersistentToken retain the token across polling cycles, or only within a single execution?
 By design/docs it should persist the cursor across cycles (that's what separates it from NextPageToken). But your test showed a fresh cursor each poll, so real behavior didn't match. Docs are contradictory here — confirm with the CCF product team via a support case. Not a blocker for this PR.
 
I hope this information is helpful. We are ready to merge this PR. Please confirm if you’re okay with us proceeding with the merge. Thanks

eileen-zhaoo:
Yes, we can proceed with this merge

antipharmakon:
+1!

v-atulyadav (approved):

PR #14924, opened 2026-08-16
Fix TI Map Domain Entity to CommonSecurityLog matching on a literal string

nzdsf2-gif:
Change(s):
- `Solutions/Threat Intelligence (NEW)/Analytic Rules/DomainEntity_CommonSecurityLog.yaml`: `DomainName` is now derived from `ObservableValue` instead of `IndicatorType`.

Reason for Change(s):
- In the `Domain_Indicators` block, `DomainName` is assigned `tolower(IndicatorType)`. Two lines earlier the query filters on `IndicatorType == "domain-name"`, so at that point `IndicatorType` is the constant string `domain-name`, not the indicator value.

```kusto
| extend IndicatorType = replace(@"\[|\]|\""", "", tostring(split(ObservableKey, ":", 0)))
| where IndicatorType == "domain-name"
| extend DomainName = tolower(IndicatorType)     // constant "domain-name"
...
Domain_Indicators | join kind=innerunique (SecurityLog) on $left.DomainName == $right.Domain
```

- The join therefore compares the constant `domain-name` against the domain parsed out of `CommonSecurityLog`, so the rule can only return a result for a host literally named `domain-name`. In practice it produces no matches and no error.
- The fix matches what two sibling rules in the same solution already do: `DomainEntity_DnsEvents.yaml` uses `DomainName = tolower(ObservableValue)`, and so does the `Domain_Indicators` block in `DomainEntity_CloudAppEvents_Updated.yaml`.

Related, not changed here:
- `DomainEntity_CloudAppEvents_Updated.yaml` has the same `tolower(IndicatorType)` assignment in its earlier TLD-collection block, which makes the resulting `tld` set contain only `domain-name`. Its join uses the correct `Domain_Indicators` block, so the impact is different and I left it alone rather than widening this PR. Happy to submit that separately if you would like it fixed.
- `Package/mainTemplate.json` also carries the old expression, so it will need regenerating for the packaged solution to pick this up.

Version Updated:
- Yes, 1.4.5 to 1.4.6.

Testing Completed:
- No, not end to end. I do not have a `CommonSecurityLog` source to run the full join against, so I have not observed the corrected rule firing on real traffic and did not want to claim otherwise.
- What I did verify: the corrected expression is the one already in use by `DomainEntity_DnsEvents.yaml` and by the `Domain_Indicators` block of `DomainEntity_CloudAppEvents_Updated.yaml`; `ObservableValue` is in scope at that point in the query and survives the following `summarize arg_max(TimeGenerated, *) by Id, ObservableValue`; and `ObservableValue` is present in the `ThreatIntelIndicators` schema under `.script/tests/KqlvalidationsTests/CustomTables`.
- Note this template is listed in `.script/tests/KqlvalidationsTests/SkipValidationsTemplates.json`, so the KQL validation job does not cover it either way.

- Yes. Single file changed, version field updated in the diff as `checkThatTemplatesVersionWasChanged.sh` requires.


v-shukore:
Hi @nzdsf2-gif, 
Please repackage this solution using V3 tool.
https://github.com/Azure/Azure-Sentinel/blob/master/Tools/Create-Azure-Sentinel-Solution/V3/README.md.

nzdsf2-gif:
Hello v-shukore,

Repackaged with the V3 tool as requested — Package/mainTemplate.json
<https://github.com/nzdsf2-gif/relayshield/blob/main/Package/mainTemplate.json>,
createUiDefinition.json and 3.0.21.zip are regenerated in 7c871321. Version
stayed at 3.0.21 since the catalog reports 3.0.20 as published.

Two notes from the run. The tool's ARM-TTK step reports Api versions must
be the latest or under 2 years old — 2020-06-01 of
Microsoft.Resources/deployments is 2270 days old; that comes from the
packaging templates rather than this change, so I've left it alone.
Separately, while verifying I noticed TI Map Domain entity to Cloud App
Events has the same defect this PR fixes —
DomainEntity_CloudAppEvents_Updated.yaml:32 sets DomainName =
tolower(IndicatorType) and then splits it to build the TLD list, so that
list resolves to {"name"}. Happy to open a separate PR for it if useful.

Andrew / RelayShield

nzdsf2-gif:
Agree. Company name for this CLA: RelayShield LLC.

Regards,
Andrew

v-shukore:
Hi @nzdsf2-gif, please accept CLA in given format for example 

`@microsoft-github-policy-service agree company="RelayShield LLC"`

Also, check the arm-ttk validation failure causing due to hardcoded value in maintemplate please check and update
image

Thanks!

nzdsf2-gif:
@microsoft-github-policy-service agree company="RelayShield LLC"

v-shukore (dismissed):

v-shukore (approved):

v-atulyadav (approved):

PR #14948, opened 2026-08-19
Add hunting queries: break-glass account hunting pack (3 queries)

descambiado:
## Summary

Three hunting queries for monitoring designated emergency break-glass accounts, which should show no sign-in, credential, or membership activity outside a documented recovery test.

### BreakGlassAccountSignIn (SigninLogs)

Identifies any sign-in from a designated break-glass account. These accounts are deliberately excluded from Conditional Access and day-to-day MFA, so a successful sign-in bypasses the same controls protecting every other identity in the tenant.

MITRE: T1078.004, InitialAccess, Persistence

### BreakGlassAccountCredentialsModified (AuditLogs)

Identifies password resets, security-info registrations, and MFA changes on a break-glass account. Matches official Microsoft phrasing for password-reset detection (noun+verb match) rather than an exact-string list, since operation names vary across tenants and service versions.

MITRE: T1098, T1556.006, Persistence, DefenseEvasion

### BreakGlassAccountRoleOrGroupMembershipChanged (AuditLogs)

Identifies direct role assignment and group-mediated privilege changes on a break-glass account. Covers both paths since a group-mediated addition to a role-assignable group grants the inherited role without ever generating a direct role-assignment event.

MITRE: T1098.003, PrivilegeEscalation, Persistence

## Validation

- All three descriptions under 255 characters (204, 245, 214)
- Names sentence case, under the 100-char hard limit
- No non-ASCII characters
- GUIDs unique (verified against existing queries in the repo)
- All three share and document the `BreakGlassAccounts` watchlist dependency
- MITRE tactics/techniques verified against ATT&CK v16

### DIFF
--- a/Hunting Queries/AuditLogs/BreakGlassAccountCredentialsModified.yaml
+++ b/Hunting Queries/AuditLogs/BreakGlassAccountCredentialsModified.yaml
@@ -15,9 +15,9 @@ description-detailed: |
   named `BreakGlassAccounts` with a single column, `AccountUPN`, listing the user principal names
   of the tenant's designated emergency access accounts.
   Every match should be checked against the change calendar for a planned access-recovery test
-  before escalating. A match with no corresponding test window, or one immediately followed by a
-  sign-in from the same account (see the companion hunting query), should be treated as high
-  priority.
+  before escalating. A match with no corresponding test window, or one closely followed by a
+  sign-in or a role and group membership change on the same account (see the two companion
+  hunting queries in this pack), should be treated as high priority.
   References:
   - https://learn.microsoft.com/entra/identity/role-based-access-control/security-emergency-access
   - https://attack.mitre.org/techniques/T1098/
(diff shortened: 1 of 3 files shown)


v-atulyadav:
Hi @descambiado,
Please review the suggestions and make the necessary changes accordingly. Once you have completed the requested changes, please click the **Resolve conversation** button.

descambiado:
Thanks for the review. All four points addressed in `ee49fbf`:

**Role target extraction.** Fair catch, and the inconsistency was mine: the group path already read the user from whichever `TargetResources` entry describes it, while the role path still used a fixed index, and my own comment explained why a fixed index is unsafe. Both paths now use the same `mv-apply` on `type =~ "User"`, taking the UPN and that entry's `modifiedProperties` together. Description updated so it reflects both paths rather than only the group one.

**Group name on removals.** Right, `newValue` is empty on removals and the name sits in `oldValue`. `GroupNames` now uses `coalesce(newValue, oldValue)`, matching what `RoleNames` was already doing.

**Singular credential term.** Added `"credential"` alongside `"credentials"` in `PasswordWords`, since `has_any` is term-based and would otherwise miss singular variants.

**Dotted remove variant.** Added `"Remove member from role."` so add and remove are normalised the same way. I kept the explicit list rather than moving to a pattern match, since the surrounding queries in this folder use explicit `OperationName` lists, but happy to switch if you would rather see the pattern form.

v-atulyadav (approved):

PR #12927, opened 2025-10-07
Fix field name inconsistency in Fortigate parsers for ASim tables

v-atulyadav:
Required items, please complete

Change(s):

  Files from PR #12794

PR #14958, opened 2026-08-20
Fix Fortinet DNS parser default event type filtering

michaelatmsft:
## Summary

Fix the Fortinet FortiGate DNS filtering parser so the parameterless DNS unifying parser returns Fortinet DNS events.

`_Im_Dns()` uses the legacy default `eventtype='lookup'` and passes it to each source-specific filtering parser. Fortinet DNS events are normalized with `EventType='Query'`, but `vimDnsFortinetFortiGate` only accepted `Query` or `*`. As a result:

```kusto
_Im_Dns(starttime=ago(1d), endtime=now())
```

excluded Fortinet DNS records, while either of these returned them:

```kusto
_Im_Dns(starttime=ago(1d), endtime=now(), eventtype='Query')
_Im_Dns(starttime=ago(1d), endtime=now(), eventtype='*')
```

This causes incomplete results for investigations, hunting queries, workbooks, analytics, and detections that use the ASIM DNS abstraction without overriding `eventtype`.

## Root cause and regression history

- [PR #10150](https://github.com/Azure/Azure-Sentinel/pull/10150) originally fixed this compatibility issue by accepting both the schema-standard `Query` value and legacy `lookup` value.
- [PR #12927](https://github.com/Azure/Azure-Sentinel/pull/12927), specifically commit [`7a88704015f888f441b3ccf9628f2da0a3c45c2a`](https://github.com/Azure/Azure-Sentinel/commit/7a88704015f888f441b3ccf9628f2da0a3c45c2a), replaced that check with a direct comparison between `EventTypeTmp='Query'` and the incoming parameter. This unintentionally removed `lookup` compatibility and reintroduced the issue.

Changing the unifying parser default from `lookup` to `Query` is not safe because other DNS source parsers still rely on the legacy value. Compatibility therefore belongs in the source-specific filtering parser.

## Changes

- Preserve the source parser's schema-standard `Query` default.
- Accept `Query`, legacy `lookup`, and `*` in `vimDnsFortinetFortiGate`.
- Add a KQL comment explaining why `lookup` support must not be removed.
- Update the parser from version 0.1.2 to 0.1.3 and add a changelog entry.
- Regenerate the corresponding ARM template.

## Regression coverage

#14970 adds an explicit trusted-workflow assertion that `vimDnsFortinetFortiGate` returns the same nonzero count for `eventtype='Query'` and `eventtype='lookup'`. It also stabilizes sample timestamps so the generic datetime assertions can validate the parser rather than the ingestion batch timing.

Because the `pull_request_target` workflow intentionally downloads its test scripts from `master`, #14970 must merge before this PR's ASIM workflow is rerun.

## Validation

- ASIM parser template validation passed for both Fortinet DNS parsers.
- Focused assertions verified that the YAML compatibility filter is present and that the generated ARM query and function parameters match the YAML source.
- ARM JSON parsing and diff checks passed.
- CI KQL validation passed.
- CI ASIM template validation, sample-data ingestion, and schema/data tests passed.
- The current filtering job reaches all 21 Fortinet rows but fails generic time-window assertions because Azure Monitor replaces the historical sample timestamps with effectively identical received times. #14970 addresses that test-infrastructure issue and adds direct coverage of the legacy `lookup` path.

## Merge order

1. #14970
2. This PR


michaelatmsft:
### CI status after rerun

All current checks pass except [`Run ASim Parser Filtering tests`](https://github.com/Azure/Azure-Sentinel/actions/runs/32418988807/job/96639308832).

The rerun confirms the behavior fixed here: invoking the modified parser with its `lookup` default returns all 21 Fortinet DNS sample rows. The job subsequently fails only because its generic datetime test assumes the newly ingested rows span more than one timestamp:

- `starttime`: 21 rows returned out of 21, so the count was not reduced.
- `endtime`: 0 rows returned at the computed midpoint.
- `eventtype`: one normalized value (`Query`), which the workflow itself identifies as the expected partial-validation case for DNS.

KQL validation passed on rerun, as did ASIM template validation, sample-data ingestion, and schema/data validation. No failure identifies the changed Fortinet filter or generated ARM template as invalid.

hunngu-ms (approved):

PR #14962, opened 2026-08-21
prevent credentials ingestion

TwistedAlex:
   
   Change(s):
   - rollback changes done in https://github.com/Azure/Azure-Sentinel/pull/14612 for version 3.1.0, as 
   1). it's breaking change that removed over 100 columns and use the same DCR. We can discuss if some of the changes are valid, however, to have the hotfix to prevent credentials ingestion, a rollback is a must in case breaking it.
   2). The change (version 3.1.0) is approved in the PR 14612 by [v-atulyadav](https://github.com/v-atulyadav) back on July 20, however, it never got certified or published.
   
   - Leave identity credentials information as blank, keep existing table schema, so customers' existing integration won't break.

   Reason for Change(s):
   - Prevent Identity Credentials property ingestion.
   It matches OCI security-token authentication, not the simpler user credential-ID format.
ST$ + JWT indicates OCI service-to-service/instance-principal authentication.
Its claims identify the OCI Streaming service as the principal.
OCI places the complete security token in the request-signing keyId; Audit then records that value under data.identity.credentials.
The embedded jwk is a public key, not a private key.
Oracle’s tenantId/userId/fingerprint description applies to ordinary user API-key authentication. The field can contain different formats depending on authentication type.
Security: Treat this value as sensitive. Oracle’s SDK explicitly redacts ST$... tokens from logs. The token alone normally cannot sign requests—the corresponding ephemeral private key is also required—but it exposes infrastructure identity information
   https://docs.oracle.com/en-us/iaas/Content/Audit/Reference/logeventreference.htm
   https://docs.oracle.com/en-us/iaas/tools/python/latest/api/audit/models/oci.audit.models.Identity.html

   Version Updated:
   - Updated to have new version that prevent credentials ingestion

   Testing Completed:
   - Yes
image

   - See guidance below
image


TwistedAlex:

jlheard (dismissed):

TwistedAlex:
Found another issue and pushed the fix:

rollback changes done in https://github.com/Azure/Azure-Sentinel/pull/14612 for version 3.1.0, as 
   1). it's breaking change that removed over 100 columns and use the same DCR. We can discuss if some of the changes are valid, however, to have the hotfix to prevent credentials ingestion, a rollback is a must in case breaking it.
   2). The change (version 3.1.0) is approved in the PR 14612 by [v-atulyadav](https://github.com/v-atulyadav) back on July 20, however, it never got certified or published.

TwistedAlex:
image

image

tested, non breaking, migrated from 3.0.10 to 3.0.11 without issue, credentials are removed.

we can have the discussion on performance issue that mentioned in https://github.com/Azure/Azure-Sentinel/pull/14612 later, however, the hot fix in this PR is needed now.

jlheard (approved):

PR #14974, opened 2026-08-25
Protect Infoblox API keys in playbooks

idoshabi07:
## Summary

- preserves Infoblox API keys as `SecureString` workflow parameters
- passes deployment secrets through `properties.parameters` instead of embedding them in workflow-definition defaults
- suppresses inputs and outputs for every HTTP action that consumes the key
- applies the protection consistently across 12 affected Infoblox playbooks

## AI scan item

- https://dev.azure.com/MSecProductSecurity/AI%20Vuln%20Scanning/_workitems/edit/130206

## Validation

- parsed all affected ARM templates as JSON
- verified 12 secure workflow parameter bindings
- verified secure input/output configuration on 26 HTTP actions
- deployed the cited playbook to `OptimizeSentinelResourceGroup` with a unique fake key
- confirmed the deployed resource GET did not return the fake key or runtime parameter value
- triggered the workflow and confirmed the fake key was absent from action metadata and inaccessible through the secured input link
- removed the test workflow and its role assignment after validation


idoshabi07:
image

v-rusraut (approved):

v-atulyadav (approved):

PR #14977, opened 2026-08-25
Fix Blob CCF role assignment GUID collisions

jlheard:
Change(s):
- Updated Blob CCF role-assignment GUID generation to include the resource scope, principal ID, and role definition ID.
- Applied the same deterministic logic to blob-container, notification-queue, and dead-letter-queue role assignments.
- Updated the shared packaging generator and standalone Blob CCF template.

Reason for Change(s):
- Prevents different connector principals targeting the same storage resources from receiving identical role-assignment GUIDs.
- Azure role assignments are immutable; reusing an existing GUID for another principal causes `RoleAssignmentUpdateNotPermitted`.

Version Updated:
- Not applicable. No Detection or Analytic Rule templates were modified.

Testing Completed:
- Yes.
- Validated PowerShell syntax for `storageAccountDeploymentTemplate.ps1`.
- Validated JSON syntax for `Connector_StorageBlob_CCF_template.json`.
- Exercised the generator and confirmed all three GUID expressions include scope, principal ID, and role definition ID.
- Confirmed `git diff --check` passes.
- Deployment Error:image
- Successful Connection on an impacted connector: image

- 👍


artafres (dismissed):

TwistedAlex (approved):

TwistedAlex (dismissed):

jlheard:
We have agreed to revert the CloudFare CCF solution changes and focus this PR solely on the packaging script fix.

TwistedAlex (dismissed):

artafres (approved):

v-atulyadav (approved):

PR #15025, opened 2026-09-01
Parser Cyble ThreatIntelIndicators

prasad-cyble:
Change(s):
- Added `Parsers/Cyble/ThreatIntelParser.yaml`, a KQL parser for threat intelligence ingested from the Cyble TAXII connector, using the `FunctionQuery` schema with function alias `CybleThreatIntel`.
- Added `Parsers/Cyble/README.md` covering installation, the fields the parser adds to `Data`, and sample queries.

Reason for Change(s):
- The Sentinel TAXII connector maps standard STIX properties to columns on `ThreatIntelIndicators` and stores the full object in `Data`, but `external_references` has no column of its own. Cyble ships the per-source `risk_score` and `confidence_rating` packed into `external_references[].description` as a single string, formatted as `risk_score=<int>; confidence_rating=<string>`, so those values are not filterable without hand-written string extraction in every query.
- The parser returns `ThreatIntelIndicators` unchanged and merges the parsed values into `Data` as top-level fields alongside `external_references`: `ext_source_name`, `<source_name>.risk_score` and `<source_name>.confidence_rating`. Keying by source name keeps indicators enriched by multiple sources (AbuseIPDB, censys, Shodan) from colliding.
- Source names are lowercased in both `ext_source_name` and the generated keys, so the two stay consistent and key names remain stable across the mixed casing the feed uses. An indicator enriched by AbuseIPDB returns `ext_source_name` of `abuseipdb` alongside `abuseipdb.risk_score` and `abuseipdb.confidence_rating`.
- The parser never filters. Every indicator that `ThreatIntelIndicators` would return is returned: indicators the source did not enrich come back with a `risk_score` of `0` and an empty `confidence_rating`, and indicators that carry no `external_references` at all pass through unchanged.
- `SourceSystemName` is left empty with a comment instructing the user to set it to the SourceSystem of their own TAXII connector.

Version Updated:
- N/A - not a detection or analytic rule template.

Testing Completed:
- Yes - validated against a Microsoft Sentinel workspace ingesting the Cyble TAXII collection.

- Yes


prasad-cyble:
Thanks for the review — all four points are valid. Here's how each is being addressed in an upcoming commit:

**1. `materialize()` on `Base`**

Agreed. `Base` was being inlined into both `WithRefs` and `WithoutRefs`, so the `ThreatIntelIndicators` scan and the `todynamic()` parse ran twice. It's now wrapped in `materialize()` so the table is scanned once and both branches reuse the same intermediate result.

**2. Casing inconsistency between `ext_source_name` and the generated keys**

Fixed at the source rather than in two places: `SourceName` is now lowercased in the `mv-apply` `extend`, so `make_set()` and the `strcat()` key names both derive from the same normalized value. The two `tolower()` calls inside `bag_pack` are removed as a result, and output is self-consistent — `ext_source_name` of `abuseipdb` next to `abuseipdb.risk_score`.

**3. PR description vs the `tolower()` behaviour**

Keeping the lowercasing and documenting it. The feed emits mixed casing across sources (`AbuseIPDB`, `censys`, `Shodan`), so normalizing keeps the generated key names stable for anyone writing queries against them. This is now stated explicitly in the parser header, and the PR description will be updated to match.

**4. Header comment referencing `External_SourceRatings`**

Replaced — that name doesn't appear anywhere in the parser. The new header names the actual source table, documents the expected `risk_score=<int>; confidence_rating=<string>` format of `external_references[].description` so users can validate their data shape before running it, lists the output fields with example values, and notes that indicators without `external_references` pass through unchanged.

v-rusraut:
Hi @prasad-cyble, The parser file extension should be .yaml. Please convert the file from .txt to .yaml and resolve the Copilot comment.

prasad-cyble:
Hello @v-rusraut i have resolved the bot comments 
please review and merge this PR. 

**Prasad Boke 
Cyble inc**

v-rusraut (approved):

v-dvedak (approved):

prasad-cyble:
Thanks @v-rusraut & @v-dvedak

PR #15040, opened 2026-09-03
Add AWS GuardDuty AlertEvent ASIM parsers

adbhogeGit:
## Summary
- add regular and filtering AlertEvent ASIM parsers for AWS GuardDuty
- normalize finding identity, severity, lifecycle, affected resource, user, action, and threat context
- register both parsers in the AlertEvent unifying parsers and include changelogs and synthetic sample data

## Validation
- passed `.script\tests\asimParsersTest\VerifyASimParserTemplate.py`
- Log Analytics execution is pending against a workspace containing AWSGuardDuty data

## Workspace tests
```kusto
AWSGuardDuty | getschema
AWSGuardDuty | count
AWSGuardDuty | take 2000
ASimAlertEventAWSGuardDuty(disabled=false, pack=false) | take 10
ASimAlertEventAWSGuardDuty(disabled=true) | count
ASimAlertEventAWSGuardDuty(disabled=false, pack=true) | take 10
vimAlertEventAWSGuardDuty(starttime=ago(30d), eventseverity_has_any=dynamic(["High"])) | take 10
```

PR #15079, opened 2026-09-08
Add AWS GuardDuty AlertEvent ASIM parsers

adbhogeGit:
## Summary
- add regular and filtering AlertEvent ASIM parsers for AWS GuardDuty
- normalize finding identity, severity, lifecycle, affected resource, user, action, and threat context
- register both parsers in the AlertEvent unifying parsers and include changelogs and synthetic sample data

## Validation
- passed `.script\tests\asimParsersTest\VerifyASimParserTemplate.py`
- Log Analytics execution is pending against a workspace containing AWSGuardDuty data

## Workspace tests
```kusto
AWSGuardDuty | getschema
AWSGuardDuty | count
AWSGuardDuty | take 2000
ASimAlertEventAWSGuardDuty(disabled=false, pack=false) | take 10
ASimAlertEventAWSGuardDuty(disabled=true) | count
ASimAlertEventAWSGuardDuty(disabled=false, pack=true) | take 10
vimAlertEventAWSGuardDuty(starttime=ago(30d), eventseverity_has_any=dynamic(["High"])) | take 10
```

Supersedes #15040 by moving the head branch into Azure/Azure-Sentinel so the ASIM ARM-generation workflow can run.


Parsers/ASimAlertEvent/CHANGELOG/imAlertEvent.md
```diff
@@ -1,5 +1,9 @@
 # Changelog for imAlertEvent.yaml
 
+## Version 0.1.8
+
+- (2026-09-02) AWS GuardDuty - AlertEvent ASIM Parser
```
bqiao-microsoft: Could you please add the PR link to the changelog like the other entries?

Parsers/ASimAlertEvent/Parsers/vimAlertEventAWSGuardDuty.yaml
```diff
@@ -0,0 +1,326 @@
...
+    | extend ActivityTypeParts = split(ActivityType, "/")
+    | extend
+        ThreatPurpose = tostring(split(tostring(ActivityTypeParts[0]), ":")[0]),
+        ActivityResourceType = tostring(split(tostring(ActivityTypeParts[0]), ":")[1]),
+        ThreatName = tostring(split(tostring(ActivityTypeParts[1]), ".")[0]),
+        ActionType = tostring(ServiceDetails.action.actionType),
+        IsArchived = tobool(ServiceDetails.archived),
+        OriginalUserType = tostring(ResourceDetails.accessKeyDetails.userType)
+    | extend
+        ResourceType = coalesce(tostring(ResourceDetails.resourceType), ActivityResourceType),
```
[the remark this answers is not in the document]
bqiao-microsoft: I don't think this is a good comment.

Parsers/ASimAlertEvent/CHANGELOG/ASimAlertEventAWSGuardDuty.md
```diff
@@ -0,0 +1,11 @@
+# Changelog for ASimAlertEventAWSGuardDuty
+
+## Version 0.1.1 - 2026-09-04
+
+- (2026-09-04) Ensured `ThreatRiskLevel` is emitted as an integer
+
+## Version 0.1.0 - 2026-09-02
```
bqiao-microsoft: This PR is for the initial version right? Why is there a version `0.1.0` and `0.1.1`?

Parsers/ASimAlertEvent/Parsers/vimAlertEventAWSGuardDuty.yaml
```diff
@@ -0,0 +1,326 @@
...
+          "Unknown"
+        ),
+        AlertVerdict = ""
+    | where array_length(ipaddr_has_any_prefix) == 0 or has_any_ipv4_prefix(DvcIpAddr, ipaddr_has_any_prefix)
+    | where array_length(hostname_has_any) == 0 or DvcHostname has_any (hostname_has_any)
+    | where array_length(username_has_any) == 0 or Username has_any (username_has_any)
+    | where array_length(attacktactics_has_any) == 0 or AttackTactics has_any (attacktactics_has_any)
+    | where array_length(attacktechniques_has_any) == 0
+    | where array_length(threatcategory_has_any) == 0 or ThreatCategory in~ (threatcategory_has_any)
+    | where array_length(alertverdict_has_any) == 0
```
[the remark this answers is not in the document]
adbhogeGit: This behavior is intentional and follows the ASIM filtering-parser convention. These parameters are retained to preserve the standard AlertEvent parser signature. Because AWSGuardDuty does not populate `AttackTechniques` or `AlertVerdict`, a non-empty constraint cannot match any source event and should return no rows rather than silently ignore the requested filter. The description no longer advertises these filters as supported

Parsers/ASimAlertEvent/Parsers/vimAlertEventAWSGuardDuty.yaml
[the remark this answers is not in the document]
adbhogeGit: Thanks for the suggestion, but I’m keeping the source-specific filtering parser self-contained. ASIM filtering parsers are designed to apply time and field predicates as early as possible, before completing normalization. Calling the parameter-less parser first would normalize every source row before filtering and defeat the performance purpose of the `vim` parser.

Parsers/ASimAlertEvent/CHANGELOG/imAlertEvent.md
adbhogeGit: Added the PR link to the GuardDuty entry

Parsers/ASimAlertEvent/CHANGELOG/ASimAlertEventAWSGuardDuty.md
adbhogeGit: Correct, this is the initial release. The intermediate `0.1.1` entry has been removed. Both GuardDuty parsers and their changelogs now use only version `0.1.0`.

Parsers/ASimAlertEvent/Parsers/ASimAlertEventAWSGuardDuty.yaml
```diff
@@ -0,0 +1,277 @@
...
+          ThreatPurpose =~ "Cryptocurrency", "Cryptominer",
+          ThreatPurpose =~ "Trojan", "Trojan",
+          ThreatPurpose =~ "Policy", "Security Policy Violation",
+          "Unknown"
+        ),
+        ThreatOriginalCategory = ThreatPurpose,
+        ThreatFirstReportedTime = TimeCreated,
+        ThreatLastReportedTime = TimeGenerated,
+        ThreatIsActive = iff(isnull(IsArchived), bool(null), not(IsArchived)),
+        ThreatRiskLevel = iff(Severity between (1 .. 10), toint(Severity * 10), int(null)),
```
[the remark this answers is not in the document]
adbhogeGit: Keeping the existing range check. AWS documents GuardDuty severity values as `1.0` through `10.0`, with Low severity covering `1.0–3.9`. Values between `0` and `1` are outside the documented source range. Valid decimal severities are already scaled by 10 and converted to `int`, consistently in both parser variants.

Parsers/ASimAlertEvent/Parsers/vimAlertEventAWSGuardDuty.yaml
```diff
@@ -0,0 +1,349 @@
...
+    disabled: bool = false,
+    pack: bool = false
+  )
+  {
+    AWSGuardDuty
+    | where not(disabled)
+    | where isnull(starttime) or TimeGenerated >= starttime
+    | where isnull(endtime) or TimeGenerated <= endtime
+    | where array_length(attacktechniques_has_any) == 0
+    | where array_length(alertverdict_has_any) == 0
```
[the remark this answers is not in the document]
adbhogeGit: This behavior is intentional and follows the ASIM filtering-parser convention. These parameters preserve the standard AlertEvent signature used by the unifying parser. Because this parser does not populate `AttackTechniques` or `AlertVerdict`, a non-empty constraint cannot match an event and should return no rows.


Ignoring the constraint would return events that do not satisfy the requested filter, while removing the parameters would break the shared signature. Leaving the existing behavior unchanged.

Parsers/ASimAlertEvent/Parsers/vimAlertEventAWSGuardDuty.yaml
```diff
@@ -0,0 +1,349 @@
...
+          ThreatPurpose =~ "Recon", "Reconnaissance",
+          ""
+        ),
+        ThreatCategory = case(
+          ThreatPurpose =~ "Cryptocurrency", "Cryptominer",
+          ThreatPurpose =~ "Trojan", "Trojan",
+          ThreatPurpose =~ "Policy", "Security Policy Violation",
+          "Unknown"
+        )
+    | where array_length(attacktactics_has_any) == 0 or AttackTactics has_any (attacktactics_has_any)
```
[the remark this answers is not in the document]
adbhogeGit: Keeping `has_any` for consistency with the `attacktactics_has_any` parameter used by the other AlertEvent parsers. The repository’s filtering tests deliberately exercise partial terms for `_has_any` parameters. Replacing this with `in~` would narrow the existing filtering behavior, even though this parser currently emits a single tactic.

Parsers/ASimAlertEvent/Parsers/vimAlertEventAWSGuardDuty.yaml
```diff
@@ -0,0 +1,349 @@
...
+        IndicatorAssociation = case(
+          tostring(ServiceDetails.resourceRole) =~ "TARGET", "Targeted",
+          tostring(ServiceDetails.resourceRole) =~ "ACTOR", "Associated",
+          ""
+        ),
+        ThreatOriginalCategory = ThreatPurpose,
+        ThreatFirstReportedTime = TimeCreated,
+        ThreatLastReportedTime = TimeGenerated,
+        ThreatIsActive = iff(isnull(IsArchived), bool(null), not(IsArchived)),
+        ThreatRiskLevel = iff(Severity between (1 .. 10), toint(Severity * 10), int(null)),
```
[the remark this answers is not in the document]
adbhogeGit: The example severity `0.1` is outside GuardDuty’s documented `1.0–10.0` range. The existing guard covers valid GuardDuty severities, including decimals, and leaves out-of-range risk values unmapped. Both parser variants use the same integer conversion, so I’m retaining the current mapping rather than expanding the supported range for this example.

Parsers/ASimAlertEvent/Parsers/ASimAlertEventAWSGuardDuty.yaml
```diff
@@ -0,0 +1,277 @@
...
+        Dvc = coalesce(DvcHostname, DvcId, DvcIpAddr),
+        Hostname = DvcHostname,
+        IpAddr = DvcIpAddr,
+        User = Username,
+        DvcEntityKey = "",
+        FileEntityKey = "",
+        ProcessEntityKey = "",
+        UserEntityKey = "",
+        UserAdditionalIds = dynamic(null),
+        AdditionalEntities = dynamic(null)
```
bqiao-microsoft: PR Review skill says we should use empty arrays i.e. dynamic([]), not dynamic(null).

Applies to UserAdditionalIds and AdditionalEntities.

Parsers/ASimAlertEvent/Parsers/vimAlertEventAWSGuardDuty.yaml
```diff
@@ -0,0 +1,349 @@
...
+        Rule = RuleName,
+        Dvc = coalesce(DvcHostname, DvcId, DvcIpAddr),
+        Hostname = DvcHostname,
+        IpAddr = DvcIpAddr,
+        User = Username,
+        DvcEntityKey = "",
+        FileEntityKey = "",
+        ProcessEntityKey = "",
+        UserEntityKey = "",
+        UserAdditionalIds = dynamic(null),
```
bqiao-microsoft: PR Review skill says we should use empty arrays i.e. dynamic([]), not dynamic(null).

Applies to UserAdditionalIds and AdditionalEntities.

Parsers/ASimAlertEvent/CHANGELOG/ASimAlertEventAWSGuardDuty.md
```diff
@@ -0,0 +1,8 @@
+# Changelog for ASimAlertEventAWSGuardDuty
+
+## Version 0.1.0 - 2026-09-09
+
+- (2026-09-09) Initial creation of the parser - [PR #15079](https://github.com/Azure/Azure-Sentinel/pull/15079)
+- Normalized GuardDuty finding identity, severity, lifecycle, resource, user, action, and threat context
```
bqiao-microsoft: I suggest removing all but the first line of the changelog. I think the other lines are confusing. The only thing necessary to say is that we're publishing this brand new parser. 

Applies to the changelog for the 'vim' parser as well.

Parsers/ASimAlertEvent/Parsers/ASimAlertEventAWSGuardDuty.yaml
adbhogeGit: Updated `UserAdditionalIds` and `AdditionalEntities` to `dynamic([])`. I also corrected the shared data tester to accept empty arrays while retaining element-type checks for non-empty arrays. No new exclusions were added.

Parsers/ASimAlertEvent/Parsers/vimAlertEventAWSGuardDuty.yaml
adbhogeGit: Updated both placeholders to `dynamic([])` in the filtering parser as well, consistent with the parameter-less parser and the review skill requirements.

Parsers/ASimAlertEvent/CHANGELOG/ASimAlertEventAWSGuardDuty.md
adbhogeGit: Simplified both GuardDuty changelogs to retain only the initial-creation entry with the PR link. The title and existing version/date headings are preserved, and the additional detail bullets have been removed.

bqiao-microsoft (approved):

PR #15080, opened 2026-09-08
[ASIM] Add more ingest-time parsers

bqiao-microsoft:
Add ingest-time parsers for CrowdStrikeDetections and AWSVPCFlow

### DIFF
--- a/Parsers/ASimAlertEvent/IngestTimeParsers/ASimAlertEventCrowdStrikeFalcon.yaml
+++ b/Parsers/ASimAlertEvent/IngestTimeParsers/ASimAlertEventCrowdStrikeFalcon.yaml
@@ -1,0 +1,186 @@
+Parser:
+  Title: CrowdStrike Falcon Alert Event ASIM ingestion-time parser
+  Version: '0.1.0'
+  LastUpdated: 'Sep 4, 2026'
+Product:
+  Name: CrowdStrike Falcon
+Normalization:
+  Schema: AlertEvent
+  Version: '0.1'
+Author: Microsoft
+Description: |
+  Normalizes CrowdStrike Falcon detections from the CrowdStrikeDetections
+  raw-table shape into the native ASimAlertEventLogs table during ingestion.
+ParserName: ASimAlertEventCrowdStrikeFalcon
+DCRType: Workspace
+InputStream: none
+SourceRawTable: CrowdStrikeDetections
+TargetASIMTable: ASimAlertEventLogs
+ParserFilterQuery: |
+  source
+ParserQuery: |
+  source
+  | extend Type = "CrowdStrikeDetections"
+  | extend
+      EventCount = int(1),
+      EventEndTime = CreatedTimestamp,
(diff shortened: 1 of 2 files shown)


Parsers/ASimAlertEvent/IngestTimeParsers/ASimAlertEventCrowdStrikeFalcon.yaml
```diff
@@ -0,0 +1,186 @@
...
+      UsernameType = case(
+          Username contains "@", "UPN",
+          isnotempty(Username), "Simple",
+          ""
+      ),
+      User = Username,
+      IpAddr = DvcIpAddr,
+      Hostname = DvcHostname,
+      Dvc = DvcId,
+      AdditionalFields = parse_json('null')
```
yummyblabla: AdditionalFields = dynamic({})

Parsers/ASimNetworkSession/IngestTimeParsers/ASimNetworkSessionAWSVPC.yaml
```diff
@@ -0,0 +1,221 @@
...
+      SrcAppName = PktSrcAwsService,
+      DstAppName = PktDstAwsService
+  | extend
+      IpAddr = SrcIpAddr,
+      Src = SrcIpAddr,
+      Dst = DstIpAddr,
+      Dvc = DvcId,
+      EventStartTime = TimeGenerated,
+      DvcInterface = DvcInboundInterface
+  | project-away Action, AzId, Bytes, FlowDirection, InstanceId, Packets, Protocol, Region, SourceSystem, SublocationId, SublocationType, SubnetId, TcpFlags, TenantId, TrafficPath, Version
```
yummyblabla, line 216: Use project instead of project-away

bqiao-microsoft: I agree with your reasoning why we shouldn't use project-away, but it's here in the ingest version because the query-time version uses it. I want to maintain as similar functionality between the two as possible. 

We should remove project-away from the ingest-time version when we remove it from the query-time version.

project-away is supported by DCR Kql so this will be functional.

Parsers/ASimAlertEvent/IngestTimeParsers/ASimAlertEventCrowdStrikeFalcon.yaml
bqiao-microsoft: Thanks, this made me notice that the ingest time version of the parser isn't doing packing.

Parsers/ASimNetworkSession/IngestTimeParsers/ASimNetworkSessionAWSVPC.yaml
yummyblabla: Fair point.

yummyblabla (approved):

PR #15095, opened 2026-09-10
Update the Workbench parser to retrieve the complete details of neste…

V1ManagedServices:
Update the Workbench parser to retrieve the complete details of nested fields.

## Change(s):
- Updated `TrendAIWorkbench_Complete.yaml` parser: replaced `maxif()` with `make_set_if()` for all multi-valued indicator/entity fields (FileName_s, FileHashValue_s, DomainName_s, FileDirectory_s, IPAddress, URL_s, RegistryKey_s, RegistryValue_s, RegistryValueName_s, ProcessCommandLine_s, MalwareName_s, HostHostName_s, MailboxPrimaryAddress_s), so all distinct values are returned instead of just one.
- Reworked `UserAccountNTDomain_s`/`UserAccountName_s` extraction to split the domain\username pair before aggregation (per-row), so domain/username stay correctly paired per account when there are multiple account entities on one alert.
- Synced the same query changes into `Package/mainTemplate.json` (both embedded copies of the parser query) and regenerated `Package/3.0.0.zip` to match.

## Reason for Change(s):
- `maxif()` collapses multi-valued indicator/entity fields to a single value, silently dropping data when an alert has more than one indicator of the same type (e.g. multiple file hashes or hostnames on one Workbench alert).
- `make_set_if()` preserves all distinct values as a JSON array, matching the actual cardinality of the source data.

## Version Updated:
- N/A — no Analytic Rule template changes in this PR.

## Testing Completed:
- Yes — manually verified in a live Sentinel test workspace (`ccf-field-test-instance`) against real Vision One Workbench alert data, confirming multi-valued fields (e.g. multiple file hashes/hostnames) are now returned as complete sets instead of a single value.
- Confirmed `mainTemplate.json` parser query matches the standalone parser YAML exactly.

## Checked that the validations are passing and have addressed any issues that are present:
- Yes — ran local validation scoped to the solution (`node .script/local-validation/validate.js --path "Solutions/TrendAI Vision One(CCF)"`): JSON/YAML syntax, Data Connector schema, MS branding, Solution metadata, Workbook template, Playbook, and ARM-TTK all passed. KQL/Detection-Schema/Non-ASCII .NET validators could not run locally due to missing .NET 8.0/3.1 runtimes on this machine (environment limitation, unrelated to this change).


v-skharode (approved):
Reviewed the parser update and the generated package. The source parser change and the packaged template are aligned, and I do not see a blocker in the diff. I did not find a source-level issue in the change set. Optional runtime validation in a Log Analytics workspace can be done separately if needed.

v-skharode (approved):

PR #15098, opened 2026-09-10
fix(Egress Defend): set offerId to egress-sentinel-defend and repackage

olliespires:
Marketplace certification requires the SolutionMetadata offerId to match the Partner Center offer id. Package regenerated with createSolutionV3 so solutionId picks up the new value.

   Change(s):
   - This was requested by validation after previous PR, screenshot shown below
image

   Reason for Change(s):
   - As above

   Version Updated:
   - Not updated, 3.1.0 was never released as it failed validation so just used that

   Testing Completed:
   - We have checked using the package upload tool and it looks as before, but we can't find where this name is updated. Screenshot below of the connectors. Let us know if we are missing something or if you are happy with this as is
image

   - Ran validations locally


v-skharode (approved):
Reviewed the changes. The PR updates the offerId to egress-sentinel-defend as requested in the certification feedback and regenerates the package artifacts. The additional mainTemplate.json changes appear to be packaging-generated updates associated with the metadata correction. No blocker identified.

PR #15012, opened 2026-08-28
Feedly: ingest IoCs into native ThreatIntelIndicators (3.0.5)

Edouard360:
Change(s):

- Point the Feedly CCF connector at Sentinel native `ThreatIntelIndicators` (`Microsoft-ThreatIntelIndicators`) instead of custom `feedly_indicators_CL`.
- Map Feedly `format=sentinel` IoCs to STIX fields (`ObservableKey` / `ObservableValue` / `Pattern` / `Data`) with `SourceSystem == "Feedly"` so they participate in TI matching, analytics, and hunting the same way as other TIP integrations.
- Bump the solution to 3.0.5 and drop the unused custom table resource.

Version Updated:
Yes, to 3.0.5

Testing Completed:
Yes

image

PR #15099, opened 2026-09-10
Fix Feedly CCF DCR transform for Azure Monitor KQL

Edouard360:
Change(s):

Fix bad Copilot suggestion from #15012 , my bad for not verifying the Copilot suggestion. DCR creation failed with InvalidTransformQuery / replace_string (runtime scalar function provider not found). That function came from a Copilot review suggestion on the 3.0.5 STIX Pattern escaping. It is valid Log Analytics KQL, but Azure Monitor DCR ingestion transforms only support replace (and pack, not bag_pack).
Keep the same native ThreatIntelIndicators mapping (ObservableKey / ObservableValue / Pattern / Data, SourceSystem == "Feedly"); Pattern still escapes \ and ' via replace().

Bump the solution to 3.0.6.

Version Updated:
Yes, to 3.0.6

Testing Completed:
Yes 

(3.0.5 Connect/ingestion was validated before the Copilot DCR change; this patch restores DCR-safe KQL)

image


Edouard360:
Hi @v-atulyadav, flagging that the failing check here isn't related to our solution content, it's failing on infrastructure/CI grounds.

The workflow is running as an unauthenticated Octokit client since this is a forked repo, and it's hitting GitHub's public API rate limit (60 req/hr, unauthenticated) before it can even fetch the PR details:

```
Creating unauthenticated Octokit client.
GET /repos/Azure/Azure-Sentinel/pulls/15099 - 403
API rate limit exceeded for 20.161.30.226
```

So changedFilesValidator never gets far enough to actually validate any files, it fails before that step. 
Could you take a look or re-run the check? Happy to provide the full log if useful.

v-atulyadav (approved):

PR #15103, opened 2026-09-10
Fix Google Apigee jsonPayload ingestion

v-krishnachi:
Fixes #

## Proposed Changes

- Preserve Google Apigee `jsonPayload` records in the existing dynamic `ProtoPayload` column.
- Retain the original `ProtoPayload` value when `jsonPayload` is absent.
- Regenerate the Google Apigee solution package as version 3.1.2.

## Validation

- Packaging succeeded and generated `Package/3.1.2.zip`.
- JSON, connector schema, solution, hyperlink, field type, and ARM-TTK validations passed.
- The package contents match the generated `mainTemplate.json` and `createUiDefinition.json`.
- The KQL validator reports two failures in unchanged legacy files (`ApigeeX_FunctionApp.json` and `Unified_ApigeeX.yaml`); the same failures reproduce on clean `origin/master` and are not regressions from this change.


v-krishnachi:
Corrected the DCR transform after deployment validation rejected dynamic(null). The transform now uses the repository-established todynamic('') fallback, also used by the GKE CCF DCR, and the 3.1.2 package was refreshed. JSON and solution validation plus all 48 ARM-TTK checks pass; the two KQL failures remain unchanged baseline failures in legacy files.

v-krishnachi:
Updated the DCR transform in commit `4c03de3bae` after live deployment showed that `column_ifexists` is not available in this DCR runtime.

The transform now uses the standard GCP stream fields directly:

```kusto
source | extend ProtoPayload = iff(isnotempty(jsonPayload), jsonPayload, protoPayload)
```

This preserves `jsonPayload` in the destination `ProtoPayload` column for runtime/application logs while retaining `protoPayload` for audit logs. The source DCR, generated `mainTemplate.json`, and `3.1.2.zip` were synchronized and validated.

v-krishnachi:
The latest deployment error (`Undefined symbol: jsonPayload`) confirmed that the standard input stream `Microsoft-GCPApigee` removes `jsonPayload` before the DCR transform runs.

Commit `7e4f7ec185` changes the connector to the established custom-stream pattern used by other GCP connectors:

- Collector writes raw Google LogEntry records to `Custom-GCPApigee`.
- The DCR explicitly declares `jsonPayload`, `protoPayload`, and the other source fields.
- The transform writes `jsonPayload` to the existing `ProtoPayload` destination when present, otherwise preserves audit-log `protoPayload`.
- Output remains the standard `Microsoft-GCPApigee` / `GCPApigee` table.

Validation:
- Azure server-side `az deployment group validate`: **Succeeded** for the exact DCR and transform.
- ARM-TTK: **48/48 passed**.
- Generated `mainTemplate.json` and `3.1.2.zip` verified in sync.
- The two existing parser KQL baseline failures remain unchanged.

v-atulyadav:
Hi @v-krishnachi,
For the PR below, please review Copilot’s suggestions and take the necessary action. Once addressed, click the Resolve conversation button.
Also, please share a screenshot of the successfully running connector along with the relevant logs. Thanks

PR #15144, opened 2026-09-16
fix(hunting): repair entity mappings in hunt_LOLBins.yaml

vl43den:
Change(s):
   - Pointed 'Account' and 'Host' entity mappings to valid columns 'SubjectAccount' and 'Computer'
   - Replaced 'SubjectMachineName' with 'Computer' in the project statement and added 'SubjectAccount'
   - Bumped version from 1.0.0 to 1.0.1

   Reason for Change(s):
   - 'entityMappings' pointed to 'AccountCustomEntity' and 'HostCustomEntity' but neither was created via 'extend' nor projected in the query, which would cause entity extraction to fail silently
   - SubjectMachineName  is not available/populated in Event ID 4688 Fields, 'Computer' would provide the actual host NetBios/FQDN
   - 'SubjectAccount' offers 'DOMAIN\User' which matches the required Account -> FullName
   - https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-10/security/threat-protection/auditing/event-4688
   - https://learn.microsoft.com/en-us/azure/azure-monitor/reference/tables/securityevent

   Version updated:
   - Yes

   Testing Completed:
   - Need Help (not tested on live sentinel)
   - Tested on dataexplorer with dummy data
   
image

   - Yes
   
image

   

> References: 
> - [Guidance for Detection checks](https://github.com/Azure/Azure-Sentinel#pull-request-detection-template-structure-validation-check)
> - [General contribution guidance](https://github.com/Azure/Azure-Sentinel/wiki#what-can-you-contribute-and-how-can-you-create-contributions)
> - [PR validation troubleshooting](https://github.com/Azure/Azure-Sentinel#pull-request)


v-skharode:
Copilot raised a valid concern regarding SubjectAccount. I could not find where this field is created before being projected and used in the Account entity mapping. 
Could you please confirm that SubjectAccount exists for these SecurityEvent records or explicitly construct it before projection/mapping?

vl43den:
Hi @v-skharode, thank you for the review & support. SubjectAccount is a built-in column of the SecurityEvent table, which is populated at ingestion as SubjectDomainName\SubjectUserName so it would not need to be constructed in the query.

 In  [NRT_base64_encoded_pefile.yaml](https://github.com/Azure/Azure-Sentinel/blob/b5d05ae933f559242ccae832dc59c53ca81ecf25/Solutions/Windows%20Security%20Events/Analytic%20Rules/NRT_base64_encoded_pefile.yaml#L31), the query runs with no project at all, only extending on HostName and DnsDomain from Computer, yet still mapping Account.FullName -> SubjectAccount and [NRT_execute_base64_decodedpayload.yaml](https://github.com/Azure/Azure-Sentinel/blob/master/Solutions/Windows%20Security%20Events/Analytic%20Rules/NRT_execute_base64_decodedpayload.yaml) also queries SecurityEvent with EventID ==4688, projects SubjectAccount directly with no extend and maps Account.FullName to it, as in my PR. Happy to switch the mapping instead if it's preferred!

Cheers,
vlad

v-skharode (approved):
Reviewed the changes and the clarification provided regarding SubjectAccount being a built-in SecurityEvent column. The entity mappings are consistent with existing Sentinel content and the version has been updated appropriately. No further issues found. Approving.

v-atulyadav (approved):

